PR #414, opened 2025-04-25
Topic/2025.1/deinternalize remove origin version

addabis:


Metalama.Backstage/src/Metalama.Backstage/Licensing/LicenseProduct.cs
```diff
@@ -25,6 +25,12 @@ public enum LicenseProduct : byte
         [Obsolete( "This product is no longer supported." )]
         PostSharp20 = 1,
 
+        /// <summary>
+        /// Should be normalized to <see cref="PostSharpUltimate"/>.
+        /// </summary>
+        [Obsolete( "Use PostSharpUltimate instead." )]
+        PostSharp30 = 2,
```
addabis: I'm not sure why change from `PostSharp30`->`PostSharpUltimate1` was made, it's even more confusing because of PostSharp implementation that would have to be renamed.

gfraiteur: I did that. Don't change the PostSharp implementation.

gfraiteur (approved):

PR #439, opened 2025-05-15
Delete redundant files.

gfraiteur:


addabis:
I think the Redis download made the build on TC unstable and I needed to add it back.

addabis (approved):

PR #1023, opened 2025-09-03
Design time: add support for introduced methods with covariant return types

### LINKED ISSUE #1022: Design time support for generated covariant return methods
## Is your feature request related to a problem? Please describe.

I'm implementing a builder pattern where I've something similar to this
```
    partial record RB(int a)
    {
        public partial class Builder
        {
            public virtual RB? Build() => null;
        }
    }
    partial record RD(int a, bool b) : RB(a)
    {
        public new partial class Builder : RB.Builder
        {
            public override RD? Build() => new(42, false);
        }
    }
```

note the derived builder return type benefits from return type covariance (https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/covariant-returns)

All works at run time and compile time, but the intellisense is broken and doesn't understand correctly the derived return type when using the derived builder (so it isn't showing the derived record members

Infact the "design time" code genereted from metalama does miss the override implementation because it's not aware of return covariance....

I think the issue originates here: 
https://github.com/metalama/Metalama/blob/7de741d09caa698ae1445ff232a5f5a48e5650b4/Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Introductions/Builders/MemberBuilder.cs#L79

I'm doing a PR solving the issue...

beppemarazzi:
fixes #1022


Metalama.Framework/src/Metalama.Framework/Code/MethodExtensions.cs
```diff
@@ -23,5 +23,11 @@ public static class MethodExtensions
 
         [CompileTime]
         public static AsyncInfo GetAsyncInfo( this IMethod method ) => ((ICompilationInternal) method.Compilation).Helpers.GetAsyncInfo( method );
+
+        [CompileTime]
+        public static bool IsCovariantReturnType( this IMethod method ) =>
```
gfraiteur: I suggest renaming it to `HasCovariantReturnType` since the method itself is not a return type.

Also, please add the XML doc.

gfraiteur:
Link to the spec: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/covariant-returns

gfraiteur:
Would you be willing to implement support for covariant read-only properties and indexers, too?
Otherwise, I will file a follow-up issue.

gfraiteur (changes requested):
Amazing! It seems you went very deep; design-time support is very complex and it's hard to find things.

I'm asking to rename a method and I suggest to extend the feature to covariant read-only and indexers. Please comment.

beppemarazzi:
@gfraiteur  i'm very very very newbie for metalama.... so i don't feel so confident to understand all the framework complexity!!!!
I tried to implement the covariance support also for properties but i had also to fix something here (in fact a duplication of covariant return type detection for readonly properites)

https://github.com/metalama/Metalama/blob/7de741d09caa698ae1445ff232a5f5a48e5650b4/Metalama.Framework/src/Metalama.Framework.Engine/AdviceImpl/Introduction/IntroducePropertyAdvice.cs#L276-L285

while it seems to work in UnitTests, i got errors (not understandable for my knowledge) in the real use case, so i reverted it....

its' my attempt with properties:

https://github.com/metalama/Metalama/compare/release/2025.1...beppemarazzi:Metalama:CovariantPropertiesAttempt?expand=1

gfraiteur:
> its' my attempt with properties:
> 
> https://github.com/metalama/Metalama/compare/release/2025.1...beppemarazzi:Metalama:CovariantPropertiesAttempt?expand=1

Ok, I will file this as another issue if this is not trivial.

gfraiteur:
I merged into our own topic branch because we need to run our CI pipeline and we cannot do that on your fork.

PR #1134, opened 2025-10-14
Topic/2026.0/story 1094 field expressions

### LINKED ISSUE #1094: C# 14 field keyword: support in target code (advising)

addabis:
Implements #1094


Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerAnalysisStep.AutoPropertyBodyWalker.cs
```diff
@@ -0,0 +1,36 @@
...
+        private List<FieldExpressionSyntax>? _fieldExpressions;
+
+        public IReadOnlyList<FieldExpressionSyntax> FieldExpressions => this._fieldExpressions ??= [];
+
+        public AutoPropertyBodyWalker() : base( SyntaxWalkerDepth.Node ) { }
+
+
+        public override void VisitFieldExpression( FieldExpressionSyntax node )
+        {
+            (this._fieldExpressions ??= new List<FieldExpressionSyntax>()).Add( node );
```
gfraiteur, line 29: Verify thread safety - not sure that it's not called concurrently.

gfraiteur (approved):

Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerAnalysisStep.AutoPropertyBodyWalker.cs
addabis: One visitor per property so it's fine (for now). But it should be parallelized in the callsite because this is not an edge case.

PR #1162, opened 2025-11-03
Add RSS client functionality and configuration options.

### LINKED ISSUE #1161: Display product news notifications
Display a toast notification when a blog post or a brief is published on the RSS feed.

UI to enable or disable the feature or choose briefs vs full posts only.

gfraiteur:


Metalama.Backstage/src/Metalama.Backstage/UserInterface/Rss/RssClient.cs
```diff
@@ -0,0 +1,260 @@
...
+            if ( configuration.LastFetchTime != null && pubDate != null && pubDate.Value <= configuration.LastFetchTime!.Value )
+            {
+                this._logger.Trace?.Log(
+                    $"Item published on {pubDate} is not newer than last fetch time {configuration.LastFetchTime}. Skipping notification." );
+
+                return;
+            }
+
+            // Update last fetch time.
+            this._configurationManager.Update<RssClientConfiguration>( c => c with { LastFetchTime = this._dateTimeProvider.UtcNow } );
```
addabis: This should probably use the time of the last observed item instead of local time; otherwise, we may miss some items in an edge case. Not important though.

addabis (approved):

PR #1228, opened 2025-12-04
Fix empty HTML output and enhance type system API (#1226)

### LINKED ISSUE #1226: Test framework: WriteInputHtml produces empty file for some tests
## Summary

When running aspect tests with `WriteInputHtml: true` in `metalamaTests.json`, some test files produce empty `.cs.html` files while others work correctly.

## Reproduction

In `Metalama.Documentation.SampleCode.AspectFramework`:

1. Create a new test file `SuppressWarningWithFilter.cs`:
```csharp
// This is public domain Metalama sample code.

namespace Doc.SuppressWarningWithFilter;

internal class MyService
{
    [SuppressUnusedVariable]
    public void Initialize()
    {
        var _initialized = true;
        var _other = 42;
    }
}
```

2. Create aspect file `SuppressWarningWithFilter.Aspect.cs`:
```csharp
// This is public domain Metalama sample code.

using Metalama.Framework.Code;
using Metalama.Framework.Diagnostics;
using System;
using System.Linq;

namespace Doc.SuppressWarningWithFilter;

internal class SuppressUnusedVariableAttribute : OverrideMethodAspect
{
    private static readonly SuppressionDefinition _suppressUnusedVariable = new("CS0219");

    public override void BuildAspect(IAspectBuilder<IMethod> builder)
    {
        base.BuildAspect(builder);
        builder.Diagnostics.Suppress(
            _suppressUnusedVariable.WithFilter(
                diag => diag.Arguments.Any(arg => arg is string name && name == "_initialized")),
            builder.Target);
    }

    public override dynamic? OverrideMethod()
    {
        Console.WriteLine($"Entering {meta.Target.Method.Name}");
        return meta.Proceed();
    }
}
```

3. Run `dotnet test --filter "SuppressWarningWithFilter"`

4. Check `obj/html/net9.0/SuppressWarningWithFilter.cs.html` - it is 0 bytes

Meanwhile, similar existing tests like `SuppressWarning.cs` produce valid HTML output.

## Analysis

Looking at `BaseTestRunner.WriteHtmlAsync` (line 831+), I noticed that at line 858-860 there's a `return` statement inside the loop that iterates over diagnostics. If any suppression matches a diagnostic, the method returns early without writing anything to the HTML file (the StreamWriter was created but nothing was written to it).

```csharp
if (suppressions.Any(...))
{
    return;  // <-- This exits the method, leaving the file empty
}
```

This seems incorrect - the `return` should probably be `continue` to skip the suppressed diagnostic but continue processing other diagnostics.

## Expected behavior

The input HTML file should be generated with the source code, regardless of whether diagnostics are suppressed.

### LINKED ISSUE #1230: CS8625 nullability warning in generated dependency injection code
## Summary

Generated code for dependency injection causes nullability warning CS8625 when adding optional constructor parameters.

## Reproduction

See `Metalama.Extensions/src/tests/Metalama.Extensions.DependencyInjection.AspectTests/Bugs/Issue426.cs`

```csharp
public class ClientWeb
{
    public ClientWeb( object appSettings, bool throwOnError = true ) { }

    [ExtractedResultLoggingAspect]
    public void Foo() { }
}

public class ExtractedResultLoggingAspect : OverrideMethodAspect
{
    [IntroduceDependency]
    private readonly ILogger _logger;
    // ...
}
```

## Generated Code (Issue)

```csharp
public ClientWeb(object appSettings, bool throwOnError = true, [AspectGenerated] ILogger<ClientWeb> logger = default)
```

The parameter `logger` is a non-nullable type `ILogger<ClientWeb>` but has a default value of `default` (which is `null`), causing **CS8625**: *Cannot convert null literal to non-nullable reference type*.

## Expected

The generated parameter should be nullable or use `null!`:

```csharp
// Option 1: Make parameter nullable
public ClientWeb(object appSettings, bool throwOnError = true, [AspectGenerated] ILogger<ClientWeb>? logger = default)

// Option 2: Use null!
public ClientWeb(object appSettings, bool throwOnError = true, [AspectGenerated] ILogger<ClientWeb> logger = default!)
```

### LINKED ISSUE #1232: Enhance nullability handling in type system API
## Summary

Improve nullability handling in the type system API to provide better control over nullable reference types and value types.

### New methods on IType

- StripNullabilityAnnotation() - Strips the nullability annotation from a type, returning an unannotated version (where IsNullable is null)

### New methods on derived type interfaces

- INamedType.StripNullabilityAnnotation() - Returns INamedType
- IArrayType.StripNullabilityAnnotation() - Returns IArrayType  
- IDynamicType.StripNullabilityAnnotation() - Returns IDynamicType
- ITypeParameter.StripNullabilityAnnotation() - Returns ITypeParameter

### New expression factory method

- ExpressionFactory.WithNullForgivingOperator(IExpression, bool force) - Adds the null-forgiving operator (!) to an expression

### Internal changes

- DeclarationFactory.MakeNullableType() - Now accepts bool? isNullable to support three-state nullability (nullable, non-nullable, unannotated)

## Breaking change

The elvis operator "?." in templates is no longer simplified out for non-nullable reference types. It is only simplified for non-nullable value types. This ensures proper nullability semantics in generated code.

## Related

- Documentation issue: https://github.com/metalama/Metalama.Documentation/issues/238

gfraiteur:
## Summary

- Fix empty HTML output when test has filtered suppressions
- Add `IType.StripNullabilityAnnotation()` method to strip nullability annotations without marking as non-nullable
- Add `IMetaTarget.Expression` property providing unified access to fields, properties, and parameters as `IExpression`
- Add `TypedConstant.Create(IField)` to reference const fields instead of literal values
- Add `TypedConstant.Default(IType, bool)` overload with null-forgiving operator support
- Add `HasNullForgivingOperator` property to `TypedConstant`
- Report warnings by default in test framework
- The elvis operator `?.` in templates is no longer simplified out for non-nullable reference types

## Breaking Changes

The elvis operator `?.` in templates is no longer simplified out for non-nullable reference types. It is only simplified for non-nullable value types.

## Issues Fixed

- Fixes #1226 - Test framework: WriteInputHtml produces empty file for some tests
- Fixes #1232 - Enhance nullability handling in type system API

### DIFF
--- a/.claude/agents/api-docs-reviewer.md
+++ b/.claude/agents/api-docs-reviewer.md
@@ -94,6 +94,9 @@ Based on [Microsoft Style Guide for Developer Content](https://learn.microsoft.c
 - Use vague phrases like "This method does something"
 - Include internal implementation details
 - Write long code examples - keep them in conceptual docs instead
+- Explain concepts C# developers already know (e.g., what a null-forgiving operator does, what default values are)
+- Document benefits or use cases that are obvious from the API name and signature
+- Add verbose remarks explaining "when to use" for straightforward APIs
 
 ## Workflow
 
--- a/.claude/agents/git-workflow.md
+++ b/.claude/agents/git-workflow.md
@@ -47,7 +47,16 @@ Examples:
      - Add the `breaking` label to both the PR and the issue
 
 4. Create PR targeting `develop/YYYY.N` (NOT the default branch)
-5. Include `Closes #XXXX` or `Fixes #XXXX` in PR body for issue linking
+5. PR body format:
+   - Summary section with bullet points describing key changes
+   - Breaking Changes section (if applicable) listing new interface members or behavioral changes
+   - **NO test plan section** - tests are verified through CI
+   - Issues Fixed section: List ALL issues with one-line descriptions, e.g.:
+     ```
+     ## Issues Fixed
+     - #1226 - Test framework: WriteInputHtml produces empty file for some tests
+     - #1232 - Enhance nullability handling in type system API
+     ```
 
 ### After Creating a PR - Checklist
 
--- a/.claude/agents/git-workflow.md
+++ b/.claude/agents/git-workflow.md
@@ -6,12 +6,12 @@ model: sonnet
 
 You are an expert Git workflow specialist with deep knowledge of branching strategies, commit best practices, and CI/CD integration.
 
-## Responsibilities
+## Repository Info
 
-1. **Branch Creation**: Create branches following naming conventions
-2. **Commits**: Craft concise, well-formatted commit messages
-3. **Pull Requests**: Create PRs targeting the correct merge branch
-4. **Build Scheduling**: Trigger and check TeamCity builds
+- **Default branch**: `release/2025.1`
+- **Development branches**: `develop/YYYY.N` (e.g., `develop/2026.0`)
+- **Organization**: `metalama`
+- **Repository**: `Metalama`
 
 ## Branch Naming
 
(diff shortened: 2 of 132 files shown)


gfraiteur:
Closes #1226

gfraiteur:
Released in [Metalama 2026.0.8-rc](https://github.com/metalama/Metalama/releases/tag/release/2026.0.8-rc).

— Claude

PR #1251, opened 2025-12-13
Escape C# keyword names with @ prefix in generated code (#812)

### LINKED ISSUE #812: Names of introduced declarations should be @ escaped when necessary
Consider this introduction, whose name is a C# keyword:

```c#
builder.Advice.IntroduceField(builder.Target, "const", typeof(int));
```

It generates this invalid code:

```c#
private int const;
```

The identifier should be escaped using `@`:

```c#
private int @const;
```

gfraiteur:
## Summary
- Added `SafeIdentifier` and `SafeIdentifierName` methods in `SyntaxFactoryEx.cs` that escape C# keywords with `@` prefix when necessary
- Added `WellKnownIdentifier` and `WellKnownIdentifierName` methods for generated names that don't need escaping (like `__o_p__` prefixed names)
- Created `LAMA0850` analyzer (`UnsafeIdentifierAnalyzer.cs`) to detect unsafe `Identifier`/`IdentifierName` calls and guide developers to use safe alternatives
- Fixed all `LAMA0850` warnings across the codebase by using the appropriate safe methods
- Fixed `LAMA0832` warnings in EventRaise substitution files by using `WithOptionalLeadingTrivia`/`WithOptionalTrailingTrivia` with context options
- Added `SyntaxGenerationContext` overloads for `WithOptionalLeadingTrivia`, `WithOptionalTrailingTrivia`, and `WithTriviaFromIfNecessary` for simpler code
- Created Bug812 test files to verify keyword escaping works correctly for fields, properties, methods, events, parameters, and nested types

Closes #812

-- Claude

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerInjectionStep.AuxiliaryMemberFactory.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerInjectionStep.AuxiliaryMemberFactory.cs
@@ -102,7 +102,7 @@ internal sealed partial class LinkerInjectionStep
                                     Argument(
                                         null,
                                         p.Modifiers.FirstOrDefault( m => !m.IsKind( SyntaxKind.ParamsKeyword ) ),
-                                        SyntaxFactoryEx.SafeIdentifierName( p.Identifier.ValueText ) ) ) ) ) ),
+                                        SyntaxFactoryEx.WellKnownIdentifierName( p.Identifier ) ) ) ) ) ),
                 Block(),
                 null,
                 default );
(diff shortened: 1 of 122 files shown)


Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerInjectionStep.AuxiliaryMemberFactory.cs
```diff
@@ -102,7 +102,7 @@ public ConstructorDeclarationSyntax GetAuxiliarySourceConstructor( IFullRef<ICon
                                     Argument(
                                         null,
                                         p.Modifiers.FirstOrDefault( m => !m.IsKind( SyntaxKind.ParamsKeyword ) ),
-                                        IdentifierName( p.Identifier.ValueText ) ) ) ) ) ),
+                                        SyntaxFactoryEx.SafeIdentifierName( p.Identifier.ValueText ) ) ) ) ) ),
```
addabis: This is a round-trip that is not necessary.

addabis (approved):

Issue #828, opened 2025-07-10
Field pseudo accessors should have IsImplicitDeclaration=false


PostSharpBot:
Changing this to false violates the assertion in Ref<T>.PseudoAccessor and if this is disabled a lot of tests cease to work properly (there is probably some logic in transformations/linker depending on this value).

PostSharpBot:
The ticket was imported from [TP-32474](https://postsharp.tpondemand.com/entity/32474).
Original creator: d***l [at] postsharp.net
Original creation time: 2023-01-06T15:47:22Z
Original severity: Annoying
Original business value: Low
Original source: Internal

PostSharpBot:
Original ticket contained comments:

2023-03-15T17:08:13Z DB:
We forgot to note the reason for this change, but I think is the following inconsistency:

Field has both accessors "IsImplicitlyDeclared=true" but when promoted, the accessors will become "IsImplicitlyDeclared=false". This means that the aspect cannot depend on the value of this property.

Additionally, backing fields are now visible in the code model and have IsImplicitlyDeclared=true, which differentiates them from normal fields.

Pseudo declarations are concept that does not exist in symbol model and is synthesized. Field's getter DOES exist semantically the same way as get-only property setter does exists.

gfraiteur:
# Issue #828: Field Pseudo-Accessors and IntroduceAttribute

## Summary
Fixed issues with pseudo members (field pseudo-accessors) in Metalama:
1. `IntroduceAttribute` advice now correctly rejects pseudo members
2. Multicast aspects no longer flow from fields to their pseudo-accessors (regression fix)
3. Exposed `ImplementationKind` property to public API for identifying pseudo members

## Breaking Changes

### ⚠️ `IsImplicitlyDeclared` Behavior Change
**Impact:** Pseudo members (field pseudo-accessors, pseudo parameters)

The `IsImplicitlyDeclared` property behavior for pseudo members has changed in a recent PR.

**Note:** This breaking change was implemented in a separate PR and is documented here for completeness.

## New Public API

### `IDeclaration.ImplementationKind` Property

**Location:** `Metalama.Framework.Code.IDeclaration`

```csharp
/// <summary>
/// Gets the implementation kind of this declaration, indicating whether it is backed by a Roslyn symbol,
/// a pseudo member (Metalama abstraction), or introduced by an aspect.
/// </summary>
DeclarationImplementationKind ImplementationKind { get; }
```

**Enum:** `DeclarationImplementationKind`
[...]

PR #1254, opened 2025-12-18
Revert #828: Move DeclarationImplementationKind to internal API

### LINKED ISSUE #1253: Reverting #828 - Field pseudo accessors should have IsImplicitDeclaration=false
It is only logical that field accessors have `IsImplicitDeclaration == true`.

The initial ticket #828 says that this generates inconsistencies while overriding, i.e. an implicitly-declared (pseudo) field accessor becomes an explicitly declared property accessor. But this is also a logical effect of the aspect, and should not be worked around by making an illogical model of the input.

Will be reverted:

- Pseudo declarations will always have IsImplicitDeclaration == true.
- ImplementationKind will be moved back to internal
- Eligibility rule MustNotBePseudo will be removed
- Eligibility rule for IntroduceAttribute will need to be explicit - which cases are legit on implicitly-defined declarations (default construct, field-backing property, return-value parameter), which are not (field accessors - others?)

gfraiteur:
## Summary
- Moved DeclarationImplementationKind from public API back to internal API
- This is a partial revert of #828, making field pseudo accessors properly marked as implicit declarations
- The enum is now located in Metalama.Framework.Engine instead of Metalama.Framework

## Breaking Changes
- DeclarationImplementationKind enum is no longer part of the public API
- The enum has been moved from Metalama.Framework.Code to Metalama.Framework.Engine.CodeModel.Abstractions

## Issues Fixed
- #1253 - Reverting #828 - Field pseudo accessors should have IsImplicitDeclaration=false

— Claude for Gael


gfraiteur:
## Response to Copilot PR Review

I've reviewed all the Copilot comments. Here's the status:

**Fixed:**
- ✅ **Parameter filtering inconsistency** (line 370 in MulticastImplementation.cs): Added missing \!p.IsImplicitlyDeclared\ check for consistency with lines 448, 501, and 523. All tests pass. (Commit 88a63a3984)

**Copilot Comments That Are Incorrect:**
- ❌ **Comments about 'value parameter' being wrong** (comments on AdviceFactory.cs#393, EligibilityRuleFactory.cs#192-197): The comments are actually CORRECT. The code does allow the \alue\ parameter of property setters through the second condition that checks if the declaring member is a property setter method.

- ❌ **Logic error in AdviceFactory.cs#393-404**: The logic is correct. Line 394 filters to parameters of explicitly-declared members, then line 396 allows return parameters and line 401 allows any parameter of property setters (the \alue\ parameter).

- ❌ **Unreachable code in DockerBuild.ps1#888**: This comment is outdated. The current code at line 887 is \\ = \\, not the problematic code mentioned.

**Questions:**
- ❓ **DockerBuild.ps1 behavior change** (line 334): The parameter changed from \-NoClean\ (opt-out, cleaned by default) to \-Clean\ (opt-in, only cleans when requested). Was this intentional? This could affect existing workflows.

-- Claude

PR #1294, opened 2026-01-27
Enable contracts on receiver parameters of extension blocks (#1127)

### LINKED ISSUE #1127: C# 14 extension members: contract to receiver parameter
Not trivial because the advice must be added to the members and not the parameter itself, because the parameter is shared among many members, and members might be added after the contract is added.

The envisioned approach is to make sure the advise receive the explicit member, because the receiver parameter itself does not have a member.

gfraiteur:
## Summary
- Contracts applied to receiver parameters of C# 14 extension blocks are now fanned out to all instance members (methods, property accessors)
- Static members and operators are correctly skipped
- Added comprehensive tests for methods, properties, operators, static members, multiple contracts, and ref receivers

## Issues Fixed
- #1127 - Enable contracts on receiver parameters of extension blocks

— Claude for gael


Metalama.Framework/src/Metalama.Framework/Eligibility/EligibilityRuleFactory.Contracts.cs
```diff
@@ -150,6 +187,34 @@ static void AddCommonReturnParameterRules( IEligibilityBuilder<IParameter> param
...
+                        p => p.AddRule( regularParameterEligibilityBoth ) ) );
+
+            var parameterEligibilityDefault =
+                CreateRule<IParameter>(
+                    parameter => parameter.MustSatisfyAny(
+                        p => p.AddRule( receiverParameterEligibilityDefault ),
+                        p => p.AddRule( regularParameterEligibilityDefault ) ) );
+
             _contractEligibilityBoth = CreateRule<IDeclaration>(
                 d =>
```
gfraiteur, line 216: Additionally, the conditions might be merged to have less duplication: for instance `In` contracts requires non-out parameter, then the conditions depend on whether the parameter is a receiver or not.

Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerInjectionStep.cs
```diff
@@ -674,11 +680,71 @@ var syntaxGenerationContext
                     break;
                 }
 
+            case IExtensionBlock extensionBlock:
+                {
+                    // For extension blocks, statements are generated per-method (each with a ContextDeclaration
+                    // that is contained in the target method). Route using standard IsContainedIn filtering.
```
gfraiteur: This algorithm is very inefficient - O(N^2) if N is the number of extension members. It would be better to do a two-phase algorithm: first insertedStatements by member (use GetClosestContainingMember) into a MultiDictionary, then access the MultiDictionary by key.

Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerInjectionStep.TransformationCollection.cs
```diff
@@ -619,10 +620,25 @@ private static IEnumerable<InsertedStatement> OrderInputContractStatements( IEnu
                 // Makes sure that the order is not changed when override is added in the middle of aspects that insert statements.
                 statements
                     .OrderBy(
```
gfraiteur: A lot of duplicate code between OrderInputContractStatements and OrderOutputContractStatements. Can this be unified?

PR #1318, opened 2026-02-06
Report LAMA0288 for compile-time method calls in run-time conditional blocks (#829)

### LINKED ISSUE #829: Run-time variable in compile-time loop acts odd
Consider this code ([Try link](https://try.metalama.net/#N4IghiBcIK4M4EsB2BzABAWQKYBcwBswBbMAOgDEAnYrAdwHtKBrUgQTgAcsBjHOAbgA6lQUnjJ02PIRIVqROoxYBhegBMspAMoBPJHgAeAIRgJ8GygOGjxqNLrg4sRUgBlkARyEixiOw6cXABUsAxxvUWskJBpOMG4sNAARem5SIPotHEoJKOAotEK0ZCdKGPw0bkI4ODQMrJzUVhxshAAjGCc0SDqdLnYuXgKi-J8i8bQAbQBJfUp1GASACjQAdQALLCQAUQMER1qAXjQAeQA3LEocjQawJxQdUnPL66wAGjQAORo0Y8EQeq1VD-NAASgAusMJmgODA2vgENw0I5GuhZtkFgk1IDUUtQVCJqNRNDoWcwJRkUCUCYzBZfmgFHhSKoiBwzFggggFCskHR7FSaeZLniwREkCSJiiJIKLGwOFwkGoVoyyEFyShcOk+ppvgpRQSJVLUDLLnKFUq0CCQaCxQaSWSKQAzBBYcxHBm4VXqzVBbUUF1u1iKgAK8y4lBwLrgpA2lywK0dvwAfGgAISO0jTODTVkI7gIHD4HRJHiEShYNRoABkVbTGazWTuiLB6Xo7kceNtYwlRQdxXpAAYu8Se4VHYwsPF1mgVn3na7K8g0PO3WC7dCieLRxMEImVgg0CmB2vu9uRuuz5TUSbKGathb-h9rWLL0UAL5RC+GgWmIW31jyveCYBmopC6okNqfqel4IAA1LBL7bh+PhfkU5Y4DAZRXtKv6yjiEidheyEjoUxFIG+IBvCAHBQCAMQKHECTJKkrYNLkPibuMkz4agkLQSUlzlJU1S1OQ9D0ASnEkhwORkl0JRoAA+gYw5btCsLws2RooAA-GgACaaDAGgGrhMiuD8GgZHjGRFFvkAA)):

```c#
internal class ToStringAttribute : TypeAspect
{
    [Introduce( WhenExists = OverrideStrategy.Override, Name = "ToString" )]
    public string IntroducedToString()
    {
        var stringBuilder = meta.CompileTime( new StringBuilder() );
        stringBuilder.Append( meta.Target.Type.Name );
        stringBuilder.Append( " " );

        var fields = meta.Target.Type.FieldsAndProperties.Where( f => !f.IsImplicitlyDeclared && !f.IsStatic ).ToList();

        var i = 0;

        foreach ( var field in fields )
        {
            if ( i > 0 )
            {
                stringBuilder.Append( ", " );
            }

            stringBuilder.Append( field.Name );

            i++;
        }

        return stringBuilder.ToString();
    }
}
```

The `i` variable should be initialized as compile-time (i.e. `var i = meta.CompileTime( 0 );`). But when it isn't, the transformed code for the `foreach` is an odd mix on run-time and compile-time code. In particular, the final string which should be e.g. `"Foo _x, Y"` is actually `"Foo , _x, Y"`.

This should probably be an error.

PostSharpAgent:
## Summary

- Adds a new diagnostic **LAMA0288** that reports an error when a compile-time method call with side effects (used as a statement) appears inside a run-time conditional block.
- This addresses the issue where a compile-time `StringBuilder.Append()` inside a run-time `if (i > 0)` block would produce incorrect mixed compile-time/run-time code without any error.
- The diagnostic detects invocations on compile-time local variables declared outside the run-time conditional block.

## Changes

- `TemplateAnnotator.cs`: Added `IsMethodCallOnCompileTimeLocalVariable()` method and diagnostic reporting in `VisitExpressionStatement()`.
- `TemplatingDiagnosticDescriptors.cs`: Added LAMA0288 diagnostic definition.
- New test: `RunTimeVariableInCompileTimeLoop.cs` — reproduces the original issue.
- Updated: `RunTimeIfCondition.cs.html` — updated expected HTML output to include the new diagnostic.

## Test plan

- [x] Regression test `RunTimeVariableInCompileTimeLoop` passes (error correctly reported)
- [x] All 1998 aspect tests pass (net8.0)
- [x] All 312 template tests pass (net8.0)
- [x] All 1528 unit tests pass (net8.0)
- [x] All 178 linker tests pass (net8.0)

Closes #829

— Claude for @PostSharpAgent


Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateAnnotator.cs
```diff
@@ -2164,6 +2187,24 @@ public override SyntaxNode VisitExpressionStatement( ExpressionStatementSyntax n
...
+        // A compile-time method call on a compile-time local variable used as a statement is
+        // assumed to have side effects that mutate the variable's state. Such calls cannot be
+        // conditionally executed based on a run-time condition because the compile-time state
+        // would be incorrect. For example, stringBuilder.Append(", ") inside if (i > 0) where
+        // stringBuilder is a compile-time variable and i is a run-time variable.
+        // We exclude compile-time methods that return run-time values (like IMethod.Invoke())
+        // since those are template expansion patterns that generate run-time code.
+        if ( statementScope == CompileTimeOnly
+             && !expressionScope.IsCompileTimeMemberReturningRunTimeValue()
+             && this._currentScopeContext.IsRunTimeConditionalBlock
```
gfraiteur: Here you're just looking in the immediate current block, but this should apply if _any_ parent is IsRunTimeConditionalBlock

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateAnnotator.cs
```diff
@@ -1918,6 +1918,29 @@ public override SyntaxNode VisitEventDeclaration( EventDeclarationSyntax node )
...
+        {
+            return false;
+        }
+
+        // Check if the receiver is a compile-time local variable declared outside the run-time conditional block.
+        var receiverSymbol = this._syntaxTreeAnnotationMap.GetSymbol( memberAccess.Expression );
+
+        return receiverSymbol is { Kind: SymbolKind.Local } and ILocalSymbol local
+               && this.GetNodeScope( memberAccess.Expression ).GetExpressionExecutionScope() == CompileTimeOnly
+               && this._currentScopeContext.IsRunTimeConditionalBlock
```
gfraiteur: `this._currentScopeContext.IsRunTimeConditionalBlock` seems duplicate with the logic in the caller and actually not relevant to this method.

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateAnnotator.cs
PostSharpAgent: Fixed. Removed the `IsRunTimeConditionalBlock` and `RunTimeConditionalBlockVariables` checks from `IsMethodCallOnCompileTimeLocalVariable`. The method now only checks whether the expression is an invocation on a compile-time local variable, and returns the receiver local via an `out` parameter. The run-time conditional block checks are handled entirely in the caller (`VisitExpressionStatement`).

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateAnnotator.cs
PostSharpAgent: Fixed. The `IsRunTimeConditionalBlock` check in the caller (`VisitExpressionStatement`) already works for any ancestor because the flag is propagated through all child scope contexts — once set to `true` by `RunTimeConditional()`, it stays `true` through all nested `CompileTimeOnly()`, `RunTimeOrCompileTime()`, `RunTimePreferred()`, `BreakOrContinue()`, and `ForbidDynamic()` calls.

I've also added a nested test case (`RunTimeVariableInCompileTimeLoop_Nested`) that verifies the diagnostic is correctly raised when a compile-time method call is inside a compile-time `foreach` loop that is itself nested inside a run-time `if` block.

Additionally, I discovered that the previous check was producing false positives on sub-template invocations (e.g., `templates.OnInvalidEnumValue(value, context)` in `Metalama.Patterns.Contracts`). These are compile-time method calls that generate run-time code rather than causing compile-time side effects. I added an `IsSubtemplateInvocation()` exclusion to handle this case, and the full `Build.ps1 build` now succeeds.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateAnnotator.cs
```diff
@@ -1918,6 +1918,55 @@ public override SyntaxNode VisitEventDeclaration( EventDeclarationSyntax node )
 
     private static bool IsMutatingUnaryOperator( SyntaxToken token ) => token.Kind() is SyntaxKind.PlusPlusToken or SyntaxKind.MinusMinusToken;
 
+    /// <summary>
+    /// Determines whether the expression is an invocation of a method on a compile-time local variable.
+    /// For example, <c>stringBuilder.Append(", ")</c> where <c>stringBuilder</c> is a compile-time local.
+    /// If the receiver is a local variable, it is returned via <paramref name="receiverLocal"/>.
+    /// </summary>
+    private bool IsMethodCallOnCompileTimeLocalVariable( ExpressionSyntax expression, out ILocalSymbol? receiverLocal )
```
gfraiteur: There should be a top-level method "HasCompileTimeSideEffect" in a helper class CompileTimeSideEffectDetector. The responsibility of this method will be to determine if an expression has a side effect on the compile-time side, using heuristics.  Basically, it should detect any side effect that it visible out of the top-level run-time conditional block. This is not an easy task. First, you need to track/propagate this block in the context (now you just have a flag). Then, you need to check that the statement possibly assigns or mutates a compile-time variable declared _outside_ of this context. Here you can use heuristics: assignments, ref/out arguments, or invoking methods without other effect (this is the case with stringBuilder.Append). You should try harder creating a list of cases of legit / illegit cases, and isolate this logic in CompileTimeSideEffectDetector. Prepare many unit tests, commit and even them even, together with their expected output, even before you try to implement the changes, so I can check your thinking. Keep in mind that the key thing we're trying to capture/detect is the modification of compile-time state declared outside of the top-level run-time conditional block.

PostSharpAgent: Addressed. Created `CompileTimeSideEffectDetector` with a `HasCompileTimeSideEffect` method that uses heuristics to detect compile-time side effects in run-time conditional blocks. The detector:

- Excludes sub-template invocations, `meta` code-generating methods (`InsertComment`, `InsertStatement`, `InvokeTemplate`, `Return`), assignments/increments (already covered by LAMA0108), and compile-time members returning run-time values.
- For invocations with a `MemberAccessExpression`: checks if the receiver is compile-time, and if it's a local, checks if it was declared inside the run-time conditional block.
- Simple name calls (no explicit receiver) are not flagged.

Also wrote 10 test cases (7 illegitimate, 3 legitimate) committed before the implementation.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateAnnotator.cs
PostSharpAgent: Addressed. The new `CompileTimeSideEffectDetector` receives the `RunTimeConditionalBlockVariables` from the top-level run-time conditional block context, which is already propagated to all child scopes via `ScopeContext.RunTimeConditional()`. The detector doesn't check parent blocks itself — the caller in `VisitExpressionStatement` checks `IsRunTimeConditionalBlock` which already reflects the full scope chain.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateAnnotator.cs
PostSharpAgent: Addressed. The `IsRunTimeConditionalBlock` check is now only in the caller (`VisitExpressionStatement`), and the detector method focuses on determining whether the expression itself has side effects. The detector receives `runTimeConditionalBlockVariables` (which is `null` if not in a run-time conditional block) but does not duplicate the conditional block check.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateAnnotator.cs
gfraiteur: I'm still missing test coverage for the following scenarios:
- the compile-time variable is defined within the run-time conditional blocks
- nested run-time conditional blocks with compile-time variables defined at different levels.

Also, improve processing of LAMA0108 and cover case where the variable is modified by `out` or `ref` argument passing.

PostSharpAgent: Addressed. Added 5 new test cases covering the requested scenarios:

**Nested run-time conditional blocks with variables at different levels:**
- `CompileTimeSideEffect_NestedRunTimeConditionals` — compile-time variable declared in outer run-time conditional block, method call in inner nested run-time conditional block. Correctly reports LAMA0288 because the inner block creates a fresh `RunTimeConditionalBlockVariables` list that does not contain the outer block's variables.
- `CompileTimeSideEffect_NestedRunTimeConditionals_InnerOnly` — compile-time variable declared AND used inside the same inner run-time conditional block. No error (legitimate).

**ref/out argument passing (LAMA0108 coverage):**
- `CompileTimeSideEffect_RefArgument` — passing a compile-time variable (declared outside the run-time conditional block) as a `ref` argument. Reports LAMA0108.
- `CompileTimeSideEffect_OutArgument` — passing a compile-time variable (declared outside the run-time conditional block) as an `out` argument. Reports LAMA0108.
- `CompileTimeSideEffect_RefArgument_InnerLocal` — passing a compile-time variable (declared inside the run-time conditional block) as a `ref` argument. No error (legitimate).

Note: LAMA0108 for `ref`/`out` arguments was already implemented in `VisitInvocationExpression` (line 1060-1073 of `TemplateAnnotator.cs`) via `CheckForMutatingCompileTimeExpressionInRunTimeConditionalBlock`. The existing logic correctly handles both outer locals (reports error) and inner locals (allows them). These new tests provide aspect-level test coverage for this existing behavior.

### Test results (net8.0):
- 2014 aspect tests: 0 failures (1982 passed, 32 skipped)
- 312 template tests: 0 failures
- 1531 unit tests: 0 failures (1525 passed, 6 skipped)
- `Build.ps1 build`: success

— Claude for @PostSharpAgent

gfraiteur (approved):

PR #1322, opened 2026-02-08
Fix trivia processing for deconstructed get-only auto properties

### LINKED ISSUE #837: Trivia processing for deconstructed get-only auto properties
When get-only auto-property is overridden, it is changed to PrivatelyWriteableProperty, which is meant to add the missing set accessor.

However, since this is a builder that wraps a code property, it should not lose trivia (especially doc comments), which it now does.

Also, the type name and attributes are regenerated from code model, instead of taking the original syntax.

PostSharpAgent:
## Summary

Fixes #837

When Metalama transforms declarations (overriding get-only auto properties or promoting fields to properties), this PR ensures doc comment trivia is correctly placed on the public member.

### Doc comments stay on the property (not the backing field)

When a get-only auto property is overridden and deconstructed into a backing field + property, the doc comments correctly remain on the property — the public API member. The backing field is an implementation detail.

### Field override: doc comments transfer from field to property

When a field is overridden (promoted to a property), the doc comments from the original field declaration are now transferred to the generated property, since the property becomes the public API member. Previously, doc comments were lost during this transformation.

**Changed:** `LinkerInjectionStep.Rewriter.cs` — When an injected property replaces a field (`IReplaceMemberTransformation`), doc comment trivia is extracted from the original field declaration and applied to the injected property.

**Changed:** `TriviaHelper.cs` — Provides `GetDocumentationTrivia` and `WithDocumentationTrivia` helper methods for extracting and applying XML doc comment trivia.

## Test plan

- [x] `Auto_GetOnly_Trivias` regression test passes (doc comments stay on property)
- [x] `Trivias` field override regression test passes (doc comments transfer to property)
- [x] `MultipleDeclarators_Trivias` test updated and passes
- [x] All 1994 aspect tests pass (0 failures)
- [x] All 1526 unit tests pass
- [x] All 62 observability tests pass
- [x] `Build.ps1 test` succeeds
- [x] Review feedback addressed (round 1 and 2)
- [ ] TC CI build passes


Metalama.Framework/src/Metalama.Framework.Engine/AdviceImpl/Override/OverridePropertyBaseTransformation.cs
```diff
@@ -100,6 +106,22 @@ protected IEnumerable<InjectedMember> GetInjectedMembersImpl(
...
+    private static TypeSyntax? GetSourcePropertyType( IProperty property )
+    {
+        if ( property.DeclarationKind == DeclarationKind.Property
+             && property is IDeclarationImpl { DeclaringSyntaxReferences: { Length: > 0 } syntaxRefs } )
+        {
+            var syntax = syntaxRefs[0].GetSyntax();
+
+            if ( syntax.IsKind( SyntaxKind.PropertyDeclaration ) && syntax is PropertyDeclarationSyntax propertyDeclaration )
+            {
+                return propertyDeclaration.Type.WithoutTrivia();
```
gfraiteur: This is a smell. The syntax should not be copied: if the propertyDeclaration is in another syntax tree, it will be missing namespaces. There is probably no value in this whole method.

Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerRewritingDriver.Properties.cs
```diff
@@ -59,13 +59,47 @@ private IReadOnlyList<MemberDeclarationSyntax> RewriteProperty(
...
+                        propertyDeclaration.Initializer,
+                        FilterAttributeListsForTarget( propertyDeclaration.AttributeLists, SyntaxKind.FieldKeyword, false, false ),
+                        symbol,
+                        generationContext );
+
+                    // Transfer documentation comment trivia from the property to the backing field,
+                    // so that doc comments stay above the first emitted member.
+                    var leadingTrivia = propertyDeclaration.GetLeadingTrivia();
+
+                    if ( leadingTrivia.ShouldBePreserved( this.SyntaxGenerationOptions ) )
```
gfraiteur: The trivia-processing logic should be in a helper method. Check if there is already a helper method or type doing something similar and, if not, define a new type TriviaHelper

gfraiteur (changes requested):

gfraiteur:
@PostSharpAgent Work on the feedback

Metalama.Framework/src/Metalama.Framework.Engine/AdviceImpl/Override/OverridePropertyBaseTransformation.cs
PostSharpAgent: Agreed — removed the `GetSourcePropertyType` method entirely. The override property now uses `context.SyntaxGenerator.PropertyType()` as before (producing `global::System.Int32`). Reverted the 4 `.t.cs` test expectations accordingly.

Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerRewritingDriver.Properties.cs
PostSharpAgent: Extracted the trivia-processing logic into a new \ static class in \. The method \ takes a backing field and property declaration, filters doc comment trivia from the property's leading trivia, and transfers it to the backing field. No existing helper was doing exactly this, so created the new type as suggested. Also updated the \ test in Observability which was affected by this trivia transfer.

PostSharpAgent: Extracted the trivia-processing logic into a new `TriviaHelper` static class in `Metalama.Framework.Engine.Linking`. The method `TransferDocumentationTrivia` takes a backing field and property declaration, filters doc comment trivia from the property leading trivia, and transfers it to the backing field. No existing helper was doing exactly this, so created the new type as suggested. Also updated the `InheritFromAbstractBase.t.cs` test in Observability which was affected by this trivia transfer.

gfraiteur (changes requested):
The issue was incorrectly formulated. The desired behavior is that the  XML doc trivia stays on the property, because the property is the public member and the field is an implementation detail.

In case of overriding a field, the trivia must move from the field to the property because the property becomes the public API.

the same when overriding a method or anything else - the public member keeps the trivia

I think the current implementation (before the fix) is correct. Add aspect tests to verify this, and fix if necessary.

PostSharpAgent:
## Review feedback addressed (round 2)

1. **Doc trivia stays on the property** (the public member) — Reverted the `TransferDocumentationTrivia` logic from `LinkerRewritingDriver.Properties.cs`. When a get-only auto property is deconstructed into backing field + property, the doc comments now stay on the property (not the backing field), since the property is the public API.

2. **Field override: doc trivia moves from field to property** — Added new logic in `LinkerInjectionStep.Rewriter.cs` to transfer doc comment trivia from the original field declaration to the promoted property when a field is overridden. This is the correct behavior because the property becomes the public member.

3. **New test `Trivias` for field overrides** — Verifies that doc comments transfer from overridden fields to their promoted properties. Also updated `MultipleDeclarators_Trivias.t.cs` to reflect the new (correct) behavior.

4. **Reverted `InheritFromAbstractBase.t.cs`** — The Observability test expectation is back to its original state (doc comment stays on the property).

### Verification:
- All 1994 aspect tests pass (0 failures)
- All 1526 unit tests pass
- All 62 observability tests pass
- `Build.ps1 build` succeeds with zero warnings
- `Build.ps1 test` passes (4 `AwaitsMainMethod_*` timeouts are transient and pass when run alone)

— Claude for @gfraiteur

gfraiteur (approved):

Metalama.Framework/src/Metalama.Framework.Engine/Linking/TriviaHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        if ( documentationTrivia.Count == 0 )
+        {
+            return member;
+        }
+
+        var existingTrivia = member.GetLeadingTrivia();
+
+        return member.WithRequiredLeadingTrivia( documentationTrivia.AddRange( existingTrivia ) );
+    }
```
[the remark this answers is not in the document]
gfraiteur: We won't go to this level of details since we don't have a reformatting-resistant test infrastructure.

Metalama.Framework/src/Metalama.Framework.Engine/Linking/TriviaHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            if ( trivia.IsKind( SyntaxKind.SingleLineDocumentationCommentTrivia )
+                 || trivia.IsKind( SyntaxKind.MultiLineDocumentationCommentTrivia ) )
+            {
+                docCommentTrivia.Add( trivia );
+            }
+        }
+
+        return new SyntaxTriviaList( docCommentTrivia );
+    }
```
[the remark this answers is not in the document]
gfraiteur: We won't go to this level of details since we don't have a reformatting-resistant test infrastructure.

PR #1327, opened 2026-02-08
Fix template type parameter name mapping after introduction rename (#766)

### LINKED ISSUE #766: Run-time template parameters (both type and normal) may not match the name of the type parameters when the template overrides an existing method
Template type parameters are mapped by name, but an introduction can change the name:

```
public class TestAspect1 : TypeAspect
{
    // This aspect changes the type parameter name.
    [Introduce(WhenExists = OverrideStrategy.Override)]
    public void Foo<U>()
    {
        meta.Proceed();
    }
}

public class TestAspect2 : TypeAspect
{
    public override void BuildAspect(IAspectBuilder<INamedType> builder)
    {
        if (builder.Target.Is(typeof(Base)))
        {
            builder.Advice.Override(
                builder.Target.Methods.OfName("Foo").Single(),
                nameof(FooTemplate));
        }
    }

    // The aspect uses the template type parameter it is aware of, i.e. how it is defined in the base type.
    [Template]
    public void FooTemplate<T>()
    {
        meta.Proceed();
    }
}

class Base
{
    public virtual void Foo<T>()
    {
    }
}

// <target>
[TestAspect1]
class Target1 : Base
{
    // TestAspect2 can override the base method.
}

// <target>
[TestAspect1]
[TestAspect2]
class Target2 : Base
{
    // TestAspect2 cannot override because TestAspect1 changed the type parameter name.
}
```

This makes runtime template parameters unusable for override templates.

PostSharpAgent:
## Summary

Fixes #766 - Template type parameter name mapping fails when an introduction renames type parameters.

When an `[Introduce]` aspect overrides a method and renames type parameters (e.g., `Foo<S>` → `Foo<U>`), subsequent override templates using the original names would fail. This PR fixes the mapping to use ordinal-based matching for type parameters and resolves names at expansion time through `ITemplateSyntaxFactory`.

### Implementation
- **`ITemplateSyntaxFactory.cs`**: Added `RunTimeTypeParameterIdentifier(string templateParameterName)` method that maps template type parameter names to correct target names at expansion time
- **`TemplateSyntaxFactoryImpl.cs`**: Implemented `RunTimeTypeParameterIdentifier` using `TemplateExpansionContext.TemplateGenericArguments` for ordinal-based mapping. Also updated `TypeOf()` to apply run-time type parameter name substitutions during expansion
- **`TemplateCompilerRewriter.cs`**: For run-time type parameters in Transform mode, generates a call to `RunTimeTypeParameterIdentifier` instead of embedding the template's type parameter name via `SyntaxGenerator.TypeOrNamespace()`
- **`TemplateMemberGenericContext.cs`**: Use ordinal matching among run-time type parameters (skipping compile-time ones) instead of name-based matching
- **`TemplateBindingHelper.cs`**: Use ordinal-only matching for type parameters; keep name-based matching with ordinal fallback for regular parameters
- **`TemplateExpansionContext.cs`**: Map template run-time type parameter names to target type parameters by ordinal position
- **`TemplateDriver.cs`**: Removed `TypeParameterRenamingRewriter` and `RenameTypeParametersIfNecessary` — renaming is no longer done by post-processing, it happens at expansion time through the factory

### Tests
- **`IntroduceOverrideRenamesTypeParameter`**: Tests type parameter renaming (base `S` → intro `U`, override template `T` → output `U`)
- **`IntroduceOverrideRenamesParameterAndTypeParameter`**: Tests both type parameter and regular parameter renaming
- **`IntroduceOverrideRenamesMixedTypeParameters`**: Tests mixed compile-time and run-time type parameters with 5 type params (3 run-time, 2 compile-time)

## Checklist
- [x] Reproduce bug with regression test
- [x] Implement fix
- [x] Address all review feedback from @gfraiteur
- [x] All aspect tests pass (1990/1990)
- [x] All template tests pass (312/312)
- [x] `Build.ps1 build` passes with zero warnings
- [x] `Build.ps1 test` passes with zero failures and zero warnings

--- Claude for @PostSharpAgent

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Advising/TemplateBindingHelper.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Advising/TemplateBindingHelper.cs
@@ -472,6 +472,10 @@ internal static class TemplateBindingHelper
 
             case { OperatorKind: OperatorKind.None }:
                 // For non-operator methods, we match parameters by name.
+                // An introduction can rename parameters, so if name-based lookup fails,
+                // we fall back to matching by ordinal position among run-time parameters.
+                var runTimeParameterIndex = 0;
+
                 foreach ( var templateParameter in templateMethodSymbol.Parameters )
                 {
                     if ( template.TemplateClassMember.Parameters[templateParameter.Ordinal].IsCompileTime )
@@ -483,11 +487,18 @@ internal static class TemplateBindingHelper
 
                     if ( methodParameter == null )
                     {
-                        var parameterNames = string.Join( ", ", targetMethod.Parameters.SelectAsImmutableArray( p => "'" + p.Name + "'" ) );
+                        if ( runTimeParameterIndex < targetMethod.Parameters.Count )
+                        {
+                            methodParameter = targetMethod.Parameters[runTimeParameterIndex];
+                        }
+                        else
+                        {
+                            var parameterNames = string.Join( ", ", targetMethod.Parameters.SelectAsImmutableArray( p => "'" + p.Name + "'" ) );
 
-                        throw new InvalidTemplateSignatureException(
-                            MetalamaStringFormatter.Format(
-                                $"Cannot use the template '{templateMethodSymbol}' to override the method '{targetMethod}': the target method does not contain a parameter '{templateParameter.Name}'. Available parameters are: {parameterNames}." ) );
+                            throw new InvalidTemplateSignatureException(
+                                MetalamaStringFormatter.Format(
+                                    $"Cannot use the template '{templateMethodSymbol}' to override the method '{targetMethod}': the target method does not contain a parameter '{templateParameter.Name}'. Available parameters are: {parameterNames}." ) );
+                        }
                     }
 
                     if ( !VerifyTemplateType( templateParameter.Type, methodParameter.Type, template, targetMethod, arguments ) )
@@ -498,6 +509,7 @@ internal static class TemplateBindingHelper
                     }
 
                     AddParameter( methodParameter, templateParameter );
+                    runTimeParameterIndex++;
                 }
 
                 // Check that template generic parameters match the target.
@@ -508,11 +520,24 @@ internal static class TemplateBindingHelper
                         continue;
                     }
 
-                    var methodParameter = targetMethod.TypeParameters.SingleOrDefault( p => p.Name == templateParameter.Name )
-                                          ??
-                                          throw new InvalidTemplateSignatureException(
-                                              MetalamaStringFormatter.Format(
-                                                  $"Cannot use the template '{templateMethodSymbol}' to override the method '{targetMethod}': the target method does not contain a generic parameter '{templateParameter.Name}'." ) );
+                    // First try to match by name. If that fails, fall back to matching by ordinal position.
+                    // An introduction can rename type parameters, so the template may use the original name
+                    // while the target method uses the new name.
+                    var methodParameter = targetMethod.TypeParameters.SingleOrDefault( p => p.Name == templateParameter.Name );
+
+                    if ( methodParameter == null )
+                    {
+                        if ( templateParameter.Ordinal < targetMethod.TypeParameters.Count )
+                        {
+                            methodParameter = targetMethod.TypeParameters[templateParameter.Ordinal];
+                        }
+                        else
+                        {
+                            throw new InvalidTemplateSignatureException(
+                                MetalamaStringFormatter.Format(
+                                    $"Cannot use the template '{templateMethodSymbol}' to override the method '{targetMethod}': the target method does not contain a generic parameter '{templateParameter.Name}'." ) );
+                        }
+                    }
 
                     if ( !templateParameter.IsCompatibleWith( methodParameter ) )
                     {
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Advising/TemplateBindingHelper.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Advising/TemplateBindingHelper.cs
@@ -471,9 +471,8 @@ internal static class TemplateBindingHelper
                 break;
 
             case { OperatorKind: OperatorKind.None }:
-                // For non-operator methods, we match parameters by name.
-                // An introduction can rename parameters, so if name-based lookup fails,
-                // we fall back to matching by ordinal position among run-time parameters.
+                // For non-operator methods, we match parameters by name, falling back to ordinal position
+                // among run-time parameters when an introduction has renamed the parameter.
                 var runTimeParameterIndex = 0;
 
                 foreach ( var templateParameter in templateMethodSymbol.Parameters )
@@ -512,7 +511,9 @@ internal static class TemplateBindingHelper
                     runTimeParameterIndex++;
                 }
 
-                // Check that template generic parameters match the target.
+                // Check that template generic parameters match the target by ordinal position among run-time type parameters.
+                var runTimeTypeParameterIndex = 0;
+
                 foreach ( var templateParameter in templateMethodSymbol.TypeParameters )
                 {
                     if ( template.TemplateClassMember.TypeParameters[templateParameter.Ordinal].IsCompileTime )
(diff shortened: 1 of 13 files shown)


Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Generic/IntroduceOverrideRenamesTypeParameter.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Metalama.Framework.Tests.AspectTests.Tests.Aspects.Generic.IntroduceOverrideRenamesTypeParameter
+{
+    public class TestAspect1 : TypeAspect
+    {
+        // This aspect introduces/overrides the method changing the type parameter name from T to U.
+        [Introduce( WhenExists = OverrideStrategy.Override )]
+        public void Foo<U>()
+        {
+            meta.Proceed();
```
gfraiteur, line 21: Reference parameter U in the template body to proof it's correct e.g. 

Console.WriteLine( typeof(U))

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Generic/IntroduceOverrideRenamesTypeParameter.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                builder.With( builder.Target.Methods.OfName( "Foo" ).Single() ).Override( nameof(FooTemplate) );
+            }
+        }
+
+        // The template uses the original type parameter name T (as defined in Base).
+        [Template]
+        public void FooTemplate<T>()
+        {
+            meta.Proceed();
```
gfraiteur, line 40: Also reference T

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Generic/IntroduceOverrideRenamesTypeParameter.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    internal class Base
+    {
+        public virtual void Foo<T>()
+        {
```
gfraiteur, line 47: Use a different name than T, e.g. S

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Generic/IntroduceOverrideRenamesParameterAndTypeParameter.cs
```diff
@@ -0,0 +1,56 @@
...
+using Metalama.Framework.Code;
+
+[assembly: AspectOrder( AspectOrderDirection.CompileTime, typeof(Metalama.Framework.Tests.AspectTests.Tests.Aspects.Generic.IntroduceOverrideRenamesParameterAndTypeParameter.TestAspect1), typeof(Metalama.Framework.Tests.AspectTests.Tests.Aspects.Generic.IntroduceOverrideRenamesParameterAndTypeParameter.TestAspect2) )]
+
+namespace Metalama.Framework.Tests.AspectTests.Tests.Aspects.Generic.IntroduceOverrideRenamesParameterAndTypeParameter
+{
+    public class TestAspect1 : TypeAspect
+    {
+        // This aspect introduces/overrides the method, renaming both the type parameter (T -> U)
+        // and the regular parameter (value -> item).
```
gfraiteur, line 17: Same remarks as for IntroduceOverrideRenamesTypeParameter

Metalama.Framework/src/Metalama.Framework.Engine/Advising/TemplateBindingHelper.cs
```diff
@@ -498,6 +509,7 @@ _ when targetMethod.OperatorKind.GetCategory() is OperatorCategory.UnaryAssignme
                     }
 
                     AddParameter( methodParameter, templateParameter );
+                    runTimeParameterIndex++;
```
gfraiteur, line 511: This is incremented even for compile-time parameters.

Metalama.Framework/src/Metalama.Framework.Engine/Advising/TemplateBindingHelper.cs
```diff
@@ -472,6 +472,10 @@ _ when targetMethod.OperatorKind.GetCategory() is OperatorCategory.UnaryAssignme
 
             case { OperatorKind: OperatorKind.None }:
                 // For non-operator methods, we match parameters by name.
+                // An introduction can rename parameters, so if name-based lookup fails,
+                // we fall back to matching by ordinal position among run-time parameters.
+                var runTimeParameterIndex = 0;
```
gfraiteur, line 476: I don't think there should be a "preferred" and "fallback" mechanism. It's better to have a single mechanism, which is to match by index only. So let's completely remove the match by name.

gfraiteur (changes requested):

gfraiteur:
@PostSharpAgent Implement my feedback. Also add tests where there are BOTH run-time and compile-time type parameters, and not packed adjacently but e.g. `Type<T1, [COmpileTIme] T2, T3, [CompileTime] T4, T5>`

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Generic/IntroduceOverrideRenamesTypeParameter.cs
PostSharpAgent: Done. Both templates now reference `typeof(U)` and `typeof(T)` respectively in their body to prove type parameter substitution works correctly in the expanded output.

--- Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Generic/IntroduceOverrideRenamesTypeParameter.cs
PostSharpAgent: Done. TestAspect2's template now uses `Console.WriteLine(typeof(T))` which gets correctly substituted to `typeof(U)` in the expanded output.

--- Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Generic/IntroduceOverrideRenamesTypeParameter.cs
PostSharpAgent: Done. Base class now uses `S` as type parameter name: `public virtual void Foo<S>()`.

--- Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Generic/IntroduceOverrideRenamesParameterAndTypeParameter.cs
PostSharpAgent: Done. Applied the same changes: base uses `S`, templates reference `typeof(U)` and `typeof(T)`.

--- Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Advising/TemplateBindingHelper.cs
PostSharpAgent: Fixed. The `runTimeParameterIndex` is now only incremented for run-time parameters (compile-time parameters are skipped with `continue`). Same for `runTimeTypeParameterIndex` in the type parameter section.

--- Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Advising/TemplateBindingHelper.cs
PostSharpAgent: For **type parameters**, I've implemented ordinal-only matching as requested (skipping compile-time type parameters).

For **regular parameters**, I kept name-based matching with ordinal fallback. Switching to ordinal-only breaks existing tests that depend on name-based parameter mapping:
- `Overrides/Methods/ParameterMapping`: template `InvertedParameters(int y, string x)` overriding `Method_InvertedParameters(string x, int y)` — ordinal matching causes type mismatch (int vs string)
- `Subtemplates/DynamicParameter_Virtual`: template `OverrideMethod(dynamic x, dynamic y)` overriding `Method2(int y, int x)` — ordinal matching swaps the parameter values

These tests demonstrate intentional design where template parameters map to target parameters by name regardless of position. The ordinal fallback only activates when an introduction has renamed a parameter.

Please let me know if you'd like me to change this approach.

--- Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateDriver.cs
```diff
@@ -85,5 +92,68 @@ public bool TryExpandDeclaration(
 
             return errorCountAfter == errorCountBefore;
         }
+
+        private static BlockSyntax RenameTypeParametersIfNecessary( BlockSyntax block, TemplateExpansionContext context )
+        {
```
gfraiteur: This should not be done this way -- not by rewriting. 

The way it should work is that the template code does not write the parameter name directly, but should use a mapping method in ITemplateSyntaxFactory, which relies on TemplateExpansionContext, which contains the mapping between run-time type parameter ordinals and their name in the current context. 

So the name if type parameters in the template itself is irrelevant: we always take the original names.

Any other design is cumbersome and brittle.

gfraiteur (changes requested):

gfraiteur:
@PostSharpAgent Address my feedback.

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateDriver.cs
PostSharpAgent: Done. Replaced the `TypeParameterRenamingRewriter` approach entirely with proper mapping through `ITemplateSyntaxFactory`:

1. **`ITemplateSyntaxFactory.RunTimeTypeParameterIdentifier(string)`** — new method that maps a template type parameter name to the correct target name at expansion time, using `TemplateExpansionContext.TemplateGenericArguments`
2. **`TemplateCompilerRewriter`** — for run-time type parameters in Transform mode, now emits a call to `RunTimeTypeParameterIdentifier("T")` instead of directly embedding the template's identifier via `SyntaxGenerator.TypeOrNamespace()`
3. **`TemplateSyntaxFactoryImpl.TypeOf()`** — now also applies run-time type parameter substitutions from the expansion context, so `typeof(T)` correctly resolves to `typeof(U)` during expansion
4. **`TemplateDriver`** — `RenameTypeParametersIfNecessary()` and `TypeParameterRenamingRewriter` removed entirely

The template type parameter names are now completely irrelevant — correct names come from the expansion context mapping.

--- Claude for @PostSharpAgent

gfraiteur (approved):

PR #1336, opened 2026-02-08
Fix CS9230: dynamic in interpolated strings (#580)

### LINKED ISSUE #580: T#: error CS9230: Cannot perform a dynamic invocation on an expression
Is caused by a change in Roslyn that started to report this error message in 4.12 or 4.13.

InterpolatedStringBuilders cannot receive dynamic, so the cast from dynamic to e.g. object must be made manually.

**INSERT ATTACHMENT #7031 (ConsoleApp81.zip)**

PostSharpAgent:
## Summary

Fix CS9230 error ("Cannot perform a dynamic invocation on an expression") that occurs when Metalama-generated code uses `dynamic` expressions inside interpolated strings on Roslyn 4.12+ (which uses `DefaultInterpolatedStringHandler`).

### Approach

The fix casts `dynamic` expressions to `object` before embedding them in interpolated strings. Type detection relies on the **compile-time object model** (not the template semantic model) to distinguish truly-dynamic runtime types from types that are only `dynamic` at the template level.

Two detection mechanisms:

1. **`DynamicLocalDeclaration`** records the compile-time object model type (`value.Type`) for each local variable declared via `meta.Proceed()` or similar. `FixInterpolationSyntax` checks this dictionary when a local variable identifier appears in an interpolation.

2. **`GetDynamicSyntax` type annotations** annotate `IExpression` values (like `parameter.Value`) with their actual type. `FixInterpolationSyntax` checks these annotations.

3. **InterpolatedStringBuilder path** (`InterpolatedStringUserExpression`): Separate code path for programmatically built interpolated strings also handles dynamic-to-object casting.

### Files changed

- `ITemplateSyntaxFactory.cs`: `FixInterpolationSyntax` signature (removed `isDynamic` parameter)
- `TemplateCompilerRewriter.cs`: Simplified `TransformInterpolation` — removed semantic model dynamic detection
- `TemplateSyntaxFactoryImpl.cs`: Added local variable type dictionary in `DynamicLocalDeclaration`, annotation-based + dictionary-based dynamic detection in `FixInterpolationSyntax`
- `InterpolatedStringUserExpression.cs`: Cast dynamic tokens to `object`
- 2 regression test files for Bug580
- Expected output files updated where `(object)` cast is correctly generated

## Test plan

- [x] Bug580_DynamicInInterpolatedString — dynamic local in interpolated string (with conflicting variable names in if/else)
- [x] Bug580_DynamicParameterInInterpolatedString — dynamic parameter.Value in interpolated string
- [x] All 1989 aspect tests pass (net8.0) + 1869 (net48)
- [x] All 1522 unit tests pass (net8.0) + 1477 (net48)
- [x] Build.ps1 build — zero warnings
- [x] Build.ps1 test — zero warnings, only pre-existing AspectTestRunnerTests timeouts
- [x] TC build 316499 (commit 722ae8d) — SUCCESS

Fixes #580


Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateSyntaxFactoryImpl.cs
```diff
@@ -244,6 +245,15 @@ public StatementSyntax DynamicLocalDeclaration(
                 throw new AssertionFailedException( "The expression should not be null." );
             }
 
+            // Track whether the variable will actually have type 'dynamic' at run time,
+            // so that FixInterpolationSyntax can add an (object) cast when needed to avoid CS9230.
+            var effectiveType = awaitResult ? value.Type.GetAsyncInfo().ResultType : value.Type;
+
+            if ( effectiveType.TypeKind == Code.TypeKind.Dynamic )
```
gfraiteur: The tracking of dynamic variables through a hashset of names seems like a hack. The root cause of the issue rather seems a missing type annotation in references of the local variable. The type analysis should be performed when we have access to the semantic model and not in TemplateSyntaxFactoryImpl which should assume that things upstream are fine.

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Bugs/Bug580_DynamicInInterpolatedString.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Metalama.Framework.Tests.AspectTests.Tests.Aspects.Bugs.Bug580_DynamicInInterpolatedString;
+
+internal class LogAttribute : OverrideMethodAspect
+{
+    public override dynamic? OverrideMethod()
+    {
+        // Using dynamic value (meta.Proceed() result, which is dynamic?) directly in an interpolated string.
+        // This should NOT produce CS9230 in the generated code.
+        var result = meta.Proceed();
```
gfraiteur, line 19: Create a method body with different blocks and conflicting variable names and you will see your hack does not work.

gfraiteur (changes requested):

gfraiteur:
@PostSharpAgent Work on my review.

PostSharpAgent:
## Progress Update

I've addressed the review feedback from @gfraiteur. Here's a summary of changes:

### What changed

1. **Removed the `_dynamicLocalVariableNames` HashSet hack** from `TemplateSyntaxFactoryImpl` — this was the name-tracking approach that wouldn't work with conflicting variable names across scopes.

2. **Proper fix: Pass type information from the template compiler to `FixInterpolationSyntax`** — The template compiler (in `TemplateCompilerRewriter.TransformInterpolation`) has access to the semantic model and knows when an interpolation expression is `dynamic`. It now passes `isDynamic: true` to `FixInterpolationSyntax` via a new optional parameter.

3. **Kept `GetDynamicSyntax` type annotation** — `GetDynamicSyntax` now annotates `ExpressionSyntax` values with the `dynamic` type (fixing the TODO comment). This ensures `FixInterpolationSyntax` can also detect dynamic expressions via annotation (e.g., for `IExpression` values like `parameter.Value`).

4. **Kept `InterpolatedStringUserExpression.cs` fix** — This handles a separate code path (the `InterpolatedStringBuilder` path).

5. **Improved test** — Updated `Bug580_DynamicInInterpolatedString` to include nested scopes with different variables to ensure the fix doesn't rely on name tracking.

### Current status
Building with `Build.ps1 build`. Will run tests after the build completes.

— Claude for @gfraiteur

PostSharpAgent:
## Progress update — session 2

**Addressing code review feedback from @gfraiteur.**

### Changes made

1. **Removed the `_dynamicLocalVariableNames` HashSet hack** — the reviewer was right that tracking variable names by string is fragile and fails with conflicting names across scopes.

2. **Proper approach: type detection at two levels:**
   - **At template compile time** (`TemplateCompilerRewriter.TransformInterpolation`): Detects `dynamic` type via the semantic model (`_syntaxTreeAnnotationMap.GetExpressionType()`) and passes `isDynamic: true` to `FixInterpolationSyntax`. This handles local variables like `var result = meta.Proceed()` which don't go through `GetDynamicSyntax`.
   - **At expansion time** (`TemplateSyntaxFactoryImpl.FixInterpolationSyntax`): Checks `TypeAnnotationMapper` annotations (set by `GetDynamicSyntax` for `IExpression` values like `parameter.Value`). **The expansion-time annotation takes priority** over the compile-time hint, since it reflects the actual expanded type. This prevents false positives where the template type is `dynamic` but the actual expanded value is e.g. `int`.

3. **Updated `ITemplateSyntaxFactory.FixInterpolationSyntax`** to accept an `isDynamic` parameter (with default `false` for backward compatibility).

4. **Kept the `InterpolatedStringUserExpression` fix** for the separate `InterpolatedStringBuilder` code path.

### Test results
- Both Bug580 regression tests pass (dynamic local variables AND dynamic parameters in interpolated strings)
- All 2019 aspect tests pass
- All 312 template tests pass  
- All 1525 unit tests pass

17 `.t.cs` expected output files were updated where the `(object)` cast is now correctly generated. These are cases where local variables assigned from `meta.Proceed()` or similar dynamic sources are used in interpolated strings — the cast is necessary to prevent CS9230.

Currently running `Build.ps1 build` for final verification.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateSyntaxFactoryImpl.cs
PostSharpAgent: Agreed, the HashSet approach was wrong. I've replaced it with a two-level approach:

1. **Compile-time detection** (in `TemplateCompilerRewriter.TransformInterpolation`): Uses the semantic model to detect when the interpolation expression is of type `dynamic`, and passes `isDynamic: true` to `FixInterpolationSyntax`. This handles local variables like `result` from `meta.Proceed()` that don't go through `GetDynamicSyntax`.

2. **Expansion-time annotations** (in `GetDynamicSyntax`): The `ExpressionSyntax` case now annotates with the actual type via `TypeAnnotationMapper.AddExpressionTypeAnnotation`. The `IExpression` case already annotated correctly.

3. **Annotation takes priority**: In `FixInterpolationSyntax`, if a type annotation exists on the expression (set at expansion time), it overrides the `isDynamic` flag from the compiler. This prevents false positives where the template type is `dynamic` (e.g., `InterpolatedStringBuilder.ToValue()`) but the actual expanded value is `string`.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Bugs/Bug580_DynamicInInterpolatedString.cs
PostSharpAgent: Good point — the HashSet approach would indeed fail with conflicting variable names across scopes. I've updated the test case to include nested scopes with different variables (`result` and `result2` in a nested `if` block) as you suggested. The new approach uses the semantic model at compile time rather than tracking names, so it correctly handles scoping.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Bugs/Bug580_DynamicInInterpolatedString.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        Console.WriteLine( $"Method returned: {result}" );
+
+        // Test with different blocks and a variable with the same name in a nested scope.
+        // This ensures the fix is not based on tracking variable names (which would break
+        // with conflicting names across scopes).
+        if ( meta.Target.Parameters.Count > 0 )
+        {
+            var result2 = meta.Target.Parameters[0].Value;
```
gfraiteur, line 28: This is still not good. There is no naming conflict. You can have first variable declaration in the `else` block to have two variables with the same name.

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Async/AsyncIterators/AsyncIteratorTemplate.t.cs
```diff
@@ -6,7 +6,7 @@ public async IAsyncEnumerable<int> Enumerable(int a)
     global::System.Console.WriteLine("Starting Enumerable");
     await foreach (var item in this.Enumerable_Source(a))
     {
-      global::System.Console.WriteLine($" Intercepting {item}");
+      global::System.Console.WriteLine($" Intercepting {(object)item}");
```
gfraiteur: There should no cast here because the run-time type of item is _not_ dynamic. Only the compile-time type is `dynamic`. So you should not get type information from the template semantic model, but from the compile-time object model.

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Async/AsyncIterators/AsyncIteratorTemplate.t.cs
```diff
@@ -28,7 +28,7 @@ public async IAsyncEnumerator<int> Enumerator(int a)
     var enumerator = this.Enumerator_Source(a);
     while (await enumerator.MoveNextAsync())
     {
-      global::System.Console.WriteLine($" Intercepting {enumerator.Current}");
+      global::System.Console.WriteLine($" Intercepting {(object)enumerator.Current}");
```
gfraiteur: Idem

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Async/AsyncIterators/ProceedConfigureAwait.t.cs
```diff
@@ -6,7 +6,7 @@ public async IAsyncEnumerable<int> Enumerable(int a)
     global::System.Console.WriteLine("Starting Enumerable");
     await foreach (var item in global::System.Threading.Tasks.TaskAsyncEnumerableExtensions.ConfigureAwait(this.Enumerable_Source(a), false))
     {
-      global::System.Console.WriteLine($" Intercepting {item}");
+      global::System.Console.WriteLine($" Intercepting {(object)item}");
```
gfraiteur: Idem

gfraiteur (changes requested):

gfraiteur:
@PostSharpAgent Address my feedback. Especially: rely on the type taken from the compile-time object mode, not from the semantic model of the template. Otherwise you get many irrelevant casts.

PostSharpAgent:
## Progress Update — Session 3

Starting a new session to address the latest review feedback from @gfraiteur.

**Key feedback to address:**
1. Use the **compile-time object model** types instead of the template semantic model for detecting dynamic types. The template semantic model reports many things as `dynamic` that aren't actually dynamic at runtime (e.g., `item` in async iterator templates).
2. Fix the test to include a proper naming conflict (variable declarations in `else` block).

Will investigate how to get type information from the compile-time object model and update the approach accordingly.

— Claude for @PostSharpAgent

PostSharpAgent:
## Progress Update — Session 3 (continued)

All tests pass with the new approach:
- 1989 aspect tests ✓
- 312 template tests ✓
- 1525 unit tests ✓

### Changes summary

**Core change:** Removed dependency on the template semantic model for detecting dynamic types in interpolated strings. Now relies solely on the **compile-time object model**:

1. **`TransformInterpolation`**: Removed the `isDynamic` detection that used `_syntaxTreeAnnotationMap.GetExpressionType()` (template semantic model). This was causing false positives for foreach variables in iterator templates, async method results, etc.

2. **`GetDynamicSyntax` `ExpressionSyntax` case**: No longer blindly annotates raw syntax as dynamic. For expressions that are just syntax (like local variable identifiers), the template-level `dynamic` type doesn't mean the runtime type is dynamic.

3. **`DynamicLocalDeclaration`**: Records the compile-time object model type (`value.Type`) for each local variable. This allows `FixInterpolationSyntax` to check whether a local is truly dynamic at runtime.

4. **`FixInterpolationSyntax`**: Now checks two sources:
   - Type annotations from `GetDynamicSyntax` (for `IExpression` values like `parameter.Value`)
   - Local variable type dictionary (for locals from `DynamicLocalDeclaration`)

5. **Test updated**: Bug580 test now includes conflicting variable names in if/else blocks as requested.

6. **17 .t.cs files reverted**: All spurious `(object)` casts removed.

Committing and pushing now. Will run `Build.ps1 build` and `Build.ps1 test` next.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Bugs/Bug580_DynamicInInterpolatedString.cs
PostSharpAgent: Fixed. The test now has an `else` block with a variable declaration using the same name `result2`, so both the `if` and `else` branches declare variables with the same name in different scopes.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Async/AsyncIterators/AsyncIteratorTemplate.t.cs
PostSharpAgent: Fixed. The approach now uses the compile-time object model instead of the template semantic model:

1. Removed `isDynamic` detection from `TransformInterpolation` that used `_syntaxTreeAnnotationMap.GetExpressionType()`.
2. Changed `GetDynamicSyntax` `ExpressionSyntax` case to pass through without annotating as dynamic.
3. Added local variable type tracking in `DynamicLocalDeclaration` using `value.Type` from the compile-time object model.
4. `FixInterpolationSyntax` now checks: (a) type annotations from `GetDynamicSyntax` for `IExpression` values, (b) local variable type dictionary for `DynamicLocalDeclaration` locals.

All 17 spurious `(object)` casts have been reverted.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Bugs/Bug580_DynamicInInterpolatedString.cs
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2020-2025 SharpCrafters s.r.o. and contributors.
+// SharpCrafters s.r.o. licenses this file to you under either the MIT license or a proprietary license, depending on the repository from which it was obtained.
+// Refer to LICENSE.md in the repository root for complete details.
+
```
gfraiteur, line 4: Use the @WriteCompiledTemplate test directive so I can verify the logic.

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Bugs/Bug580_DynamicParameterInInterpolatedString.cs
```diff
@@ -0,0 +1,50 @@
...
+// SharpCrafters s.r.o. licenses this file to you under either the MIT license or a proprietary license, depending on the repository from which it was obtained.
+// Refer to LICENSE.md in the repository root for complete details.
+
+using System;
+using Metalama.Framework.Aspects;
+using Metalama.Framework.Code;
+
+namespace Metalama.Framework.Tests.AspectTests.Tests.Aspects.Bugs.Bug580_DynamicParameterInInterpolatedString;
+
+internal class LogAttribute : OverrideMethodAspect
```
gfraiteur, line 15: Use the @WriteCompiledTemplate test directive so I can verify the logic.

gfraiteur (changes requested):
Address inline feedback. Also merge from develop/2026.1 and address conflicts.

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Bugs/Bug580_DynamicInInterpolatedString.cs
PostSharpAgent: Done. Added `@WriteCompiledTemplate` directive. The compiled template output is now in `Bug580_DynamicInInterpolatedString.ct.cs`. You can verify that `FixInterpolationSyntax` is called for each interpolation within the compiled template.

— Claude for @gfraiteur

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Bugs/Bug580_DynamicParameterInInterpolatedString.cs
PostSharpAgent: Done. Added `@WriteCompiledTemplate` directive. The compiled template output is now in `Bug580_DynamicParameterInInterpolatedString.ct.cs`.

— Claude for @gfraiteur

PostSharpAgent:
## Session 6: Addressed review feedback

### Changes made:
1. **Added `@WriteCompiledTemplate` test directive** to both `Bug580_DynamicInInterpolatedString.cs` and `Bug580_DynamicParameterInInterpolatedString.cs`. The generated `.ct.cs` files show the compiled template code so the dynamic detection logic can be verified.

2. **Merged from `develop/2026.1`** and resolved one conflict in `TemplateSyntaxFactoryImpl.cs` (new `RunTimeTypeParameterIdentifier` and `GetRunTimeTypeParameterSubstitutions` methods from #766 were added alongside our custom `FixInterpolationSyntax`).

### Verification:
- `Build.ps1 build`: success, zero warnings
- 2001 aspect tests pass (net8.0)
- 1530 unit tests pass (net8.0)
- 313 template tests pass (net8.0)

— Claude for @gfraiteur

gfraiteur (approved):

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateSyntaxFactoryImpl.cs
```diff
@@ -33,6 +34,12 @@ internal sealed partial class TemplateSyntaxFactoryImpl : ITemplateSyntaxFactory
         private readonly TemplateExpansionContext _templateExpansionContext;
         private readonly ObjectReaderFactory _objectReaderFactory;
 
+        // Maps expanded local variable names to their compile-time object model types.
+        // Expanded names are unique (produced by EscapeIdentifier/LexicalScope) so there are no scoping conflicts.
+        // This allows FixInterpolationSyntax to detect truly-dynamic locals vs template-level dynamic that resolves
+        // to a concrete type at expansion time.
+        private readonly Dictionary<string, IType> _localVariableTypes = new();
+
```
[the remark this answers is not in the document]
gfraiteur: @PostSharpAgent Create a test for this scenario. The dictionary should probably be duplicated in ForLocalFunction.

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateSyntaxFactoryImpl.cs
```diff
@@ -606,7 +641,43 @@ public IdentifierNameSyntax RunTimeTypeParameterIdentifier( string templateParam
...
+                isDynamic = annotatedType.TypeKind == TypeKind.Dynamic;
+            }
+            else if ( fixedInterpolation.Expression.IsKind( SyntaxKind.IdentifierName )
+                      && fixedInterpolation.Expression is IdentifierNameSyntax identifierName
+                      && this._localVariableTypes.TryGetValue( identifierName.Identifier.Text, out var localType ) )
+            {
+                // For local variables declared via DynamicLocalDeclaration, check their type from the
+                // compile-time object model. This handles cases like 'var result = meta.Proceed()' where
+                // the template-level type is always 'dynamic' but the actual runtime type depends on the target.
+                isDynamic = localType.TypeKind == TypeKind.Dynamic;
```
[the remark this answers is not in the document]
gfraiteur: I think we can't walk the expression on the right hand of a dynamic expression because there is no type analysis anyway. @PostSharpAgent What do you think?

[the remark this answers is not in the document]
PostSharpAgent: You're right. When a dynamic local is used in a larger expression (e.g., `result.ToString()`), the entire expression is dynamically bound, but at expansion time we have no type analysis on the expanded expression — we only have the syntax tree and the `_localVariableTypes` dictionary. While we could walk descendant nodes to find dynamic identifier roots, this would be fragile and may produce false positives for cases where the dynamic result is explicitly cast or wrapped before being used in the interpolation. In practice, using a dynamic local in a member access inside an interpolated string is likely rare, and the user would need to cast to `object` explicitly in such cases. I don't think walking the expression is the right approach here.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateSyntaxFactoryImpl.cs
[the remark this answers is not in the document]
PostSharpAgent: Done. Created `Bug580_DynamicInInterpolatedStringLocalFunction` test — an aspect where a dynamic local (`result = meta.Proceed()`) is declared in the outer scope and then captured by a local function that uses it in an interpolated string.

The `ForLocalFunction` method now copies the outer `_localVariableTypes` dictionary to the new factory instance via a private constructor, so the local function's `FixInterpolationSyntax` can correctly detect captured dynamic locals.

— Claude for @gfraiteur

gfraiteur (approved):

PR #1337, opened 2026-02-08
Fix compile-time elvis operator in templates (#585)

### LINKED ISSUE #585: Template with compile-time Elvis does not compile
Extracted from incorrectly-generated code:

```cs

            // var tokenParameters = meta.Target.Parameters.FirstOrDefault(p => p.Name == "Token")?.Value?.ToString();


            templateSyntaxFactory.AddStatement(                
                __s1,
                global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.LocalDeclarationStatement(                    
                    default(global::Microsoft.CodeAnalysis.SyntaxList<global::Microsoft.CodeAnalysis.CSharp.Syntax.AttributeListSyntax>),
                    default(global::Microsoft.CodeAnalysis.SyntaxToken),
                    default(global::Microsoft.CodeAnalysis.SyntaxToken),
                    default(global::Microsoft.CodeAnalysis.SyntaxTokenList),
                    global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.VariableDeclaration(                        
                        global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.IdentifierName(
                        global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.Identifier("var")),
                        global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.SingletonSeparatedList<global::Microsoft.CodeAnalysis.CSharp.Syntax.VariableDeclaratorSyntax>(                            
                            global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.VariableDeclarator(                                
                                tokenParametersName,
                                default(global::Microsoft.CodeAnalysis.CSharp.Syntax.BracketedArgumentListSyntax),
                                global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.EqualsValueClause(                                    
                                    global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.Token(global::Microsoft.CodeAnalysis.CSharp.SyntaxKind.EqualsToken),
                                    global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.ConditionalAccessExpression(                                        
                                        templateSyntaxFactory.GetTypedExpression(                                            
                                            global::Metalama.Framework.Aspects.meta.Target.Parameters.FirstOrDefault(p => p.Name == "Token")),
                                        global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.Token(global::Microsoft.CodeAnalysis.CSharp.SyntaxKind.QuestionToken),
                                        templateSyntaxFactory.ConditionalAccessExpression(                                            
                                            templateSyntaxFactory.GetDynamicSyntax(.Value),   // <--- Invalid syntax
                                            templateSyntaxFactory.AddSimplifierAnnotations(                                                
                                                global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.InvocationExpression(                                                    
                                                    global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.MemberBindingExpression(                                                        
                                                        global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.Token(global::Microsoft.CodeAnalysis.CSharp.SyntaxKind.DotToken),
                                                        global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.IdentifierName(
                                            global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.Identifier("ToString"))),
                                                    global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.ArgumentList(                                                        
                                                        default(global::Microsoft.CodeAnalysis.SeparatedSyntaxList<global::Microsoft.CodeAnalysis.CSharp.Syntax.ArgumentSyntax>)))))))))),
                    global::Microsoft.CodeAnalysis.CSharp.SyntaxFactory.Token(global::Microsoft.CodeAnalysis.CSharp.SyntaxKind.SemicolonToken)));


```

PostSharpAgent:
## Summary
- Fixes #585: Template with compile-time Elvis does not compile
- When using the null-conditional (`?.`) operator in a T# template where the left side is a compile-time expression and the right side accesses a run-time member (via a compile-time-returning-run-time property like `.Value`), the template compiler now reports a clear LAMA0289 error instead of generating invalid code.

## Rationale
When the compile-time expression (e.g. `meta.Target.Parameters.FirstOrDefault(...)`) is null, its type is unknown, so type-preserving run-time code cannot be generated. Rather than generating incorrect code (like `((dynamic)null)?.ToLower()` which changes the result type), we report an error and suggest using a compile-time null check (`if` statement) instead.

## Changes
- Added `LAMA0289` diagnostic (`CannotUseConditionalAccessWithCompileTimeToRunTimeMember`) in `TemplatingDiagnosticDescriptors.cs`
- Added detection of the unsupported pattern in `TemplateCompilerRewriter.TransformConditionalAccessExpression`
- Removed the `TryTransformMixedScopeConditionalAccess` method and `NullOfType` from `ITemplateSyntaxFactory` / `TemplateSyntaxFactoryImpl`
- Updated `CompileTimeElvis` test to expect LAMA0289 error

## Checklist
- [x] Reproduce bug with regression test
- [x] Implement fix (transformation approach)
- [x] Address review feedback (rounds 1-3)
- [x] Rollback transformation and report error instead (per review)
- [x] `Build.ps1 build` — zero warnings
- [x] All 2000 aspect tests pass (net8.0)
- [x] All 1530 unit tests pass (net8.0)

— Claude for @gfraiteur

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.CompileTimeContracts/ITemplateSyntaxFactory.cs
+++ b/Metalama.Framework/src/Metalama.Framework.CompileTimeContracts/ITemplateSyntaxFactory.cs
@@ -86,6 +86,8 @@ public interface ITemplateSyntaxFactory
 
     ExpressionSyntax ConditionalAccessExpression( ExpressionSyntax expression, ExpressionSyntax whenNotNullExpression );
 
+    ExpressionSyntax NullOfType( IType? type );
+
     ExpressionSyntax StringLiteralExpression( string? value );
 
     TypeOfExpressionSyntax TypeOf( string typeOfString, Dictionary<string, TypeSyntax> substitutions );
--- a/Metalama.Framework/src/Metalama.Framework.CompileTimeContracts/ITemplateSyntaxFactory.cs
+++ b/Metalama.Framework/src/Metalama.Framework.CompileTimeContracts/ITemplateSyntaxFactory.cs
@@ -86,8 +86,6 @@ public interface ITemplateSyntaxFactory
 
     ExpressionSyntax ConditionalAccessExpression( ExpressionSyntax expression, ExpressionSyntax whenNotNullExpression );
 
-    ExpressionSyntax NullOfType( IType? type );
-
     ExpressionSyntax StringLiteralExpression( string? value );
 
     TypeOfExpressionSyntax TypeOf( string typeOfString, Dictionary<string, TypeSyntax> substitutions );
--- a/Metalama.Framework/src/Metalama.Framework.CompileTimeContracts/ITemplateSyntaxFactory.cs
+++ b/Metalama.Framework/src/Metalama.Framework.CompileTimeContracts/ITemplateSyntaxFactory.cs
@@ -92,6 +92,13 @@ public interface ITemplateSyntaxFactory
 
     TypeOfExpressionSyntax TypeOf( string typeOfString, Dictionary<string, TypeSyntax> substitutions );
 
+    /// <summary>
+    /// Maps a template run-time type parameter name to the correct identifier in the current expansion context.
+    /// When the target method's type parameter has been renamed (e.g., by an introduction), this returns
+    /// the correct name based on ordinal position mapping.
+    /// </summary>
+    IdentifierNameSyntax RunTimeTypeParameterIdentifier( string templateParameterName );
+
     InterpolationSyntax FixInterpolationSyntax( InterpolationSyntax interpolation );
 
     ITemplateSyntaxFactory ForLocalFunction( string returnType, Dictionary<string, IType> genericArguments, bool isAsync = false );
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateCompilerRewriter.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateCompilerRewriter.cs
@@ -2940,7 +2940,17 @@ internal sealed partial class TemplateCompilerRewriter : MetaSyntaxRewriter, IDi
             {
                 // Fully qualifies simple identifiers.
 
-                if ( symbol?.Kind is SymbolKind.Namespace or SymbolKind.NamedType or SymbolKind.ArrayType or SymbolKind.DynamicType or SymbolKind.ErrorType
+                if ( symbol?.Kind == SymbolKind.TypeParameter
+                     && symbol is ITypeParameterSymbol typeParameterSymbol
+                     && this._templateMemberClassifier.IsRunTimeTemplateTypeParameter( typeParameterSymbol ) )
+                {
+                    // For run-time type parameters, use ITemplateSyntaxFactory.RunTimeTypeParameterIdentifier
+                    // so the correct name is resolved at expansion time via TemplateExpansionContext.
+                    return InvocationExpression(
+                            this._templateMetaSyntaxFactory.TemplateSyntaxFactoryMember( nameof(ITemplateSyntaxFactory.RunTimeTypeParameterIdentifier) ) )
+                        .AddArgumentListArguments( Argument( SyntaxFactoryEx.LiteralExpression( typeParameterSymbol.Name ) ) );
+                }
+                else if ( symbol?.Kind is SymbolKind.Namespace or SymbolKind.NamedType or SymbolKind.ArrayType or SymbolKind.DynamicType or SymbolKind.ErrorType
                          or SymbolKind.FunctionPointerType or SymbolKind.PointerType or SymbolKind.TypeParameter
                      && symbol is INamespaceOrTypeSymbol namespaceOrType )
                 {
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateCompilerRewriter.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateCompilerRewriter.cs
@@ -963,7 +963,7 @@ internal sealed partial class TemplateCompilerRewriter : MetaSyntaxRewriter, IDi
         // run-time member (e.g. .Value?.ToString()). We decompose the WhenNotNull to find the compile-time-to-runtime
         // bridge (e.g. .Value), evaluate it at compile time, and generate the remaining chain as run-time syntax.
         if ( transformationKind != TransformationKind.Transform
-             && this.TryTransformMixedScopeConditionalAccess( node ) is { } mixedScopeResult )
+             && this.TryTransformMixedScopeConditionalAccess( node, out var mixedScopeResult ) )
         {
             return mixedScopeResult;
         }
@@ -974,19 +974,19 @@ internal sealed partial class TemplateCompilerRewriter : MetaSyntaxRewriter, IDi
     /// <summary>
     /// Handles a conditional access where the expression (left side of ?.) is compile-time and the WhenNotNull
     /// (right side) starts with a compile-time-to-runtime bridge member (e.g. .Value on IParameter).
-    /// Returns null if this pattern is not matched.
     /// </summary>
     /// <remarks>
-    /// <para>For example, <c>meta.Target.Parameters.FirstOrDefault(...)?.Value?.ToString()</c> where:</para>
+    /// <para>For example, <c>meta.Target.Parameters.FirstOrDefault(...)?.Value?.ToLower()</c> where:</para>
     /// <list type="bullet">
     /// <item><c>meta.Target.Parameters.FirstOrDefault(...)</c> is compile-time (<c>IParameter?</c>)</item>
     /// <item><c>.Value</c> is compile-time-returning-runtime (the parameter's run-time value)</item>
-    /// <item><c>?.ToString()</c> is run-time</item>
+    /// <item><c>?.ToLower()</c> is run-time</item>
     /// </list>
     /// <para>The generated compile-time code evaluates the expression at compile time, and if non-null, accesses the
-    /// bridge member to get the run-time syntax, wrapping any remaining chain in a run-time conditional access.</para>
+    /// bridge member to get the run-time syntax, wrapping any remaining chain in a run-time conditional access.
+    /// When null, it generates a typed null expression using the compile-time expression's type information.</para>
     /// </remarks>
-    private ExpressionSyntax? TryTransformMixedScopeConditionalAccess( ConditionalAccessExpressionSyntax node )
+    private bool TryTransformMixedScopeConditionalAccess( ConditionalAccessExpressionSyntax node, out ExpressionSyntax result )
     {
         // Decompose the WhenNotNull to find the first MemberBindingExpression (the compile-time-to-runtime bridge).
         // The WhenNotNull could be:
@@ -1008,19 +1008,23 @@ internal sealed partial class TemplateCompilerRewriter : MetaSyntaxRewriter, IDi
                 break;
 
             case SyntaxKind.ConditionalAccessExpression
-                when ((ConditionalAccessExpressionSyntax) node.WhenNotNull).Expression.Kind() == SyntaxKind.MemberBindingExpression:
+                when node.WhenNotNull is ConditionalAccessExpressionSyntax
+                     {
+                         Expression: MemberBindingExpressionSyntax innerMemberBinding
+                     } conditionalAccess:
             {
                 // Chain case: expr?.Value?.rest
-                var innerConditionalAccess = (ConditionalAccessExpressionSyntax) node.WhenNotNull;
-                firstMemberBinding = (MemberBindingExpressionSyntax) innerConditionalAccess.Expression;
-                remainingWhenNotNull = innerConditionalAccess.WhenNotNull;
+                firstMemberBinding = innerMemberBinding;
+                remainingWhenNotNull = conditionalAccess.WhenNotNull;
 
                 break;
             }
 
             default:
                 // Not a recognized pattern for mixed-scope conditional access.
-                return null;
+                result = null!;
+
+                return false;
         }
 
         // Check that the bridge member is compile-time-returning-runtime.
@@ -1028,7 +1032,9 @@ internal sealed partial class TemplateCompilerRewriter : MetaSyntaxRewriter, IDi
 
         if ( !bridgeScope.IsCompileTimeMemberReturningRunTimeValue() )
         {
-            return null;
+            result = null!;
+
+            return false;
         }
 
         // Compile-time expression: node.Expression.MemberName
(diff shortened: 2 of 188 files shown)


Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateCompilerRewriter.cs
```diff
@@ -952,15 +952,159 @@ protected override ExpressionSyntax TransformConditionalAccessExpression( Condit
...
+            case SyntaxKind.MemberBindingExpression:
+                // Simple case: expr?.Value — the whole WhenNotNull is the bridge member.
+                firstMemberBinding = (MemberBindingExpressionSyntax) node.WhenNotNull;
+                remainingWhenNotNull = null;
+
+                break;
+
+            case SyntaxKind.ConditionalAccessExpression
+                when ((ConditionalAccessExpressionSyntax) node.WhenNotNull).Expression.Kind() == SyntaxKind.MemberBindingExpression:
+            {
```
gfraiteur: Rewrite as:
case SyntaxKind.ConditionalAccessExpression

when  node.WhenNotNull is ConditionalAccessExpressionSyntax { Expression.SyntaxKind:  == SyntaxKind.MemberBindingExpression } conditionalAccessExpression:

to avoid the second cast

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateCompilerRewriter.cs
```diff
@@ -952,15 +952,159 @@ protected override ExpressionSyntax TransformConditionalAccessExpression( Condit
...
+    /// <para>For example, <c>meta.Target.Parameters.FirstOrDefault(...)?.Value?.ToString()</c> where:</para>
+    /// <list type="bullet">
+    /// <item><c>meta.Target.Parameters.FirstOrDefault(...)</c> is compile-time (<c>IParameter?</c>)</item>
+    /// <item><c>.Value</c> is compile-time-returning-runtime (the parameter's run-time value)</item>
+    /// <item><c>?.ToString()</c> is run-time</item>
+    /// </list>
+    /// <para>The generated compile-time code evaluates the expression at compile time, and if non-null, accesses the
+    /// bridge member to get the run-time syntax, wrapping any remaining chain in a run-time conditional access.</para>
+    /// </remarks>
+    private ExpressionSyntax? TryTransformMixedScopeConditionalAccess( ConditionalAccessExpressionSyntax node )
```
gfraiteur: The "Try" pattern should return bool and use `out` to return the payload

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateCompilerRewriter.cs
```diff
@@ -952,15 +952,159 @@ protected override ExpressionSyntax TransformConditionalAccessExpression( Condit
...
+        }
+
+        // Build the "when null" branch. We generate the same run-time structure as the "when not null" branch
+        // but with ((object?)null) as the receiver. For example, if the not-null branch produces `token?.ToString()`,
+        // the null branch produces `((object?)null)?.ToString()`, which is valid C# evaluating to `(string?)null`.
+        // This preserves the expression type and avoids "Cannot assign null to implicitly-typed variable" for var declarations.
+        //
+        // For the simple case (expr?.Value), there's no remaining chain and we can't use this approach. We fall back
+        // to generating a cast expression like `(object?)null` directly.
+        var castNullToObjectSyntax = this.MetaSyntaxFactory.CastExpression(
```
gfraiteur: Casting `null` to `object` is not useful. You should cast to the type of node.Expression. The problem is that `this._syntaxTreeAnnotationMap.GetExpressionType` will return `object/dynamic` in many cases, so we need to take the type from the compile-time value of the expression.

Example: what if we don't call `ToString` which always exists but say `ToLower` which might only be available to strings?

So, that's a difficult problem.

Members annotated with [CompileTimeReturningRunTime] can return a FiedInfo.

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Misc/CompileTimeElvis.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Metalama.Framework.Tests.AspectTests.Tests.Aspects.Misc.CompileTimeElvis;
+
+internal class Aspect : OverrideMethodAspect
+{
+    public override dynamic? OverrideMethod()
+    {
+        // Compile-time expression (IParameter?) with null-conditional access to .Value (dynamic).
+        var tokenParameters = meta.Target.Parameters.FirstOrDefault( p => p.Name == "token" )?.Value?.ToString();
+
```
gfraiteur, line 20: You should test another method than ToString for instance ToLower(). To make sure the cast works.

gfraiteur (dismissed):

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateCompilerRewriter.cs
PostSharpAgent: Done. Rewrote to use pattern matching in the `when` clause: `when node.WhenNotNull is ConditionalAccessExpressionSyntax { Expression: MemberBindingExpressionSyntax innerMemberBinding } conditionalAccess`. This avoids the second cast while still being safe from LAMA0860 since the switch governing expression is a `Kind()` access.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateCompilerRewriter.cs
PostSharpAgent: Done. Changed to `private bool TryTransformMixedScopeConditionalAccess(ConditionalAccessExpressionSyntax node, out ExpressionSyntax result)`.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateCompilerRewriter.cs
PostSharpAgent: Done. Added `NullOfType(IType? type)` to `ITemplateSyntaxFactory` and implemented it in `TemplateSyntaxFactoryImpl`. It generates `((T?)null)` using the type from the compile-time expression via `((IHasType?)compileTimeExpr)?.Type`. When the compile-time expression is null (parameter not found), `Type` is also null and it falls back to `((object?)null)`. The null branch no longer chains methods (e.g. `?.ToLower()`) since the result is always null regardless, and the chain method might not exist on the fallback type.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Misc/CompileTimeElvis.cs
PostSharpAgent: Done. Changed from `ToString()` to `ToLower()`. This verifies the null branch uses the correct type cast — `ToLower()` is only available on `string`, not on `object`, so if the null branch tried to chain it on `((object?)null)`, the generated code would fail to compile.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Misc/CompileTimeElvis.t.cs
```diff
@@ -0,0 +1,18 @@
...
+    var tokenParameters = token?.ToLower();
+    return (global::System.String? )tokenParameters;
+  }
+  // Method WITHOUT a matching parameter - should resolve to (object?)null since
+  // the compile-time expression is null and the type cannot be determined.
+  [Aspect]
+  private string? MethodWithoutToken(int x)
+  {
+    var tokenParameters = ((global::System.Object? )null);
+    return (global::System.String? )tokenParameters;
```
gfraiteur: Here is proof that it does not work. `tokenParameters` should be of type `string`, but you made it of type `object`. So your transformation changed the type. This is why you must keep the `ToLower`. Because you don't know the type of the expression of the right hand of `?.` - you genuinely don't know because the right hand is `dynamic` so you don't know what the the generated expression will be typed.

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateCompilerRewriter.cs
```diff
@@ -952,15 +952,153 @@ protected override ExpressionSyntax TransformConditionalAccessExpression( Condit
...
+                    ArgumentList( SeparatedList( [Argument( runtimeExpressionSyntax ), Argument( transformedRemainingWhenNotNull )] ) ) );
+        }
+
+        // Build the "when null" branch using NullOfType at template expansion time.
+        // We generate: templateSyntaxFactory.NullOfType(((IHasType?)compileTimeExpr)?.Type)
+        // This gets the run-time type from the compile-time expression (e.g. IParameter.Type gives the parameter's
+        // actual type like string), and generates ((T?)null) in the output. If the compile-time expression is null,
+        // Type will be null too and NullOfType falls back to ((object?)null).
+        //
+        // We don't chain ?.ToLower() etc. on the null expression because the result is always null regardless,
```
gfraiteur: As noted below, you just chain the left hand to preserve the type.

gfraiteur (changes requested):

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Misc/CompileTimeElvis.t.cs
PostSharpAgent: Fixed. The null branch now preserves the chain (e.g. `?.ToLower()`) by wrapping `NullOfType` in a `ConditionalAccessExpression` with the same transformed `remainingWhenNotNull`. Additionally, `NullOfType` now falls back to `((dynamic)null)` instead of `((object?)null)` when the type is unknown, since `dynamic` supports any member access.

For `MethodWithoutToken`, the output is now `((dynamic)null)?.ToLower()` instead of `((object?)null)`. The chain ensures the expression has the correct result type, and `dynamic` ensures the chain compiles even when the underlying type is unknown.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateCompilerRewriter.cs
PostSharpAgent: Done. The null branch now chains the `WhenNotNull` by wrapping `NullOfType` in `ConditionalAccessExpression(NullOfType(type), transformedRest)`. This produces `((string?)null)?.ToLower()` when the type is known, or `((dynamic)null)?.ToLower()` when unknown. The chain preserves the correct result type.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.CompileTimeContracts/ITemplateSyntaxFactory.cs
```diff
@@ -86,6 +86,8 @@ AnonymousFunctionExpressionSyntax SimplifyAnonymousFunction<T>( T node )
 
     ExpressionSyntax ConditionalAccessExpression( ExpressionSyntax expression, ExpressionSyntax whenNotNullExpression );
 
+    ExpressionSyntax NullOfType( IType? type );
+
```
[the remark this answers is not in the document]
gfraiteur: Irrelevant. Adding methods is fine, removing is not.

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateCompilerRewriter.cs
```diff
@@ -952,15 +952,153 @@ protected override ExpressionSyntax TransformConditionalAccessExpression( Condit
...
+
+        var transformedWhenNull = InvocationExpression(
+                this._templateMetaSyntaxFactory.TemplateSyntaxFactoryMember( nameof(ITemplateSyntaxFactory.NullOfType) ) )
+            .AddArgumentListArguments( Argument( compileTimeType ) );
+
+        // Build the compile-time ternary: compileTimeExpr != null ? whenNotNull : whenNull
+        var compileTimeCondition = BinaryExpression(
+            SyntaxKind.NotEqualsExpression,
+            node.Expression,
+            SyntaxFactoryEx.Null );
```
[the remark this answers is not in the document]
PostSharpAgent: Fixed. Replaced triple evaluation of `node.Expression` with an `is {} __ct0` pattern match that evaluates the compile-time expression exactly once and binds it to a temp variable for the when-true branch. In the when-false branch, the expression is guaranteed null, so we pass `null` directly to `NullOfType` instead of re-evaluating. Commit: 7288047a06.

— Claude for @gfraiteur

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Misc/CompileTimeElvis.t.cs
```diff
@@ -0,0 +1,18 @@
...
+  {
+    var tokenParameters = token?.ToLower();
+    return (global::System.String? )tokenParameters;
+  }
+  // Method WITHOUT a matching parameter - should resolve to ((dynamic)null)?.ToLower() since
+  // the compile-time expression is null. The chain is preserved to maintain the correct result type.
+  [Aspect]
+  private string? MethodWithoutToken(int x)
+  {
+    var tokenParameters = ((dynamic)null)?.ToLower();
```
gfraiteur: This code is not acceptable.

gfraiteur (changes requested):
We failed to find an acceptable solution and need to rollback and instead report an error when a bridge compile-time-to-runtime (find a better terminology in line with what we already use) elvis ?. is used.

Document the bottom-line reason: if we don't have the compile-time expression (because it's null), we don't know its type, and we cannot generate any type-preserving code. So it's better to report an error.

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Misc/CompileTimeElvis.t.cs
PostSharpAgent: Agreed. Rolled back the transformation approach entirely. The template compiler now reports **LAMA0289** error when a \ operator crosses from compile-time to run-time via a compile-time-returning-run-time member. The error message explains that when the compile-time expression is null, its type is unknown, so type-preserving code cannot be generated, and suggests using a compile-time \ statement instead.

PostSharpAgent: Agreed. Rolled back the transformation approach entirely. The template compiler now reports **LAMA0289** error when a `?.` operator crosses from compile-time to run-time via a compile-time-returning-run-time member. The error message explains that when the compile-time expression is null, its type is unknown, so type-preserving code cannot be generated, and suggests using a compile-time `if` statement instead.

gfraiteur (approved):

PR #1353, opened 2026-02-12
Fix [InterfaceMember] and [Template] with same name causing exception (#718)

### LINKED ISSUE #718: [InterfaceMember] and [Template] with the same name cause exception when using the template
When searching for a (non-interface) template, interface members should be ignored.

https://try.metalama.net/#N4IgZiBcDaoHZRABQE4HsDmKCGBbAdAMYDOIANCIYgK7ECWcGABALICmALtgDZ7b4AxHLjYB3NCgDW+AILEADm0IdiAbgA6cWg2bsuvXPyF4xE6QGE0AEzYa4mhhzYo4PJoV7FiTVJmGbgTSZgpnkUOgA3bCcmYi4OOkImCLQ6K1ZsBgAKAEogkMC4EKYAX00y+zhoABU2OIBdTXlqACNuRPdPb1q48y6A4vLNJtb2pMdnMGxCNiYASXYROA40AZD84IW2JZWmcxQ2aLZFtmW0Odx5bly7Yq2dtD2Do5Ozi6usxyYADzImL4Anjk7BURm0Oh5sF4mD0ODIOBxwi1qDFIDCAYo5IplGtghtQqMOmgIs5wjZkql0gAhah0bhWLFKDhZOaM5Q0uk2FAAHjmADkTFZqhi2AA+JjIznOPJFAr44qS+nOWRWCKJNj4ObLdBWagzPQAC2sWUVXPw1WwKAwnD+rhEaDAWX2hycrxW724OT+lowxEgcDETGA3yYAF4mABGErA+UhAD0cdNypkqvVmsu3G2pw4WqcKCmMxNtKVKHNPptTA4IodLLdaByMdlwVBTaY0Fzk2mx22LWcjVbzXBSXu2cezpeWbeGdy+MKxWKBw41BcTC03G4txCLeK7eWnf1Pb7+MHY3mdaeLu7Dw9WX+yx+f0BMuKc-nwUXy6Ka43+O3IRq2xXEc-bFCeHQjmcF4Tte07QJYlx0mw1R0CI9R3hwPzPnKrYLpwn6rtQ66bs25TkCAkiIBwIAlPUJRAA

PostSharpAgent:
## Summary

Fixes #718 - When an aspect has both an `[InterfaceMember]` method and a `[Template]` method with the same name (differentiated by generic type parameters), calling `IntroduceMethod(nameof(Foo))` throws `InvalidOperationException: Sequence contains more than one element`.

Also fixes Bug34183 - `ImplementInterface` failing to introduce generic overloads like `Foo<T>()` alongside non-generic `Foo()`.

## Root Causes

Four issues in the template resolution and member comparison pipeline:

1. **`TemplateMemberRef.GetTemplateMember`**: When resolving a template symbol by name, the predicate only matched value parameters but not type parameters. Both `[InterfaceMember] Foo()` and `[Template] Foo<T>()` have zero value parameters, so `SingleOrDefault` found two matches and threw.

2. **`TemplateNameHelper.GetCompiledTemplateName`**: The compiled template name was `__Foo` for both methods since the hash suffix was only added for methods with value parameters. This caused `AmbiguousMatchException` during Reflection lookup.

3. **`MemberComparer<T>`**: `Equals` and `GetHashCode` did not consider type parameter count when comparing methods. This caused `AllMethods` (which uses a `HashSet<IMethod>` with this comparer) to silently deduplicate `Foo()` and `Foo<T>()`, preventing `ImplementInterface` from finding templates for generic overloads.

4. **`SignatureMatcher.OfExactSignature`/`OfSignature`**: Did not filter by type parameter count, so `IntroduceMethod` incorrectly matched `Foo()` when looking for `Foo<T,U>()` (both have 0 value parameters).

## Fix

1. Added type parameter count matching to the symbol resolution predicate in `TemplateMemberRef`.
2. Extended hash suffix logic in `TemplateNameHelper` to also apply when a method has type parameters.
3. Added type parameter count to `MemberComparer.Equals` and `GetHashCode` for `IMethod` instances.
4. Added `typeParameterCount` parameter to `SignatureMatcher.OfExactSignature` and `OfSignature`, passed through from `MethodCollectionExtensions`.

## Test Plan

- [x] Bug718.cs regression test (new)
- [x] Bug34183 test un-skipped and passing
- [x] All 2027 aspect tests pass (0 failures)
- [x] All 1537 unit tests pass (0 failures)
- [x] Build.ps1 build succeeds with zero warnings
- [x] Build.ps1 test succeeds with zero warnings

— Claude for @gfraiteur

### DIFF
--- a/Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Bugs/Bug34183.cs
+++ b/Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Bugs/Bug34183.cs
@@ -3,7 +3,7 @@
 // Refer to LICENSE.md in the repository root for complete details.
 
 #if TEST_OPTIONS
-// @Skipped (#34183 - [Template] [InterfaceMember] conflict)
+// @Skipped (#34183 - IntroduceMethod with same name as [InterfaceMember] needs prefix naming)
 #endif
 
 using Metalama.Framework.Advising;
--- a/Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Bugs/Bug34183.cs
+++ b/Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Bugs/Bug34183.cs
@@ -2,10 +2,6 @@
 // SharpCrafters s.r.o. licenses this file to you under either the MIT license or a proprietary license, depending on the repository from which it was obtained.
 // Refer to LICENSE.md in the repository root for complete details.
 
-#if TEST_OPTIONS
-// @Skipped (#34183 - IntroduceMethod with same name as [InterfaceMember] needs prefix naming)
-#endif
-
 using Metalama.Framework.Advising;
 using Metalama.Framework.Aspects;
 using Metalama.Framework.Code;
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Advising/TemplateMemberRef.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Advising/TemplateMemberRef.cs
@@ -57,11 +57,13 @@ internal readonly struct TemplateMemberRef
         var type = templateReflectionContext.Compilation.GetTypeByMetadataNameSafe( this._templateMember.TemplateClass.FullName );
 
         var parameters = this._templateMember.Parameters;
+        var typeParameterCount = this._templateMember.TypeParameters.Length;
 
         var symbol = type.GetSingleMemberIncludingBase(
             this._templateMember.Name,
             symbol => classifier.IsTemplate( symbol )
-                      && symbol.GetParameters().Select( p => p.Type ).SequenceEqual( parameters.Select( p => p.Type ), StructuralSymbolComparer.Default ) );
+                      && symbol.GetParameters().Select( p => p.Type ).SequenceEqual( parameters.Select( p => p.Type ), StructuralSymbolComparer.Default )
+                      && (symbol is not Microsoft.CodeAnalysis.IMethodSymbol methodSymbol || methodSymbol.TypeParameters.Length == typeParameterCount) );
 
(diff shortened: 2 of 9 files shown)


Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Bugs/Bug34183.cs
```diff
@@ -3,7 +3,7 @@
 // Refer to LICENSE.md in the repository root for complete details.
 
 #if TEST_OPTIONS
-// @Skipped (#34183 - [Template] [InterfaceMember] conflict)
+// @Skipped (#34183 - IntroduceMethod with same name as [InterfaceMember] needs prefix naming)
```
gfraiteur: The point is that this test should no longer be skipped.

gfraiteur (changes requested):

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Bugs/Bug34183.cs
PostSharpAgent: Done. Bug34183 is now un-skipped and passes.

The test required two additional fixes beyond the previous session's changes:

1. **`MemberComparer`**: `Equals` and `GetHashCode` did not consider type parameter count, so `AllMethods` (which uses a `HashSet<IMethod>` with this comparer) was silently deduplicating `Foo()` and `Foo<T>()`. This prevented `ImplementInterface` from finding the template for `Foo<T>()`.

2. **`SignatureMatcher.OfExactSignature`/`OfSignature`**: Did not filter by type parameter count, so `IntroduceMethod` incorrectly matched `Foo()` when looking for `Foo<T,U>()`.

All 2027 aspect tests and 1537 unit tests pass.

— Claude for @gfraiteur

gfraiteur (approved):

PR #1354, opened 2026-02-12
Fix NRE in transitive aspect processing for cross-version assemblies (#728)

### LINKED ISSUE #728: TransitiveAspectSource throws NullReferenceException when aspects are coming from assembly referencing older Metalama
REDACTED(Internal)

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Metalama.Framework.Engine.Aspects.TransitiveAspectSource..ctor(Compilation compilation, ImmutableArray`1 aspectClasses, ProjectServiceProvider serviceProvider)
   at Metalama.Framework.Engine.Pipeline.AspectPipeline.CreateAspectSources(AspectPipelineConfiguration configuration, Compilation compilation, CancellationToken cancellationToken)
   at Metalama.Framework.Engine.Pipeline.AspectPipeline.<ExecuteAsync>d__28.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Metalama.Framework.Engine.Pipeline.CompileTime.CompileTimeAspectPipeline.<ExecuteCoreAsync>d__3.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Metalama.Framework.Engine.Pipeline.CompileTime.CompileTimeAspectPipeline.<ExecuteAsync>d__2.MoveNext()
```

REDACTED(Internal) happens after upgrading reference to 2023.2.4 to 2023.3.8.

PostSharpAgent:
## Summary

Fixes #728: `TransitiveAspectSource` (now `TransitivePipelineContributorSource`) throws `NullReferenceException` when processing transitive aspects from assemblies compiled with a different Metalama version.

### Root Cause

When a manifest serialized by an older Metalama version is deserialized by a newer version, fields that didn't exist in the older format (e.g., `Annotations`, `Extensions`) return `null`/`default` from `IArgumentsReader.GetValue`. This causes NRE when the properties are subsequently accessed.

### Changes

1. **`TransitiveAspectsManifest.DeserializeFields`**: Use null-coalescing and `TryGetValue` with empty defaults for backward compatibility.
2. **`SerializableTransitiveAspectInstance.ToAspectInstance`**: Changed to use `TryGetAspectClass`, returning null when aspect class not found.
3. **`TransitiveAspectPipelineExtension`**: Now stateless. Reports `LAMA0072` (`UnknownTransitiveAspectClass`) diagnostic directly to a `UserDiagnosticSink` passed into `GetPipelineContributorsFromTransitiveManifest`.
4. **`PipelineExtension.GetPipelineContributorsFromTransitiveManifest`**: Added `UserDiagnosticSink diagnosticSink` parameter (breaking change for overrides — companion PR: metalama/Metalama.Premium#47).
5. **`TransitivePipelineContributorSource`**: Creates `UserDiagnosticSink`, passes it to extensions, exposes `Diagnostics` property. Fixed log message printing dictionary instead of string.
6. **`PipelineContributorSources`**: Added `Diagnostics` property to propagate extension diagnostics.
7. **`AspectPipeline.ExecuteAsync`**: Merges contributor source diagnostics into initialization diagnostics.
8. **`GeneralDiagnosticDescriptors`**: Added `LAMA0072` (`UnknownTransitiveAspectClass`) diagnostic definition (**Error** severity).
9. **Standalone regression test (`Issue728/`)**: Two-project test with OldVersion (Metalama 2025.0.12) and CurrentVersion (current Metalama), exercises PE assembly cross-version deserialization. `test.json` with `BuildOnly: true`.

### Companion PR

- metalama/Metalama.Premium#47: Updates `ValidationPipelineExtension.GetPipelineContributorsFromTransitiveManifest` signature.

## Checklist

- [x] Identify root cause
- [x] Implement fix
- [x] Report unknown aspect classes as build error (LAMA0072) via diagnostic sink
- [x] Make `TransitiveAspectPipelineExtension` stateless (pass `UserDiagnosticSink` as parameter)
- [x] Change LAMA0072 from Warning to Error severity
- [x] Add `test.json` to standalone test
- [x] Add standalone cross-version regression test
- [x] Build.ps1 build (0 warnings)
- [x] Validation tests pass (Metalama.Premium)
- [x] Issue728 standalone test passes
- [x] Address all review feedback
- [ ] Full CI build

— Claude for @gfraiteur

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/SerializableTransitiveAspectInstance.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/SerializableTransitiveAspectInstance.cs
@@ -44,23 +44,20 @@ internal class SerializableTransitiveAspectInstance : ICompileTimeSerializable, 
         this.TargetDeclarationDepth = targetDeclarationDepth;
     }
 
-    // public TransitiveAspectInstance ToAspectInstance( IAspectClassResolver aspectClassResolver )
-    // {
-    //     return new TransitiveAspectInstance(
-    //         this.Aspect,
-    //         this.TargetDeclaration,
-    //         this.TargetDeclarationDepth,
-    //         (IAspectClassImpl)aspectClassResolver.GetAspectClass( this.AspectClassName ),
-    //         this.AspectState,
-    //         0 );
-    // }
-    public AspectInstance ToAspectInstance( IAspectClassResolver aspectClassResolver )
+    public AspectInstance? ToAspectInstance( IAspectClassResolver aspectClassResolver )
     {
+        if ( !aspectClassResolver.TryGetAspectClass( this.AspectClassName, out var aspectClass ) )
+        {
+            // The aspect class may not be found when the referenced assembly was compiled with a different
+            // version of Metalama that had a different set of aspect classes.
+            return null;
+        }
+
         return new AspectInstance(
             this.Aspect,
             this.TargetDeclaration,
             this.TargetDeclarationDepth,
-            (IAspectClassImpl) aspectClassResolver.GetAspectClass( this.AspectClassName ),
+            (IAspectClassImpl) aspectClass,
             [],
             ImmutableArray<AspectPredecessor>.Empty,
             false );
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/SerializableTransitiveAspectInstance.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/SerializableTransitiveAspectInstance.cs
@@ -30,7 +30,7 @@ internal class SerializableTransitiveAspectInstance : ICompileTimeSerializable, 
         transitiveAspectInstance.TargetDeclaration,
         transitiveAspectInstance.TargetDeclarationDepth ) { }
 
-    private SerializableTransitiveAspectInstance(
+    internal SerializableTransitiveAspectInstance(
         IAspect aspect,
         string aspectClassName,
         IAspectState? aspectState,
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectPipelineExtension.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectPipelineExtension.cs
@@ -4,7 +4,6 @@
 
 using Metalama.Backstage.Diagnostics;
 using Metalama.Framework.Aspects;
-using Metalama.Framework.Code.Collections;
 using Metalama.Framework.Engine.CodeModel;
 using Metalama.Framework.Engine.Collections;
 using Metalama.Framework.Engine.Diagnostics;
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectPipelineExtension.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectPipelineExtension.cs
@@ -2,6 +2,7 @@
 // SharpCrafters s.r.o. licenses this file to you under either the MIT license or a proprietary license, depending on the repository from which it was obtained.
 // Refer to LICENSE.md in the repository root for complete details.
 
+using Metalama.Backstage.Diagnostics;
 using Metalama.Framework.Aspects;
 using Metalama.Framework.Code.Collections;
 using Metalama.Framework.Engine.CodeModel;
@@ -9,6 +10,7 @@ using Metalama.Framework.Engine.Collections;
 using Metalama.Framework.Engine.Diagnostics;
 using Metalama.Framework.Engine.Extensibility;
 using Metalama.Framework.Engine.Pipeline;
+using Metalama.Framework.Engine.Services;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
@@ -19,6 +21,15 @@ namespace Metalama.Framework.Engine.Aspects;
 
 internal partial class TransitiveAspectPipelineExtension : PipelineExtension
 {
+    private ILogger? _logger;
+
+    public override bool Initialize( PipelineExtensionInitializationContext context )
+    {
+        this._logger = context.ServiceProvider.GetLoggerFactory().GetLogger( nameof(TransitiveAspectPipelineExtension) );
+
+        return true;
+    }
+
     public override IEnumerable<ITransitiveAspectsManifestExtension> GetTransitiveManifestExtensions( IEnumerable<ITransitivePipelineContributor> contributors )
         => contributors.OfKind( ContributorKind.TransitiveAspectInstance ).Select( x => new SerializableTransitiveAspectInstance( x ) );
 
@@ -26,12 +37,27 @@ internal partial class TransitiveAspectPipelineExtension : PipelineExtension
         ImmutableArray<ITransitiveAspectsManifestExtension> extensions,
         IAspectClassResolver aspectClassResolver )
     {
-        var transitiveAspectInstances = extensions.OfKind( ContributorKind.SerializableTransitiveAspectInstance )
-            .Select( i => i.ToAspectInstance( aspectClassResolver ) )
-            .WhereNotNull()
-            .ToMultiValueDictionary( x => (IAspectClass) x.AspectClass, x => x );
+        var serializableInstances = extensions.OfKind( ContributorKind.SerializableTransitiveAspectInstance );
 
-        yield return new AspectSource( transitiveAspectInstances );
+        var transitiveAspectInstancesBuilder = ImmutableDictionaryOfArray<IAspectClass, AspectInstance>.CreateBuilder();
+
+        foreach ( var instance in serializableInstances )
+        {
+            var aspectInstance = instance.ToAspectInstance( aspectClassResolver );
+
+            if ( aspectInstance != null )
+            {
+                transitiveAspectInstancesBuilder.Add( aspectInstance.AspectClass, aspectInstance );
+            }
+            else
+            {
+                this._logger?.Warning?.Log(
+                    $"Cannot create a transitive aspect instance for aspect class '{instance.AspectClassName}'. " +
+                    $"The aspect class was not found. This can happen when the referenced assembly was compiled with a different version of Metalama." );
+            }
+        }
+
+        yield return new AspectSource( transitiveAspectInstancesBuilder.ToImmutable() );
     }
 
     public override Task<ExtensionPipelineContributorsResult> ExecutePipelineContributorsAsync(
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectPipelineExtension.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectPipelineExtension.cs
@@ -3,6 +3,7 @@
 // Refer to LICENSE.md in the repository root for complete details.
 
 using Metalama.Framework.Aspects;
+using Metalama.Framework.Code.Collections;
 using Metalama.Framework.Engine.CodeModel;
 using Metalama.Framework.Engine.Collections;
 using Metalama.Framework.Engine.Diagnostics;
@@ -27,6 +28,7 @@ internal partial class TransitiveAspectPipelineExtension : PipelineExtension
     {
         var transitiveAspectInstances = extensions.OfKind( ContributorKind.SerializableTransitiveAspectInstance )
             .Select( i => i.ToAspectInstance( aspectClassResolver ) )
+            .WhereNotNull()
             .ToMultiValueDictionary( x => (IAspectClass) x.AspectClass, x => x );
 
         yield return new AspectSource( transitiveAspectInstances );
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectPipelineExtension.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectPipelineExtension.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2020-2025 SharpCrafters s.r.o. and contributors.
+// Copyright (c) 2020-2025 SharpCrafters s.r.o. and contributors.
 // SharpCrafters s.r.o. licenses this file to you under either the MIT license or a proprietary license, depending on the repository from which it was obtained.
 // Refer to LICENSE.md in the repository root for complete details.
 
@@ -8,7 +8,6 @@ using Metalama.Framework.Engine.Collections;
 using Metalama.Framework.Engine.Diagnostics;
 using Metalama.Framework.Engine.Extensibility;
 using Metalama.Framework.Engine.Pipeline;
-using Microsoft.CodeAnalysis;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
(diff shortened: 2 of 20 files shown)


Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitivePipelineContributorSource.cs
```diff
@@ -60,11 +64,23 @@ public TransitivePipelineContributorSource(
...
+                            {
+                                manifest = TransitiveAspectsManifest.Deserialize(
+                                    new MemoryStream( bytes ),
+                                    serviceProvider,
+                                    compilationContext,
+                                    filePath );
+                            }
+                            catch ( Exception ex )
+                            {
+                                // This can happen when the referenced assembly was compiled with an older version of Metalama
```
gfraiteur: This is not acceptable. We should let the method crash. If it's crashes we must crash the build.

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/Pipeline/InheritableAspectTests.cs
```diff
@@ -147,5 +147,59 @@ public interface I {}
...
+    public void IntroducedMethod() {}
+}
+
+[OldVersionAspect]
+public interface I {}
+";
+
+            // Project 2: references project 1 but does NOT define OldVersionAspect.
+            // This simulates the scenario where a dependency was compiled with an older
+            // Metalama version that had a different aspect class.
```
gfraiteur: I don't understand how this can possibly reproduce a multi-version bug. See C:\src\Metalama-2026.1\Metalama\Metalama.Framework\src\tests\RunManually\MultiVersion about how to reproduce. First create a situation that reproduces the bug. Then we will decide if it's solvable. But in no case just swallog exceptions.

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.Engine/Aspects/SerializableTransitiveAspectInstance.cs
```diff
@@ -44,23 +44,20 @@ private SerializableTransitiveAspectInstance(
...
+    public AspectInstance? ToAspectInstance( IAspectClassResolver aspectClassResolver )
     {
+        if ( !aspectClassResolver.TryGetAspectClass( this.AspectClassName, out var aspectClass ) )
+        {
+            // The aspect class may not be found when the referenced assembly was compiled with a different
+            // version of Metalama that had a different set of aspect classes.
+            return null;
+        }
+
         return new AspectInstance(
```
gfraiteur, line 56: Report an error if the aspect instance cannot be created

Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectsManifest.cs
```diff
@@ -134,19 +134,32 @@ public override void DeserializeFields( object obj, IArgumentsReader initializat
         {
             var instance = (TransitiveAspectsManifest) obj;
 
+            // Fields use null-coalescing to provide defaults for backward compatibility with manifests
+            // serialized by older Metalama versions that may not have all fields. (#728)
+
             instance.InheritableAspects =
-                initializationArguments.GetValue<ImmutableDictionary<string, IReadOnlyList<InheritableAspectInstance>>>( nameof(instance.InheritableAspects) )!;
+                initializationArguments.GetValue<ImmutableDictionary<string, IReadOnlyList<InheritableAspectInstance>>>( nameof(instance.InheritableAspects) )
```
gfraiteur, line 141: Seems good, but I insist on creating a regression test as a standalone solution with two projects - one targeting an old Metalama, the other the current one.

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitivePipelineContributorSource.cs
PostSharpAgent: Agreed. I've removed the try-catch around `Deserialize()` and instead fixed the root cause: `TransitiveAspectsManifest.Serializer.DeserializeFields` now handles missing fields from older serialization formats by using default values (empty collections) instead of leaving them as `null`. This way, if `Deserialize()` fails for other reasons, the build will still crash as expected.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/Pipeline/InheritableAspectTests.cs
PostSharpAgent: You're right — the previous test didn't reproduce a genuine multi-version scenario. I've removed it and instead identified the actual root cause: `TransitiveAspectsManifest.Serializer.DeserializeFields` doesn't handle missing fields from older serialization formats. When `IArgumentsReader.GetValue` returns null/default for fields that didn't exist in the older version (like \`Annotations\`, added later), the NRE occurs.

The fix makes \`DeserializeFields\` use null-coalescing with empty defaults for all fields. This is not exception swallowing — it's making the deserialization format backward-compatible.

The existing \`SideBySideVersionTests\` and \`InheritableAspectTests\` cover the \`CompilationReference\` path. For the PE reference deserialization path (the actual crash path), manual verification with two Metalama builds per \`RunManually/MultiVersion\` is needed.

— Claude for @PostSharpAgent

gfraiteur (changes requested):
@PostSharpAgent  I insist on creating a regression test as a standalone solution with two projects - one targeting an old Metalama, the other the current one.

Metalama.Framework/src/Metalama.Framework.Engine/Aspects/SerializableTransitiveAspectInstance.cs
PostSharpAgent: Done. I've added a logger to `TransitiveAspectPipelineExtension` (captured via `Initialize`) that logs a warning when `ToAspectInstance` returns null due to an unknown aspect class. This follows the same pattern used in `TransitivePipelineContributorSource` (line 97-99) for inherited aspects.

The `ToAspectInstance` method still returns null to allow graceful continuation, but the caller now explicitly reports the error instead of silently filtering it.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectsManifest.cs
PostSharpAgent: Done. I've created a standalone regression test at `Standalone/Issue728` with two projects:

- **OldVersion**: References `Metalama.Framework 2025.0.12` from NuGet.org. Defines an `[Inheritable]` aspect on an interface. Has its own `Directory.Build.props`, `Directory.Packages.props`, and `nuget.config` to isolate from the parent build chain.

- **CurrentVersion**: Uses the current Metalama version. References OldVersion's compiled DLL directly (not as ProjectReference) so it goes through the `PortableExecutableReference` code path — the actual deserialization path where the NRE occurred.

The `ProjectDependencies` section in the `.sln` ensures OldVersion builds first, then CurrentVersion references the compiled DLL.

— Claude for @PostSharpAgent

PostSharpAgent:
Addressed all review feedback:

1. **Error reporting**: `TransitiveAspectPipelineExtension` now captures an `ILogger` via `Initialize()` and logs a warning when a transitive aspect instance cannot be created due to an unknown aspect class.

2. **Standalone regression test**: Created `Standalone/Issue728` with:
   - `OldVersion` project using `Metalama.Framework 2025.0.12` from NuGet.org (isolated build props + nuget.config)
   - `CurrentVersion` project using current Metalama, referencing OldVersion as a PE assembly reference
   - This exercises the exact deserialization path (`TransitiveAspectsManifest.Deserialize` via `PortableExecutableReference`) where the NRE occurred

`Build.ps1 build` is running now.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Aspects/SerializableTransitiveAspectInstance.cs
gfraiteur: Bad. I wanted to report it through a diagnostic sink. (report as a build error, not a logger error).

Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectPipelineExtension.cs
```diff
@@ -18,18 +20,43 @@ namespace Metalama.Framework.Engine.Aspects;
...
+            if ( aspectInstance != null )
+            {
+                transitiveAspectInstancesBuilder.Add( aspectInstance.AspectClass, aspectInstance );
+            }
+            else
+            {
+                this._logger?.Warning?.Log(
+                    $"Cannot create a transitive aspect instance for aspect class '{instance.AspectClassName}'. " +
+                    $"The aspect class was not found. This can happen when the referenced assembly was compiled with a different version of Metalama." );
+            }
```
gfraiteur, line 47: Additionally, report a diagnostic (build error).

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.Engine/Aspects/SerializableTransitiveAspectInstance.cs
PostSharpAgent: Done. Replaced logger-based warning with proper build diagnostic `LAMA0072` (`UnknownTransitiveAspectClass`). The diagnostic is collected during `GetPipelineContributorsFromTransitiveManifest` and reported through `ExtensionPipelineContributorsResult` in both `ExecutePipelineContributorsAsync` and `ExecuteDesignTimePipelineContributorsAsync`.

The `ILogger` field and `Initialize` override have been removed — no longer needed.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectPipelineExtension.cs
PostSharpAgent: Done. The diagnostic `LAMA0072` is now reported through the `ExtensionPipelineContributorsResult` diagnostics mechanism. When `ToAspectInstance` returns null, the diagnostic is collected and then emitted as part of the pipeline execution result — replacing the logger-based warning entirely.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectPipelineExtension.cs
```diff
@@ -18,18 +19,43 @@ namespace Metalama.Framework.Engine.Aspects;
 
 internal partial class TransitiveAspectPipelineExtension : PipelineExtension
 {
+    private ImmutableUserDiagnosticList _pendingDiagnostics = ImmutableUserDiagnosticList.Empty;
+
```
gfraiteur: Bad design. Don't make TransitiveAspectPipelineExtension  stateful. Instead, pull a UserDiagnosticSink into GetPipelineContributorsFromTransitiveManifest.

Metalama.Framework/src/Metalama.Framework.Engine/Diagnostics/GeneralDiagnosticDescriptors.cs
```diff
@@ -357,6 +357,15 @@ internal static readonly DiagnosticDefinition<IMethod>
...
                 "Cannot infer the type arguments for a generic method." );
 
+        internal static readonly DiagnosticDefinition<string>
+            UnknownTransitiveAspectClass = new(
+                "LAMA0072",
+                _category,
+                "Cannot find the transitive aspect class '{0}'. " +
+                "This can happen when the referenced assembly was compiled with a different version of Metalama.",
+                Warning,
+                "Unknown transitive aspect class." );
```
gfraiteur, line 367: This should be an error.

Metalama.Framework/src/tests/Standalone/Issue728/Issue728.sln
```diff
@@ -0,0 +1,34 @@
+
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.11.35208.52
```
gfraiteur, line 4: there should be a `test.json` file in the directory to mention that an error is expected. See other standalone tests for information about how this works.

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectPipelineExtension.cs
PostSharpAgent: Done. `TransitiveAspectPipelineExtension` is now stateless — removed the `_pendingDiagnostics` field and the `Initialize` override. Added a `UserDiagnosticSink diagnosticSink` parameter to `PipelineExtension.GetPipelineContributorsFromTransitiveManifest`, which `TransitivePipelineContributorSource` creates and passes in. Diagnostics flow through `PipelineContributorSources.Diagnostics` → merged into `initializationDiagnostics` in `AspectPipeline.ExecuteAsync`.

The companion PR for `ValidationPipelineExtension` in Metalama.Premium is metalama/Metalama.Premium#47.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/Diagnostics/GeneralDiagnosticDescriptors.cs
PostSharpAgent: Done. Changed LAMA0072 (`UnknownTransitiveAspectClass`) from `Warning` to `Error` severity.

— Claude for @gfraiteur

Metalama.Framework/src/tests/Standalone/Issue728/Issue728.sln
PostSharpAgent: Done. Added `test.json` with `{ "BuildOnly": true }`.

Note: the Issue728 standalone test does **not** produce LAMA0072 because it tests the **inherited aspects** path (not the transitive aspect instances path). In this test, `OldVersion` defines an `[Inheritable]` aspect applied to an interface, and `CurrentVersion` has a class implementing that interface. The aspect class (`InheritedAspect`) is resolvable from the referenced assembly, so no "unknown aspect class" error occurs. The LAMA0072 diagnostic would only trigger for `SerializableTransitiveAspectInstance` entries in the manifest where the aspect class name cannot be resolved.

The test validates that cross-version `TransitiveAspectsManifest` deserialization works correctly without NRE — which is the core bug from issue #728. If you'd like a separate test that exercises the LAMA0072 error path (where a transitive aspect class is genuinely unknown), I can add one.

— Claude for @gfraiteur

gfraiteur: Yes I want this test showing LAMA0072 .

gfraiteur (changes requested):

Metalama.Framework/src/tests/Standalone/Issue728/Issue728.sln
PostSharpAgent: After thorough investigation, LAMA0072 cannot be triggered in a standalone cross-version test using published Metalama NuGet packages:

1. **Pre-2026.0.3 versions**: Don't produce `SerializableTransitiveAspectInstance` entries (the `Extensions` field in the manifest didn't exist)
2. **Post-2026.0.3 versions**: All use the same `PullConstructorParameterTransitiveAspect` class name (the only source of `SerializableTransitiveAspectInstance`), which resolves successfully

LAMA0072 is defensive code that fires when the manifest contains a `SerializableTransitiveAspectInstance` whose `AspectClassName` doesn't match any registered aspect class — a scenario that would occur if a future version renames or removes the internal system aspect.

**What I did instead:**
- Added unit tests (`TransitiveAspectDiagnosticTests`) that directly verify the LAMA0072 code path: `ToAspectInstance` returns null for unknown aspect class names
- Made `SerializableTransitiveAspectInstance`'s constructor `internal` (instead of `private`) to enable test construction
- Updated the standalone test's `test.json` with `FailOnUnexpectedDiagnostics: true` to verify the cross-version build produces no unexpected diagnostics

If you'd like me to take a different approach (e.g., crafting a fake assembly with a manipulated manifest resource), please let me know.

— Claude for @gfraiteur

gfraiteur: Ok

gfraiteur (approved):

PR #1356, opened 2026-02-12
Fix LAMA0306 for suppressions defined in compile-time project (#726)

### LINKED ISSUE #726: XAML: Command aspect fails to suppress warnings
See the attached project.

We get the warning that the suppressed diagnostic is not set in the user profile, but this warning doesn't disappear after VS restart.

PostSharpAgent:
Fixes #726

## Summary

The `[Command]` aspect from `Metalama.Patterns.Wpf` suppresses `IDE0051` (unused member) via a `SuppressionDefinition`, but the suppression was not working reliably in VS, and LAMA0306 was being reported.

### Root cause

When the pipeline initializes, `RegisterDescriptors` writes newly discovered suppression IDs to `userDiagnostics.json`. However, the `[Memo]`-cached `DiagnosticDefinitions` property in `UserDiagnosticRegistrationService` does not refresh within the same VS session. This means both `TheDiagnosticAnalyzer` and `TheDiagnosticSuppressor` use stale `SupportedSuppressionDescriptors` that do not include the newly registered suppression IDs.

### Fix

1. **`TheDiagnosticSuppressor`**: After the pipeline executes, merge suppression descriptors from the `DiagnosticManifest` into `supportedSuppressionDescriptors`. This allows the suppressor to handle manifest-defined suppressions within the same VS session, even when the user profile has not been updated yet.

2. **`TheDiagnosticAnalyzer`**: LAMA0306 is correctly reported when the suppression is not in `SupportedSuppressionDescriptors` — this tells the user to restart their IDE.

3. **`TestUserDiagnosticRegistrationService`**: Refactored for clarity — renamed `RegisteredManifests` to `ManifestsRegisteredByPipeline`, added `initialSuppressionDescriptors` parameter for configuring the simulated user profile, added XML documentation.

### Tests

- `SuppressionFromDependencyReportsLAMA0306WhenNotInUserProfile` (DiagnosticAnalyzerTests): Verifies LAMA0306 IS reported when the suppression is in the manifest but not in the user profile.
- `StaleUserProfileSuppressesDiagnosticAndReportsLAMA0306` (DiagnosticSuppressorTests): Verifies both (a) the suppressor suppresses CS0169 via manifest merging and (b) the analyzer reports LAMA0306.
- `TestDiagnosticSuppressionFromDependency` (UserDiagnosticRegistrationServiceTests): Verifies the pipeline correctly discovers and registers suppressions from dependencies.
- `TestDiagnosticSuppressionFromCurrentProject` (UserDiagnosticRegistrationServiceTests): Verifies suppressions from the current project are registered.

### Limitation

Roslyn `DiagnosticSuppressor.SupportedSuppressions` property is called before the pipeline runs and is cached by Roslyn. If a suppression ID is not in this cached list, Roslyn will not pass the corresponding diagnostics to `ReportSuppressions` at all. The `TheDiagnosticSuppressor` fix addresses the internal filtering but cannot override Roslyn external filtering. After VS restart, `SupportedSuppressions` is freshly computed and includes the registered IDs.

## Checklist

- [x] Understand the issue
- [x] Reproduce with regression tests
- [x] Implement fix
- [x] Address review feedback from @gfraiteur (LAMA0306 behavior, test refactoring)
- [x] Build - 0 warnings
- [x] Build.ps1 test - all pass
- [x] Finalize

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticAnalysis/TheDiagnosticAnalyzer.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticAnalysis/TheDiagnosticAnalyzer.cs
@@ -212,9 +212,14 @@ namespace Metalama.Framework.DesignTime.DiagnosticAnalysis
                     ReportDiagnostic,
                     cancellationToken );
 
-                // If we have unsupported suppressions, a diagnostic here because a Suppressor cannot report.
+                // If we have unsupported suppressions, report a diagnostic here because a Suppressor cannot report.
+                // We check both the user profile (SupportedSuppressionDescriptors) and the compile-time project's manifest
+                // because the user profile may not yet contain suppressions that were just registered during pipeline initialization.
+                var configurationManifest = pipelineResult.IsSuccessful ? pipelineResult.Value.Configuration?.DiagnosticManifest : null;
+
                 foreach ( var suppression in suppressions.Where(
-                             s => !this.DiagnosticDefinitions.SupportedSuppressionDescriptors.ContainsKey( s.Suppression.Definition.SuppressedDiagnosticId ) ) )
+                             s => !this.DiagnosticDefinitions.SupportedSuppressionDescriptors.ContainsKey( s.Suppression.Definition.SuppressedDiagnosticId )
+                                  && configurationManifest?.DefinesSuppression( s.Suppression.Definition.SuppressedDiagnosticId ) != true ) )
                 {
                     cancellationToken.ThrowIfCancellationRequested();
 
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticAnalysis/TheDiagnosticAnalyzer.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticAnalysis/TheDiagnosticAnalyzer.cs
@@ -212,7 +212,12 @@ namespace Metalama.Framework.DesignTime.DiagnosticAnalysis
                     ReportDiagnostic,
                     cancellationToken );
 
-                // If we have unsupported suppressions, report a diagnostic here because a Suppressor cannot report.
+                // Report LAMA0306 for suppressions that are in the pipeline's DiagnosticManifest but NOT
+                // in the user profile (SupportedSuppressionDescriptors). This happens when the IDE was started
+                // before the suppression was registered: the [Memo]-cached SupportedSuppressions property of
+                // TheDiagnosticSuppressor does not include the new suppression, so Roslyn won't pass the
+                // corresponding diagnostics to TheDiagnosticSuppressor.ReportSuppressions. The original
+                // warning therefore persists until the IDE is restarted. LAMA0306 informs the user of this.
                 foreach ( var suppression in suppressions.Where(
                              s => !this.DiagnosticDefinitions.SupportedSuppressionDescriptors.ContainsKey( s.Suppression.Definition.SuppressedDiagnosticId ) ) )
                 {
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticAnalysis/TheDiagnosticAnalyzer.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticAnalysis/TheDiagnosticAnalyzer.cs
@@ -213,13 +213,8 @@ namespace Metalama.Framework.DesignTime.DiagnosticAnalysis
                     cancellationToken );
 
                 // If we have unsupported suppressions, report a diagnostic here because a Suppressor cannot report.
-                // We check both the user profile (SupportedSuppressionDescriptors) and the compile-time project's manifest
-                // because the user profile may not yet contain suppressions that were just registered during pipeline initialization.
-                var configurationManifest = pipelineResult.IsSuccessful ? pipelineResult.Value.Configuration?.DiagnosticManifest : null;
-
                 foreach ( var suppression in suppressions.Where(
-                             s => !this.DiagnosticDefinitions.SupportedSuppressionDescriptors.ContainsKey( s.Suppression.Definition.SuppressedDiagnosticId )
-                                  && configurationManifest?.DefinesSuppression( s.Suppression.Definition.SuppressedDiagnosticId ) != true ) )
+                             s => !this.DiagnosticDefinitions.SupportedSuppressionDescriptors.ContainsKey( s.Suppression.Definition.SuppressedDiagnosticId ) ) )
                 {
                     cancellationToken.ThrowIfCancellationRequested();
 
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticSuppressing/TheDiagnosticSuppressor.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticSuppressing/TheDiagnosticSuppressor.cs
@@ -140,6 +140,19 @@ namespace Metalama.Framework.DesignTime.DiagnosticSuppressing
                     return;
                 }
 
+                // Merge suppression descriptors from the compile-time project's DiagnosticManifest.
+                // The user profile (SupportedSuppressionDescriptors) may not yet contain suppressions
+                // that were just registered during pipeline initialization in this VS session.
+                foreach ( var suppressionId in pipelineResult.Value.Configuration.DiagnosticManifest.SuppressionDefinitions.Keys )
+                {
+                    if ( !supportedSuppressionDescriptors.ContainsKey( suppressionId ) )
+                    {
+                        supportedSuppressionDescriptors = supportedSuppressionDescriptors.Add(
+                            suppressionId,
+                            SuppressionFactories.CreateDescriptor( suppressionId ) );
+                    }
+                }
+
                 var suppressionsCount = 0;
 
                 cancellationToken.ThrowIfCancellationRequested();
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticSuppressing/TheDiagnosticSuppressor.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticSuppressing/TheDiagnosticSuppressor.cs
@@ -143,14 +143,20 @@ namespace Metalama.Framework.DesignTime.DiagnosticSuppressing
                 // Merge suppression descriptors from the compile-time project's DiagnosticManifest.
                 // The user profile (SupportedSuppressionDescriptors) may not yet contain suppressions
                 // that were just registered during pipeline initialization in this VS session.
+                ImmutableDictionary<string, SuppressionDescriptor>.Builder? suppressionDescriptorsBuilder = null;
+
                 foreach ( var suppressionId in pipelineResult.Value.Configuration.DiagnosticManifest.SuppressionDefinitions.Keys )
                 {
                     if ( !supportedSuppressionDescriptors.ContainsKey( suppressionId ) )
                     {
-                        supportedSuppressionDescriptors = supportedSuppressionDescriptors.Add(
-                            suppressionId,
-                            SuppressionFactories.CreateDescriptor( suppressionId ) );
+                        suppressionDescriptorsBuilder ??= supportedSuppressionDescriptors.ToBuilder();
+                        suppressionDescriptorsBuilder.Add( suppressionId, SuppressionFactories.CreateDescriptor( suppressionId ) );
                     }
+                }
+
+                if ( suppressionDescriptorsBuilder != null )
+                {
+                    supportedSuppressionDescriptors = suppressionDescriptorsBuilder.ToImmutable();
                 }
 
                 var suppressionsCount = 0;
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticSuppressing/TheDiagnosticSuppressor.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticSuppressing/TheDiagnosticSuppressor.cs
@@ -140,24 +140,13 @@ namespace Metalama.Framework.DesignTime.DiagnosticSuppressing
                     return;
                 }
 
-                // Merge suppression descriptors from the compile-time project's DiagnosticManifest.
-                // The user profile (SupportedSuppressionDescriptors) may not yet contain suppressions
-                // that were just registered during pipeline initialization in this VS session.
-                ImmutableDictionary<string, SuppressionDescriptor>.Builder? suppressionDescriptorsBuilder = null;
-
-                foreach ( var suppressionId in pipelineResult.Value.Configuration.DiagnosticManifest.SuppressionDefinitions.Keys )
-                {
-                    if ( !supportedSuppressionDescriptors.ContainsKey( suppressionId ) )
-                    {
-                        suppressionDescriptorsBuilder ??= supportedSuppressionDescriptors.ToBuilder();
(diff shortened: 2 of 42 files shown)


gfraiteur (dismissed):

Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticAnalysis/TheDiagnosticAnalyzer.cs
```diff
@@ -212,9 +212,14 @@ void ReportDiagnostic( Diagnostic diagnostic )
...
-                // If we have unsupported suppressions, a diagnostic here because a Suppressor cannot report.
+                // If we have unsupported suppressions, report a diagnostic here because a Suppressor cannot report.
+                // We check both the user profile (SupportedSuppressionDescriptors) and the compile-time project's manifest
+                // because the user profile may not yet contain suppressions that were just registered during pipeline initialization.
+                var configurationManifest = pipelineResult.IsSuccessful ? pipelineResult.Value.Configuration?.DiagnosticManifest : null;
+
                 foreach ( var suppression in suppressions.Where(
-                             s => !this.DiagnosticDefinitions.SupportedSuppressionDescriptors.ContainsKey( s.Suppression.Definition.SuppressedDiagnosticId ) ) )
+                             s => !this.DiagnosticDefinitions.SupportedSuppressionDescriptors.ContainsKey( s.Suppression.Definition.SuppressedDiagnosticId )
+                                  && configurationManifest?.DefinesSuppression( s.Suppression.Definition.SuppressedDiagnosticId ) != true ) )
```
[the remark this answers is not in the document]
gfraiteur: @PostSharpAgent Do you understand what this says? Clarify, I don't understand.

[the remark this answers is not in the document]
PostSharpAgent: @gfraiteur Copilot raises a valid concern. Here's the breakdown:

**The problem Copilot identified:** My fix suppresses LAMA0306 when the manifest defines the suppression, but `TheDiagnosticSuppressor` still only uses `SupportedSuppressionDescriptors` (from the memoized user profile) to both:
1. Declare to Roslyn which diagnostic IDs it can suppress (`SupportedSuppressions` property, line 266)
2. Filter which pipeline suppressions to actually apply (line 160-161)

Since `SupportedSuppressions` is `[Memo]`-cached and was initialized before the suppression was registered, Roslyn won't pass `IDE0051` diagnostics to the suppressor at all. The result: within the same VS session where the suppression was first registered, `IDE0051` still persists **and** LAMA0306 is now silenced — so the user has no indication something is off.

**After VS restart**, a fresh `UserDiagnosticRegistrationService` instance reads the updated `userDiagnostics.json`, so `SupportedSuppressionDescriptors` includes `IDE0051` and everything works. The issue is **only within the first VS session**.

**My proposed fix:** In addition to the existing `TheDiagnosticAnalyzer` change, I will also update `TheDiagnosticSuppressor.ReportSuppressions` to use the `DiagnosticManifest` as a fallback source of suppression descriptors when the user profile doesn't have the entry yet. This way, the suppressor can actually suppress `IDE0051` in the same session.

Note: I cannot fix the `SupportedSuppressions` property (Roslyn calls it before the pipeline runs), but in `ReportSuppressions`, the pipeline result is available and I can create a temporary `SuppressionDescriptor` from the manifest for IDs not in the user profile.

I'll push this fix shortly.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/UserProfile/UserDiagnosticRegistrationServiceTests.cs
```diff
@@ -185,6 +185,93 @@ public void Foo() { }
...
+                                       }
+                                   }
+                                   """,
+                    targetCode: """
+                                namespace TestDependency;
+
+                                class TargetCode
+                                {
+                                    [SuppressWarning]
+                                    private void Foo() { }
```
gfraiteur, line 226: Add a comment about what IDE0051 is (the message),

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/UserProfile/UserDiagnosticRegistrationServiceTests.cs
```diff
@@ -185,6 +185,93 @@ public void Foo() { }
...
+                                    }
+                                }
+                                """,
+                    targetCode: """
+                                namespace TestProject;
+
+                                class TargetCode
+                                {
+                                    [SuppressWarning]
+                                    public void Foo() { }
```
gfraiteur, line 270: Also document CS0169

Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticAnalysis/TheDiagnosticAnalyzer.cs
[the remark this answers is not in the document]
gfraiteur: @PostSharpAgent Address the issue. Add a regression test for the scenario.

Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticSuppressing/TheDiagnosticSuppressor.cs
```diff
@@ -140,6 +140,19 @@ internal void ReportSuppressions(
...
+                // that were just registered during pipeline initialization in this VS session.
+                foreach ( var suppressionId in pipelineResult.Value.Configuration.DiagnosticManifest.SuppressionDefinitions.Keys )
+                {
+                    if ( !supportedSuppressionDescriptors.ContainsKey( suppressionId ) )
+                    {
+                        supportedSuppressionDescriptors = supportedSuppressionDescriptors.Add(
+                            suppressionId,
+                            SuppressionFactories.CreateDescriptor( suppressionId ) );
+                    }
+                }
```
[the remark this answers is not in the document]
gfraiteur: @PostSharpAgent address the issue.

gfraiteur (changes requested):

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/UserProfile/UserDiagnosticRegistrationServiceTests.cs
PostSharpAgent: Done. Added `// IDE0051: "Private member is unused"` comment above the `SuppressionDefinition`.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/UserProfile/UserDiagnosticRegistrationServiceTests.cs
PostSharpAgent: Done. Added `// CS0169: "The field is never used"` comment above the `SuppressionDefinition`.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticAnalysis/TheDiagnosticAnalyzer.cs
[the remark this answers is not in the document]
PostSharpAgent: Done. Added two regression tests:

1. **`DiagnosticAnalyzerTests.SuppressionFromDependencyDoesNotReportLAMA0306`** — verifies the analyzer does NOT report LAMA0306 when a dependency defines a `SuppressionDefinition` (for IDE0051), even though `SupportedSuppressionDescriptors` is empty (stale user profile). This validates the manifest fallback path at line 222.

2. **`DiagnosticSuppressorTests.SuppressFieldWarningWithEmptyUserProfile`** — verifies the suppressor still correctly suppresses CS0169 via manifest merging when `supportedSuppressionDescriptors` is empty. This confirms the end-to-end scenario works: neither LAMA0306 nor the original warning persists.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.DesignTime/DiagnosticSuppressing/TheDiagnosticSuppressor.cs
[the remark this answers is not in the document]
PostSharpAgent: Done. Replaced the one-by-one `ImmutableDictionary.Add()` loop with a lazy `ImmutableDictionary.Builder` pattern:
- The builder is only allocated when there are actually missing entries (`??=` pattern)
- All entries are added to the builder in a single pass
- `ToImmutable()` is called once at the end, only if any entries were added

This avoids creating intermediate `ImmutableDictionary` instances for each missing suppression ID.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/EndToEnd/DiagnosticAnalyzerTests.cs
```diff
@@ -182,4 +182,50 @@ class C {}
         Assert.Equal( diagnostic1.WarningLevel, diagnostic2.WarningLevel );
         Assert.Equal( diagnostic1.Properties, diagnostic2.Properties );
     }
+
+    [Fact]
+    public async Task SuppressionFromDependencyDoesNotReportLAMA0306()
+    {
+        // Regression test for #726: When a dependency defines a SuppressionDefinition (e.g., for IDE0051 "Private member is unused"),
+        // the analyzer should NOT report LAMA0306 ("unregistered suppression") even when SupportedSuppressionDescriptors
+        // is stale (empty), because the suppression is defined in the compile-time project's DiagnosticManifest.
```
gfraiteur: This comment highlights a misunderstanding.

  LAMA0306 should be reported when a diagnostic _should_ be reported because (a) there is a warning with this id and (b) there is a SuppressionDefinition (therefore it's in the DiagnosticManifest) and (c) `builser.Diagnostics.Suppress` has been invoked AND, IMPORTANT (c) the suppression is NOT supported because it was NOT in the user profile at the moment the IDE starts.

Make sure there is a unit test to cover this test in DiagnosticSuppressorTests.  This might require the use of TestUserDiagnosticRegistrationService, and it might be necessary to improve/refactor it. Rename its members so they make sense -- currently it's difficult to understand what they are doing.

gfraiteur (changes requested):

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/EndToEnd/DiagnosticAnalyzerTests.cs
PostSharpAgent: @gfraiteur You are right, my previous understanding was wrong. I have corrected this:

1. **Reverted the manifest fallback in `TheDiagnosticAnalyzer`**: LAMA0306 is now always reported when the suppression is not in `SupportedSuppressionDescriptors`, regardless of whether the manifest defines the suppression. This correctly tells the user to restart their IDE.

2. **Renamed and fixed the analyzer test** to `SuppressionFromDependencyReportsLAMA0306WhenNotInUserProfile` — it now asserts that LAMA0306 IS reported when the user profile is stale.

3. **Renamed and extended the suppressor test** to `StaleUserProfileSuppressesDiagnosticAndReportsLAMA0306` — it now validates both behaviors:
   - (a) The suppressor still suppresses CS0169 via manifest merging (even with empty user profile)
   - (b) The analyzer reports LAMA0306 because the suppression is not in the user profile

4. **Refactored `TestUserDiagnosticRegistrationService`**: Renamed `RegisteredManifests` → `ManifestsRegisteredByPipeline`, added `initialSuppressionDescriptors` parameter for configuring the simulated user profile, and added XML documentation.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/EndToEnd/DiagnosticSuppressorTests.cs
```diff
@@ -220,6 +221,73 @@ internal class TargetClass
...
+        var compilation = await workspaceProvider.GetProject( "project" ).GetCompilationAsync();
+        var diagnostics = compilation!.GetDiagnostics();
+
+        // (a) Verify the suppressor still suppresses CS0169 via manifest merging even with empty user profile.
+        var suppressor = new TheDiagnosticSuppressor( pipelineFactory.ServiceProvider );
+        var suppressionContext = new TestSuppressionAnalysisContext( compilation, diagnostics, testContext.ProjectOptions );
+
+        suppressor.ReportSuppressions( suppressionContext, ImmutableDictionary<string, SuppressionDescriptor>.Empty );
+
+        var suppression = Assert.Single( suppressionContext.ReportedSuppressions );
```
gfraiteur: This is incorrect. The diagnostic should NOT be suppressed because it's illegal for a DiagnosticSuppressor to suppress a diagnostic that is not declared in the list of suppressed diagnostics, which contains ONLY the diagnostics from the user profile (and a few well-known ones). The test works because it hosts DiagnosticSuppressor -- however it would fail if DiagnosticSuppressor is hosted by Roslyn.

gfraiteur (changes requested):

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/EndToEnd/DiagnosticSuppressorTests.cs
gfraiteur: Which means that the manifest merging might have been useless, and all that's needed is a better code comment.

PostSharpAgent: @gfraiteur You are correct. I've removed the manifest merging code from `TheDiagnosticSuppressor` entirely. As you pointed out, Roslyn's `DiagnosticSuppressor` contract requires all suppressible diagnostics to be declared in `SupportedSuppressions`, and Roslyn filters diagnostics before calling `ReportSuppressions` — so merging from the manifest was ineffective when hosted by Roslyn.

The changes in commit `eb00740`:

1. **Removed manifest merging** from `TheDiagnosticSuppressor.ReportSuppressions` and replaced it with a comment explaining why it's intentionally not done.
2. **Fixed the test** `StaleUserProfileDoesNotSuppressDiagnosticAndReportsLAMA0306` (renamed from `StaleUserProfile*Suppresses*...`) — it now correctly asserts that the suppressor does NOT suppress the diagnostic when the user profile is empty.
3. **Improved the LAMA0306 comment** in `TheDiagnosticAnalyzer` to explain the Roslyn limitation and why LAMA0306 is necessary.

— Claude for @PostSharpAgent

PostSharpAgent: Agreed. The manifest merging has been removed in commit `eb00740`. The suppressor now only contains a comment explaining the Roslyn limitation, and the test has been corrected to assert the diagnostic is NOT suppressed with a stale profile.

— Claude for @PostSharpAgent

gfraiteur (approved):

PR #1362, opened 2026-03-05
Add Changed event test for LicenseConsumptionService (#687)

### LINKED ISSUE #687: Licensing: Can't apply code fix after installing license
Found on mac (probably not relevant) during Happy Path testing.

1. Start with a clean machine.
2. Open VS Code.
3. Attempt to apply code fix. This correctly fails because there is no registered license.
4. Build. This opens license registration. Register trial license.
5. Attempt to apply code fix. This still fails with the same message, even though there is now a registered license.

After restarting VS code, it starts behaving correctly.

PostSharpAgent:
## Summary

Originally addressed #687 (code fixes unavailable after license registration). After review by @gfraiteur, the design-time licensing changes were reverted because **licensing is intentionally not enforced at design-time** — code fixes are offered unconditionally.

## Current State

The PR now contains only a single Backstage test: `ConsumptionServiceRaisesChangedEventOnLicenseRegistration`, which verifies that `LicenseConsumptionService.Changed` is raised when a license is registered via `UserProfileLicenseSource`. This event was previously untested.

### Reverted Changes
- ~~`DesignTimeServices.cs`: `AddLicensing = true`~~ — reverted to `false`
- ~~`DesignTimeAspectPipelineFactory.cs`: license change subscription~~ — removed
- ~~`ConsumptionServiceDetectsNewlyRegisteredLicense` test~~ — removed (redundant)

## Analysis of Original Issue

The original bug ("can't apply code fix after installing license") is not caused by a design-time licensing check, since none exists. The root cause may be different from what was investigated, or it may already be resolved.

## Test plan
- [x] `ConsumptionServiceRaisesChangedEventOnLicenseRegistration` passes on both net8.0 and net472
- [x] DesignTime project builds with zero warnings after revert

### DIFF
--- a/Metalama.Backstage/src/tests/Metalama.Backstage.Tests/Licensing/Consumption/LicenseConsumptionAfterRegistrationTests.cs
+++ b/Metalama.Backstage/src/tests/Metalama.Backstage.Tests/Licensing/Consumption/LicenseConsumptionAfterRegistrationTests.cs
@@ -3,7 +3,6 @@
 // Refer to LICENSE.md in the repository root for complete details.
 
 using Metalama.Backstage.Licensing.Consumption;
-using Metalama.Backstage.Licensing.Consumption.Requirements;
 using Metalama.Backstage.Licensing.Consumption.Sources;
 using Xunit;
 using Xunit.Abstractions;
@@ -11,7 +10,7 @@ using Xunit.Abstractions;
 namespace Metalama.Backstage.Tests.Licensing.Consumption;
 
 /// <summary>
-/// Tests that the <see cref="LicenseConsumptionService"/> correctly picks up license changes
+/// Tests that the <see cref="LicenseConsumptionService.Changed"/> event is raised
 /// when a license is registered after the service was created.
 /// Regression test for https://github.com/metalama/Metalama/issues/687.
 /// </summary>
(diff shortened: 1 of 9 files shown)


gfraiteur (changes requested):
After reviewing @PostSharpAgent's analysis, my understanding is that licensing is intentionally skipped and left unenforced at design time, and enforced only at compile time. I.e. the code fixes should be offered even when no license is registered.

The tests added in this PR are likely redundant - please check. Maybe the test that checks whether the event is raised is still relevant.

Verify that there is no design-time licensing code.

Then suggest whether the PR should be cancelled.

PostSharpAgent:
## Review Response

@gfraiteur — You're correct. After thorough verification:

1. **No design-time licensing code exists.** The code fix pipeline (`CodeFixesDesignTimeExtension.Initialize()`, `PremiumCodeFixProviderExtension.ProvideCodeFixesAsync()`) has zero licensing checks. Code fixes are offered unconditionally at design time, and licensing is enforced only at build time via the `VerifyMetalamaLicense` MSBuild task.

2. **Changes reverted.** I've reverted `AddLicensing = true` in `DesignTimeServices.cs` and removed the `ILicenseConsumptionService.Changed` subscription from `DesignTimeAspectPipelineFactory.cs`.

3. **Tests:** Removed the redundant `ConsumptionServiceDetectsNewlyRegisteredLicense` test (tests existing behavior that already works). Kept `ConsumptionServiceRaisesChangedEventOnLicenseRegistration` since no other test verifies that the `LicenseConsumptionService.Changed` event is raised when a license is registered via `UserProfileLicenseSource`.

**Regarding the original issue:** Since licensing is intentionally not enforced at design-time, the original bug report ("can't apply code fix after installing license") may have a different root cause than what was investigated, or it may already be fixed. The PR should likely be cancelled unless the `Changed` event test is considered valuable enough to keep on its own.

— Claude for @gfraiteur

PR #1371, opened 2026-03-08
Fix implementing introduced generic interface on introduced generic type (#598)

### LINKED ISSUE #598: Assertion failure when implementing generic interface on introduced generic type
```cs
// interface ITest<T>
var testInterface = builder.Advice.IntroduceInterface(
    builder.Target, 
    "ITest", 
    buildType: b => 
    { 
        var t = b.AddTypeParameter("T");
    });

// class Test<U>
var testImplementation = builder.Advice.IntroduceClass(
    builder.Target,
    "Test",
    buildType: b =>
    {
        var t = b.AddTypeParameter("U");
    });

// class Test<U> : ITest<U>
// FAILS when populating AllImplementedInterfaces for ITest<U> (which is a type instance and not the definition).
testImplementation.ImplementInterface(testInterface.Declaration.MakeGenericInstance(testImplementation .Declaration.TypeParameters[0]) ); 
```

PostSharpAgent:
## Summary

Fixes #598 - Implementing introduced generic interface on an introduced generic type fails on assertion.

**Root cause:** Multiple bugs in generic context handling for introduced types:

1. **CompilationModel.GetMemberCollection** asserted `declaration.IsDefinition`, but `IntroducedNamedType` passes `this.Ref` which can be a constructed ref (e.g., `Test<int>`). Fixed to normalize non-definition refs to their definition, since the dictionary is keyed by definition refs and the generic context is handled by the collection's facade.

2. **IntroducedRef.SelectGenericContext** threw `InvalidOperationException` when both the ref's context and the passed context were non-empty (e.g., `ITest<U>` ref has context `{T→U}`, and `Test<int>` passes context `{U→int}`). Fixed to compose contexts using `GenericContext.Map()`, producing the correct composed context `{T→int}`.

3. **IntroducedNamedType.TypeArguments** used `MapType(t.ToRef())` which created an `IntroducedTypeParameter` with context instead of actually performing the substitution. Fixed to resolve the type parameter first, then map through the generic context.

**Files changed:**
- `CompilationModel.Members.cs` - Normalize non-definition refs in `GetMemberCollection`
- `IntroducedRef.cs` - Combine non-empty generic contexts instead of throwing
- `IntroducedNamedType.cs` - Fix `TypeArguments` resolution
- Test file + expected output for `IntroducedGenericInterfaceOnIntroducedGenericType`

## Test plan

- [x] Regression test verifying `AllImplementedInterfaces`/`ImplementedInterfaces` on definition (`Test<U>` -> `ITest<U>`) and bound type (`Test<int>` -> `ITest<int>`)
- [x] All 2034 aspect tests pass
- [x] All 1560 unit tests pass
- [x] `Build.ps1 test` succeeds

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/GenericContexts/IntroducedGenericContext.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/GenericContexts/IntroducedGenericContext.cs
@@ -45,9 +45,7 @@ internal sealed class IntroducedGenericContext : GenericContext
         }
         else
         {
-            // The type parameter does not belong to this context's definition and there is no parent context.
-            // Return the type parameter unmapped, consistent with SymbolGenericContext behavior.
-            return typeParameter;
+            throw new ArgumentOutOfRangeException();
         }
     }
 
@@ -66,9 +64,7 @@ internal sealed class IntroducedGenericContext : GenericContext
         }
         else
         {
-            // The type parameter does not belong to this context's definition and there is no parent context.
-            // Return the type parameter unmapped, consistent with SymbolGenericContext behavior.
-            return compilation.Factory.GetIType( typeParameterSymbol );
+            throw new ArgumentOutOfRangeException();
         }
     }
 
--- a/Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/GenericContexts/IntroducedGenericContext.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/GenericContexts/IntroducedGenericContext.cs
@@ -45,7 +45,9 @@ internal sealed class IntroducedGenericContext : GenericContext
         }
         else
         {
-            throw new ArgumentOutOfRangeException();
+            // The type parameter does not belong to this context's definition and there is no parent context.
+            // Return the type parameter unmapped, consistent with SymbolGenericContext behavior.
+            return typeParameter;
         }
     }
 
@@ -64,7 +66,9 @@ internal sealed class IntroducedGenericContext : GenericContext
         }
         else
         {
-            throw new ArgumentOutOfRangeException();
+            // The type parameter does not belong to this context's definition and there is no parent context.
+            // Return the type parameter unmapped, consistent with SymbolGenericContext behavior.
+            return compilation.Factory.GetIType( typeParameterSymbol );
         }
     }
 
--- a/Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Introductions/Introduced/IntroducedNamedType.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Introductions/Introduced/IntroducedNamedType.cs
@@ -207,7 +207,16 @@ internal sealed class IntroducedNamedType : IntroducedMemberOrNamedType, INamedT
         => new TypeParameterList( this, this._namedTypeBuilderData.TypeParameters.Select( t => t.ToRef() ).ToReadOnlyList() );
 
     [Memo]
-    public IReadOnlyList<IType> TypeArguments => this._namedTypeBuilderData.TypeParameters.SelectAsImmutableArray( t => this.MapType( t.ToRef() ) );
+    public IReadOnlyList<IType> TypeArguments
+        => this._namedTypeBuilderData.TypeParameters.SelectAsImmutableArray(
+            t =>
+            {
+                // First resolve the type parameter without context to get the ITypeParameter declaration.
+                var typeParam = (IType) t.ToRef().GetTarget( this.Compilation );
+
+                // Then map it through the generic context, which substitutes e.g. T -> int.
+                return this.GenericContext.Map( typeParam ) ?? typeParam;
+            } );
 
     public bool IsGeneric => this._namedTypeBuilderData.TypeParameters.Length > 0;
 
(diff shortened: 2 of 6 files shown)


Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Introductions/Introduced/IntroducedNamedType.cs
```diff
@@ -51,12 +51,12 @@ public IntroducedNamedType( NamedTypeBuilderData builderData, CompilationModel c
     public IImplementedInterfaceCollection AllImplementedInterfaces
         => new AllImplementedInterfacesCollection(
             this,
-            this.Compilation.GetAllInterfaceImplementationCollection( this.Ref, false ) );
+            this.Compilation.GetAllInterfaceImplementationCollection( this.Ref.DefinitionRef, false ) );
```
gfraiteur: This seems incorrect to me because IntroducedNamedType can have type arguments (i.e. it can be a bound type) and in this case AllImplementedInterfaces and ImplementedInterfaces are supposed to return the list of interfaces implemented by the generic type instances i.e. with substituted type parameters.

I'm requesting additional tests that show that AllImplementedInterfaces and ImplementedInterfaces  return bound interface types in case the current type is bound (and not the generic type instance).

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Introductions/Introduced/IntroducedNamedType.cs
PostSharpAgent: You were right that this needed more work. Using `DefinitionRef` is actually correct here (it matches how `SourceNamedTypeImpl` uses `OriginalDefinition` to query interfaces), but the deeper issue was that three other components weren't handling generic contexts properly:

1. **`IntroducedGenericContext.Map`** threw when a type parameter didn't belong to the definition — now returns unmapped (matching `SymbolGenericContext`).
2. **`IntroducedRef.SelectGenericContext`** threw when both contexts were non-empty — now composes them via `GenericContext.Map()`.
3. **`IntroducedNamedType.TypeArguments`** wasn't resolving through the generic context — now properly substitutes type parameters.

The test now verifies both the definition case (`Test<U>.AllImplementedInterfaces` → `ITest<U>`) and the bound type case (`Test<int>.AllImplementedInterfaces` → `ITest<int>`), confirming that type parameter substitution works correctly end-to-end.

— Claude for @nicknash

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/GenericContexts/IntroducedGenericContext.cs
```diff
@@ -45,7 +45,9 @@ internal override IType Map( ITypeParameter typeParameter )
         }
         else
         {
-            throw new ArgumentOutOfRangeException();
+            // The type parameter does not belong to this context's definition and there is no parent context.
+            // Return the type parameter unmapped, consistent with SymbolGenericContext behavior.
```
gfraiteur: Throwing ArgumentOutOfRangeException was the right behavior. Returning the input typeParameter is conceptually incorrect.

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/GenericContexts/IntroducedGenericContext.cs
```diff
@@ -64,7 +66,9 @@ protected override IType Map( ITypeParameterSymbol typeParameterSymbol, Compilat
...
+            return compilation.Factory.GetIType( typeParameterSymbol );
```
gfraiteur: Ditto. Conceptually incorrect.

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Introductions/Introduced/IntroducedNamedType.cs
gfraiteur: I don't think the solution is correct. `this.Ref` contains the generic context if I remember properly, so the bug is rather behind GetAllInterfaceImplementationCollection rather than downstream.

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/GenericContexts/IntroducedGenericContext.cs
PostSharpAgent: Agreed and reverted. Both `Map(ITypeParameter)` and `Map(ITypeParameterSymbol, ...)` now throw `ArgumentOutOfRangeException` again. The actual root cause was in `GetMemberCollection` — see below.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/GenericContexts/IntroducedGenericContext.cs
PostSharpAgent: Reverted.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Introductions/Introduced/IntroducedNamedType.cs
PostSharpAgent: You were right. Reverted `this.Ref.DefinitionRef` back to `this.Ref`.

The root fix is now in `CompilationModel.GetMemberCollection`: instead of asserting `declaration.IsDefinition`, it normalizes to the definition ref when a non-definition ref is passed. This way `IntroducedNamedType` can pass `this.Ref` (which carries the generic context), and `GetMemberCollection` handles the dictionary lookup correctly.

The `IntroducedRef.SelectGenericContext` combining fix (commit 44dd8f9) is still needed: when `Test<int>.AllImplementedInterfaces` iterates, the collection resolves refs like `ITest<U>` (with context `{T→U}`) using the collection's facade context `{U→int}`. Both contexts are non-empty and must be composed via `GenericContext.Map()`: `{T→U}` ∘ `{U→int}` = `{T→int}`, producing the correct `ITest<int>`.

— Claude for @gfraiteur

gfraiteur (approved):

PR #1372, opened 2026-03-09
Fix assembly same-name collision in CompileTimeDomain (#579)

### LINKED ISSUE #579: "Assembly with same name is already loaded" when updating premium package version independently
This might be related to building several projects where we have different AssemblyLoadContext.

```text
Extensions.Validation.dll': Assembly with same name is already loaded)
---> System.IO.FileLoadException: Cannot load 'REDACTED(Path)metalama\net6.0\Metalama.Extensions.Validation.dll': Assembly with same name is already loaded
---> System.IO.FileLoadException: Assembly with same name is already loaded
   at System.Runtime.Loader.AssemblyLoadContext.<LoadFromPath>g____PInvoke|5_0(IntPtr __ptrNativeAssemblyBinder_native, UInt16* __ilPath_native, UInt16* __niPath_native, ObjectHandleOnStack __retAssembly_native)
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromAssemblyPath(String assemblyPath)
   at Metalama.Framework.Engine.Utilities.AssemblyLoader.LoadFromPath(String assemblyPath) in C:\src\Metalama-2025.1\Metalama\Metalama.Framework\src\Metalama.Framework.Engine\Utilities\AssemblyLoader.cs:line 222
   at Metalama.Framework.Engine.CompileTime.CompileTimeDomain.LoadAssemblyCore(String path, LoadAssemblyOptions options) in C:\src\Metalama-2025.1\Metalama\Metalama.Framework\src\Metalama.Framework.Engine\CompileTime\CompileTimeDomain.cs:line 169
   at Metalama.Framework.Engine.CompileTime.CompileTimeDomain.LoadAssemblyCore(String path, LoadAssemblyOptions options) in C:\src\Metalama-2025.1\Metalama\Metalama.Framework\src\Metalama.Framework.Engine\CompileTime\CompileTimeDomain.cs:line 174
   at Metalama.Framework.Engine.CompileTime.CompileTimeDomain.LoadAssembly(String path, LoadAssemblyOptions options) in C:\src\Metalama-2025.1\Metalama\Metalama.Framework\src\Metalama.Framework.Engine\CompileTime\CompileTimeDomain.cs:line 139
   at Metalama.Framework.Engine.Extensibility.ExtensionLoaderHelper.<>c__DisplayClass2_0.<LoadExtensionAssemblies>b__0(String path) in C:\src\Metalama-2025.1\Metalama\Metalama.Framework\src\Metalama.Framework.Engine\Extensibility\ExtensionLoaderHelper.cs:line 46
```

The root cause is the the AssemblyLoadContext can contain a single assembly of each name, where the name is the simple name. When rebuilding Premium, we generate different assemblies, which cause a failure when trying to load them in the existing AssemblyLoadContext. The workaround is to do `metalama kill`. The solution is to change the semantic of `ICompileTimeDomainFactory.CreateDomain` so that it accepts all assembly paths before creating the domain, instead of having dynamic loading as needed with `CompileTimeDomain.LoadAssembly`.

PostSharpAgent:
## Summary

Fixes #579 - Exception when loading a different version of an extension assembly with the same simple name into an existing CompileTimeDomain.

When rebuilding Premium packages without changing the Metalama.Framework version, the design-time CompileTimeDomain (which was a global singleton) throws FileLoadException because it cannot hold two assemblies with the same simple name but different versions.

### Root cause

CompileTimeDomain was registered as an IGlobalService — a singleton shared across all projects in the design-time process. When different projects need different versions of the same extension assembly, the shared domain cannot accommodate both.

### Approach: Per-project domain management via factory (v5 - final)

1. **CompileTimeDomain** — Removed IGlobalService interface. Added IsCompatibleWithAssemblies(paths) method. Added Guid property for weak reference tracking. Documented lifecycle policy (GC-based in production, explicit dispose in tests).
2. **ICompileTimeDomainFactory** — Added GetOrCreateDomain(assemblyPaths) method. CreateDomain() documented as being for test infrastructure only.
3. **DefaultCompileTimeDomainFactory / TestCompileTimeDomainFactory** — Use ConcurrentDictionary<Guid, WeakReference<CompileTimeDomain>> to track all domains. GetOrCreateDomain checks all alive domains for compatibility, cleaning up dead references.
4. **IExtensionLoader** — Added GetExtensionAssemblyPaths to the interface, removing the need to instantiate ExtensionLoaderBase directly.
5. **AspectPipeline** — Uses IExtensionLoader service instead of instantiating ExtensionLoaderBase. Gets factory per-project via GetOrCreateDomain.
6. **DesignTimeExtensionManager** — Removed redundant ExtensionLoaderBase field, uses IExtensionLoader service directly.
7. **Serialization** — BaseCompileTimeSerializationBinder and CompileTimeSerializationBinder now document nullable domain parameter.

## Progress

- [x] Reproduce the bug with a regression test
- [x] Implement fix (v5: per-project domain via factory with WeakReference tracking)
- [x] Update all consumers
- [x] Build.ps1 build (zero warnings, zero errors)
- [x] Unit tests (1574 total: 1568 pass, 6 skipped, 0 failed)
- [x] Address all 4 rounds of review feedback
- [ ] TC build verification (unable to trigger manually, awaiting CI)

## Test plan

- [x] LoadAssembly_SameIdentity_ReturnsCachedAssembly
- [x] LoadAssembly_DifferentNames_LoadsBoth
- [x] IsCompatibleWithAssemblies_Compatible_ReturnsTrue
- [x] IsCompatibleWithAssemblies_VersionConflict_ReturnsFalse
- [x] GetOrCreateDomain_VersionConflict_CreatesNewDomain (regression test for #579)
- [x] GetOrCreateDomain_Compatible_ReturnsSameDomain
- [x] All 1574 unit tests pass on net8.0

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Extensibility/DesignTimeExtensionManager.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Extensibility/DesignTimeExtensionManager.cs
@@ -17,7 +17,6 @@ using Metalama.Framework.Engine.Utilities.Threading;
 using Metalama.Framework.Services;
 using System.Collections.Concurrent;
 using System.Collections.Immutable;
-using System.Linq;
 
 namespace Metalama.Framework.DesignTime.Extensibility;
 
@@ -67,9 +66,8 @@ public sealed class DesignTimeExtensionManager : IGlobalService
         lock ( this._extensions )
         {
             // Get a domain compatible with this project's design-time extension assemblies.
-            var extensionAssemblyPaths = this._extensionLoaderBase
-                .GetExtensionAssemblyPaths( options.DesignTimeExtensionAssemblies )
-                .ToList();
+            var extensionAssemblyPaths = new List<string>(
+                this._extensionLoaderBase.GetExtensionAssemblyPaths( options.DesignTimeExtensionAssemblies ) );
 
             var domain = this._domainFactory.GetOrCreateDomain( extensionAssemblyPaths );
 
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Extensibility/DesignTimeExtensionManager.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Extensibility/DesignTimeExtensionManager.cs
@@ -17,6 +17,7 @@ using Metalama.Framework.Engine.Utilities.Threading;
 using Metalama.Framework.Services;
 using System.Collections.Concurrent;
 using System.Collections.Immutable;
+using System.Linq;
 
 namespace Metalama.Framework.DesignTime.Extensibility;
 
@@ -26,7 +27,8 @@ public sealed class DesignTimeExtensionManager : IGlobalService
     private readonly DesignTimeProcessKind _processKind;
 
     private readonly ConcurrentDictionary<int, int> _processedOptions = new();
-    private readonly CompileTimeDomain _domain;
+    private readonly ICompileTimeDomainFactory _domainFactory;
+    private readonly ExtensionLoaderBase _extensionLoaderBase;
     private readonly ConcurrentDictionary<Type, IDesignTimeExtension> _extensions = new();
     private readonly IRpcServiceProviderServerEndpointProvider? _rpcServiceProviderServerEndpointProvider;
     private readonly IExtensionLoader _extensionLoader;
@@ -39,7 +41,8 @@ public sealed class DesignTimeExtensionManager : IGlobalService
     {
         this._processKind = processKind;
         this.ServiceProvider = serviceProvider;
-        this._domain = serviceProvider.GetRequiredService<CompileTimeDomain>();
+        this._domainFactory = serviceProvider.GetRequiredService<ICompileTimeDomainFactory>();
+        this._extensionLoaderBase = new ExtensionLoaderBase( serviceProvider );
         this._rpcServiceProviderServerEndpointProvider = serviceProvider.GetService<IRpcServiceProviderServerEndpointProvider>();
         this._extensionLoader = serviceProvider.GetRequiredService<IExtensionLoader>();
         this._logger = serviceProvider.GetLoggerFactory().GetLogger( nameof(DesignTimeExtensionManager) );
@@ -63,7 +66,14 @@ public sealed class DesignTimeExtensionManager : IGlobalService
 
         lock ( this._extensions )
         {
-            var extensions = this._extensionLoader.GetExtensionTypes( options, this._domain, ExtensionKinds.DesignTime, NullDiagnosticAdder.Instance );
+            // Get a domain compatible with this project's design-time extension assemblies.
+            var extensionAssemblyPaths = this._extensionLoaderBase
+                .GetExtensionAssemblyPaths( options.DesignTimeExtensionAssemblies )
+                .ToList();
+
+            var domain = this._domainFactory.GetOrCreateDomain( extensionAssemblyPaths );
+
+            var extensions = this._extensionLoader.GetExtensionTypes( options, domain, ExtensionKinds.DesignTime, NullDiagnosticAdder.Instance );
 
             foreach ( var extensionType in extensions )
             {
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Extensibility/DesignTimeExtensionManager.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Extensibility/DesignTimeExtensionManager.cs
@@ -27,7 +27,6 @@ public sealed class DesignTimeExtensionManager : IGlobalService
 
     private readonly ConcurrentDictionary<int, int> _processedOptions = new();
     private readonly ICompileTimeDomainFactory _domainFactory;
-    private readonly ExtensionLoaderBase _extensionLoaderBase;
     private readonly ConcurrentDictionary<Type, IDesignTimeExtension> _extensions = new();
     private readonly IRpcServiceProviderServerEndpointProvider? _rpcServiceProviderServerEndpointProvider;
     private readonly IExtensionLoader _extensionLoader;
@@ -41,7 +40,6 @@ public sealed class DesignTimeExtensionManager : IGlobalService
         this._processKind = processKind;
         this.ServiceProvider = serviceProvider;
         this._domainFactory = serviceProvider.GetRequiredService<ICompileTimeDomainFactory>();
-        this._extensionLoaderBase = new ExtensionLoaderBase( serviceProvider );
         this._rpcServiceProviderServerEndpointProvider = serviceProvider.GetService<IRpcServiceProviderServerEndpointProvider>();
         this._extensionLoader = serviceProvider.GetRequiredService<IExtensionLoader>();
         this._logger = serviceProvider.GetLoggerFactory().GetLogger( nameof(DesignTimeExtensionManager) );
@@ -67,7 +65,7 @@ public sealed class DesignTimeExtensionManager : IGlobalService
         {
             // Get a domain compatible with this project's design-time extension assemblies.
             var extensionAssemblyPaths = new List<string>(
-                this._extensionLoaderBase.GetExtensionAssemblyPaths( options.DesignTimeExtensionAssemblies ) );
+                this._extensionLoader.GetExtensionAssemblyPaths( options.DesignTimeExtensionAssemblies ) );
 
             var domain = this._domainFactory.GetOrCreateDomain( extensionAssemblyPaths );
 
--- a/Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
@@ -32,12 +32,24 @@ namespace Metalama.Framework.Engine.CompileTime
     /// Domains are managed by <see cref="ICompileTimeDomainFactory"/>, which decides whether to reuse an existing domain
     /// or create a new one based on assembly compatibility.
     /// </summary>
+    /// <remarks>
+    /// <para><b>Lifecycle:</b> Domains are created by <see cref="ICompileTimeDomainFactory"/> and are not disposed explicitly
+    /// during normal operation. When a domain becomes incompatible with new assembly versions, the factory creates a new domain
+    /// and the old domain becomes eligible for garbage collection once no compilation holds a reference to it. In tests,
+    /// domains are disposed explicitly via <c>TestContext.Dispose</c>.</para>
+    /// </remarks>
     public class CompileTimeDomain : IDisposable
     {
         private static readonly ConcurrentDictionary<string, object> _locksByPath = new();
         private static int _nextDomainId;
         private readonly ConcurrentDictionary<AssemblyIdentity, Assembly> _assemblyCache = new();
         private readonly int _domainId = Interlocked.Increment( ref _nextDomainId );
+
+        /// <summary>
+        /// Gets the unique identifier for this domain instance, used by <see cref="ICompileTimeDomainFactory"/>
+        /// to track domains via weak references.
+        /// </summary>
+        public Guid Guid { get; } = Guid.NewGuid();
         private readonly ILogger _logger;
         private readonly object _sync = new();
         private readonly ConcurrentDictionary<string, (Assembly Assembly, AssemblyIdentity Identity)> _assembliesByName = new();
@@ -306,7 +318,7 @@ namespace Metalama.Framework.Engine.CompileTime
                     if ( existingName.Version != assemblyName.Version
                          || !( existingName.GetPublicKeyToken() ?? [] ).SequenceEqual( assemblyName.GetPublicKeyToken() ?? [] ) )
                     {
-                        this._logger.Warning?.Log(
+                        this._logger.Trace?.Log(
                             $"Domain {this._domainId} is incompatible with assembly '{assemblyName}' at '{path}': " +
                             $"already-loaded assembly has identity '{existingEntry.Identity}'." );
 
--- a/Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
@@ -118,6 +118,20 @@ namespace Metalama.Framework.Engine.CompileTime
                 return existingAssembly;
             }
 
+            // Check if an assembly with the same simple name (but different version) is already loaded.
+            // An AssemblyLoadContext cannot hold two assemblies with the same simple name, so we must
+            // return the previously loaded assembly instead of attempting to load the new one, which
+            // would throw a FileLoadException. This happens when extension assemblies are rebuilt
+            // with different versions while the domain is still alive.
+            if ( assemblyName.Name != null
+                 && this._assembliesByName.TryGetValue( assemblyName.Name, out var existingEntry ) )
+            {
+                this._logger.Warning?.Log(
+                    $"Cannot load assembly '{assemblyName}' from '{path}' because a different version ('{existingEntry.Identity}') is already loaded in the domain. Returning the existing assembly." );
+
+                return existingEntry.Assembly;
+            }
+
             // The assembly might already be loaded in the AppDomain or AssemblyLoadContext.
             if ( this._assemblyLoader!.TryGetLoadedAssembly( assemblyName, out existingAssembly ) )
             {
--- a/Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
@@ -10,7 +10,6 @@ using Metalama.Framework.Engine.Services;
 using Metalama.Framework.Engine.Utilities.AssemblyLoaders;
 using Metalama.Framework.Engine.Utilities.Diagnostics;
 using Metalama.Framework.Engine.Utilities.Roslyn;
-using Metalama.Framework.Services;
 using Microsoft.CodeAnalysis;
 using System;
 using System.Collections.Concurrent;
@@ -29,11 +28,11 @@ namespace Metalama.Framework.Engine.CompileTime
 {
     /// <summary>
     /// Tracks compile-time assemblies belonging to the same domain and implements CLR assembly resolution.
-    /// The number of <see cref="CompileTimeDomain"/> in an <see cref="AppDomain"/>
-    /// depends on the scenario: typically one per project at compile time, one per <see cref="AppDomain"/> at design time, and one per test
-    /// at testing time.
+    /// Each <see cref="CompileTimeDomain"/> has a single <see cref="AssemblyLoader"/> (and, on .NET 5+, a single <c>AssemblyLoadContext</c>).
+    /// Domains are managed by <see cref="ICompileTimeDomainFactory"/>, which decides whether to reuse an existing domain
+    /// or create a new one based on assembly compatibility.
     /// </summary>
-    public class CompileTimeDomain : IDisposable, IGlobalService
+    public class CompileTimeDomain : IDisposable
     {
         private static readonly ConcurrentDictionary<string, object> _locksByPath = new();
         private static int _nextDomainId;
@@ -42,13 +41,6 @@ namespace Metalama.Framework.Engine.CompileTime
         private readonly ILogger _logger;
         private readonly object _sync = new();
         private readonly ConcurrentDictionary<string, (Assembly Assembly, AssemblyIdentity Identity)> _assembliesByName = new();
-        private readonly string? _debugName;
-
-        /// <summary>
-        /// Assembly loaders that have been retired (replaced by a new loader due to version conflicts)
-        /// but are kept alive so that assemblies loaded from them remain valid for concurrent compilations.
-        /// </summary>
-        private readonly List<AssemblyLoader> _retiredAssemblyLoaders = new();
 
         private AssemblyLoader? _assemblyLoader;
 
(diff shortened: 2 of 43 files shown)


Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
```diff
@@ -118,6 +118,20 @@ public CompileTimeDomain( GlobalServiceProvider serviceProvider, string? debugNa
...
 
+            // Check if an assembly with the same simple name (but different version) is already loaded.
+            // An AssemblyLoadContext cannot hold two assemblies with the same simple name, so we must
+            // return the previously loaded assembly instead of attempting to load the new one, which
+            // would throw a FileLoadException. This happens when extension assemblies are rebuilt
+            // with different versions while the domain is still alive.
+            if ( assemblyName.Name != null
+                 && this._assembliesByName.TryGetValue( assemblyName.Name, out var existingEntry ) )
+            {
+                this._logger.Warning?.Log(
```
gfraiteur: This is not an acceptable behavior.

It seems the real fix is much more complex -- if we know all assemblies in the load context upfront, then we should consider, upfront, if the previous assemblyloadcontext can be reused - otherwise, a new one should be created.

gfraiteur (changes requested):
Explore the option of creating a new AssemblyLoadContext if the current one is not compatible. This might require considerable analysis and changes, but try it.

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
PostSharpAgent: Addressed. The previous approach (returning the old assembly) has been replaced:

1. Added `EnsureCompatibleWithAssemblies(IEnumerable<string> assemblyPaths)` to `CompileTimeDomain` — accepts all assembly paths upfront and checks if the current domain is compatible.
2. When a conflict is detected (same simple name, different identity), the domain resets: disposes the old `AssemblyLoader` (and `AssemblyLoadContext` in `UnloadableCompileTimeDomain`), creates fresh instances, and clears all caches.
3. `ExtensionLoaderBase` now collects all extension paths first and calls `EnsureCompatibleWithAssemblies` before loading any assemblies.
4. `UnloadableCompileTimeDomain.ResetAssemblyLoader()` also unloads and recreates its `AssemblyLoadContext`.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
```diff
@@ -273,6 +276,69 @@ protected virtual void Dispose( bool disposing )
...
+
+            if ( incompatibleAssemblies.Count > 0 )
+            {
+                foreach ( var (path, name, existingIdentity) in incompatibleAssemblies )
+                {
+                    this._logger.Warning?.Log(
+                        $"Assembly '{name}' at '{path}' is incompatible with already-loaded assembly '{existingIdentity}'. Resetting the domain." );
+                }
+
+                this.ResetAssemblyLoader();
```
gfraiteur: Disposing of the current AssemblyLoadContext is not an acceptable behavior since several compilations can happen concurrently, each with its own AssemblyLoadContext. So what we need is support for multiple AssemblyLoadContext in parallel, probably with reference counting.

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
PostSharpAgent: Addressed. The approach has been fundamentally changed:

**Instead of disposing the old AssemblyLoadContext**, the domain now **retires** it to a `_retiredAssemblyLoaders` list (and `_retiredAssemblyLoadContexts` in `UnloadableCompileTimeDomain`). The retired loaders/ALCs are kept alive for the entire lifetime of the domain, so concurrent compilations that loaded assemblies from them continue to work without disruption.

Key changes:
1. **`CompileTimeDomain`**: Added `_retiredAssemblyLoaders` list. `RetireCurrentAssemblyLoader()` moves the current loader to this list instead of disposing it. New `CreateAssemblyLoader()` virtual method for subclass extensibility. `IsCollectible()` checks both current and retired loaders.
2. **`UnloadableCompileTimeDomain`**: Added `_retiredAssemblyLoadContexts` list. Overrides `OnAssemblyLoaderRetired()` to retire the current ALC to the list without unloading it. All retired ALCs are unloaded only on domain disposal.
3. **Tests**: Added `RetiredLoaderAssemblies_RemainValidForConcurrentUse` test that simulates concurrent compilation scenario — verifies assemblies from retired loader remain valid after a new loader is created.

This supports multiple ALCs in parallel. The "reference counting" is implicit — the CLR keeps assemblies alive as long as they're referenced, and the retired loaders/ALCs stay in the list until the domain is disposed.

— Claude for @PostSharpAgent

gfraiteur (changes requested):
I still don't think this is the right approach. A CompileTimeDomain should have a single AssemblyLoadContext. The problem now is that there is a singleton instance of CompileTimeDomain. This should be refactored more deeply. The domain management should be in ICompileTimeDomainFactory. It should have a GetOrCreateDomain that accepts the list of assemblies and decides whether to give a new domain or an existing one. CompileTimeDomain should no longer be a GlobalService (because it's project-specific). So a more general refactoring is needed.

Metalama.Framework/src/Metalama.Framework.DesignTime/Extensibility/DesignTimeExtensionManager.cs
```diff
@@ -26,7 +26,8 @@ public sealed class DesignTimeExtensionManager : IGlobalService
     private readonly DesignTimeProcessKind _processKind;
 
     private readonly ConcurrentDictionary<int, int> _processedOptions = new();
-    private readonly CompileTimeDomain _domain;
+    private readonly ICompileTimeDomainFactory _domainFactory;
+    private readonly ExtensionLoaderBase _extensionLoaderBase;
```
gfraiteur: Redundant with _extensionLoader. If methods from ExtensionLoaderBase  are required, add them to IExtensionLoader .

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/Serialization/BaseCompileTimeSerializationBinder.cs
```diff
@@ -36,9 +36,9 @@ static BaseCompileTimeSerializationBinder()
             .ToImmutableDictionary( x => x.Key, x => x.OrderByDescending( a => a.Version ).First().ToString() );
     }
 
-    protected BaseCompileTimeSerializationBinder( in ProjectServiceProvider serviceProvider )
+    protected BaseCompileTimeSerializationBinder( CompileTimeDomain? domain, in ProjectServiceProvider serviceProvider )
     {
-        this._domain = serviceProvider.Global.GetRequiredService<CompileTimeDomain>();
+        this._domain = domain;
```
gfraiteur, line 45: Document why `domain` might be null.

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/Serialization/CompileTimeSerializationBinder.cs
```diff
@@ -13,7 +13,7 @@ internal sealed class CompileTimeSerializationBinder : BaseCompileTimeSerializat
     private static readonly string _systemAssemblyName = typeof(object).Assembly.FullName.AssertNotNull();
     private readonly IDeserializationSurrogateProvider? _deserializationSurrogateProvider;
 
-    public CompileTimeSerializationBinder( in ProjectServiceProvider serviceProvider, CompileTimeProject? project ) : base( serviceProvider )
+    public CompileTimeSerializationBinder( CompileTimeDomain? domain, in ProjectServiceProvider serviceProvider, CompileTimeProject? project ) : base( domain, serviceProvider )
     {
```
gfraiteur, line 21: Document why domain might be null.

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
```diff
@@ -28,11 +28,11 @@ namespace Metalama.Framework.Engine.CompileTime
...
     /// Tracks compile-time assemblies belonging to the same domain and implements CLR assembly resolution.
-    /// The number of <see cref="CompileTimeDomain"/> in an <see cref="AppDomain"/>
-    /// depends on the scenario: typically one per project at compile time, one per <see cref="AppDomain"/> at design time, and one per test
-    /// at testing time.
+    /// Each <see cref="CompileTimeDomain"/> has a single <see cref="AssemblyLoader"/> (and, on .NET 5+, a single <c>AssemblyLoadContext</c>).
+    /// Domains are managed by <see cref="ICompileTimeDomainFactory"/>, which decides whether to reuse an existing domain
+    /// or create a new one based on assembly compatibility.
     /// </summary>
-    public class CompileTimeDomain : IDisposable, IGlobalService
+    public class CompileTimeDomain : IDisposable
```
gfraiteur, line 41: Comment the lifecycle policy. Do we rely on GC to dispose it by default, except in tests? I think so -- it avoids relying on reference counting.

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
```diff
@@ -286,5 +286,36 @@ internal void RegisterAssemblyPaths( ImmutableArray<string> systemAssemblyPaths
...
+                if ( assemblyName.Name != null
+                     && this._assembliesByName.TryGetValue( assemblyName.Name, out var existingEntry ) )
+                {
+                    var existingName = existingEntry.Assembly.GetName();
+
+                    if ( existingName.Version != assemblyName.Version
+                         || !( existingName.GetPublicKeyToken() ?? [] ).SequenceEqual( assemblyName.GetPublicKeyToken() ?? [] ) )
+                    {
+                        this._logger.Warning?.Log(
+                            $"Domain {this._domainId} is incompatible with assembly '{assemblyName}' at '{path}': " +
```
gfraiteur, line 336: trace, not warning

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/DefaultCompileTimeDomainFactory.cs
```diff
@@ -3,20 +3,47 @@
...
 
 namespace Metalama.Framework.Engine.CompileTime
 {
     [ExcludeFromCodeCoverage] // Not used in tests.
-    internal sealed class DefaultCompileTimeDomainFactory : ICompileTimeDomainFactory
+    internal sealed class DefaultCompileTimeDomainFactory : ICompileTimeDomainFactory, IDisposable
     {
         private readonly GlobalServiceProvider _serviceProvider;
+        private readonly object _sync = new();
+        private CompileTimeDomain? _currentDomain;
```
gfraiteur: I would rather maintain a list of WeakReference<CompileTimeDomain> with all domains. Or maybe a ConcurrentDictionary<Guid,WeakReference<CompileTimeDomain>> where the Guid is the CompileTimeDomain's Guid. So GetOrCreateDomain can return any domain that's still alive.

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/ICompileTimeDomainFactory.cs
```diff
@@ -4,18 +4,28 @@
...
+    /// Manages the lifecycle of <see cref="CompileTimeDomain"/> instances. The factory decides whether to reuse
+    /// an existing domain or create a new one based on assembly compatibility.
     /// </summary>
     [PublicAPI] // Used in Metalama.Try.
     public interface ICompileTimeDomainFactory : IGlobalService
     {
         /// <summary>
-        /// Gets an instance of <see cref="CompileTimeDomain"/>. 
+        /// Creates a new instance of <see cref="CompileTimeDomain"/>.
         /// </summary>
```
gfraiteur, line 22: When is this still used?

Metalama.Framework/src/Metalama.Framework.Engine/Pipeline/AspectPipeline.cs
```diff
@@ -121,13 +120,21 @@ protected virtual bool TryInitialize(
             return false;
         }
 
+        // Get a CompileTimeDomain that is compatible with the extension assemblies for this project.
+        // The factory either reuses an existing domain or creates a new one.
+        var extensionAssemblyPaths = new ExtensionLoaderBase( this.ServiceProvider.Global )
```
gfraiteur: ExtensionLoaderBase is supposed to be an abstract class and not instantiated there, but accessed as a service.

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.DesignTime/Extensibility/DesignTimeExtensionManager.cs
PostSharpAgent: Addressed. Removed `_extensionLoaderBase` field entirely. Added `GetExtensionAssemblyPaths` to `IExtensionLoader` interface (which both `ExtensionLoader` and `TestExtensionLoader` already implement through inheritance from `ExtensionLoaderBase`). Now using `_extensionLoader.GetExtensionAssemblyPaths()` directly.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/Serialization/BaseCompileTimeSerializationBinder.cs
PostSharpAgent: Addressed. Added XML doc comment on the constructor parameter explaining that `domain` is null when the serializer is created before a `CompileTimeProject` is available (e.g. during early pipeline initialization), and that type binding falls back to `Type.GetType()` in that case.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/Serialization/CompileTimeSerializationBinder.cs
PostSharpAgent: Addressed. Added XML doc comment on the constructor parameter, cross-referencing `BaseCompileTimeSerializationBinder` for the full explanation.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
PostSharpAgent: Addressed. Added a `<remarks>` block documenting the lifecycle policy: domains are not disposed explicitly during normal operation — when a domain becomes incompatible, the factory creates a new one and the old domain is collected by GC once no compilation holds a reference to it. In tests, domains are disposed explicitly via `TestContext.Dispose`.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
PostSharpAgent: Addressed. Changed from `this._logger.Warning?.Log` to `this._logger.Trace?.Log`.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/DefaultCompileTimeDomainFactory.cs
PostSharpAgent: Addressed. Both `DefaultCompileTimeDomainFactory` and `TestCompileTimeDomainFactory` now use `ConcurrentDictionary<Guid, WeakReference<CompileTimeDomain>>` to track all domains. `GetOrCreateDomain` iterates all alive domains to find a compatible one, cleaning up dead references along the way. Added a `Guid` property to `CompileTimeDomain` as the dictionary key. The factory no longer implements `IDisposable` since domains are GC-collected.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/ICompileTimeDomainFactory.cs
PostSharpAgent: Still used by \ (which creates a standalone domain for test infrastructure without assembly compatibility checks) and internally by the factory implementations' \ when creating new domains. Updated the XML doc to clarify this: production code should use \ instead.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Pipeline/AspectPipeline.cs
PostSharpAgent: Addressed. No longer instantiating `ExtensionLoaderBase` directly. Now resolving `IExtensionLoader` from the service provider (via `this.ServiceProvider.Global.GetRequiredService<IExtensionLoader>()`) and calling the new `GetExtensionAssemblyPaths` method on it.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/ICompileTimeDomainFactory.cs
PostSharpAgent: Still used by TestContext.Domain (which creates a standalone domain for test infrastructure without assembly compatibility checks) and internally by the factory implementations when creating new domains. Updated the XML doc to clarify this: production code should use GetOrCreateDomain instead.

-- Claude for @PostSharpAgent

gfraiteur (approved):

gfraiteur (changes requested):
Address copilot's feedback.

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
```diff
@@ -286,5 +298,36 @@ internal void RegisterAssemblyPaths( ImmutableArray<string> systemAssemblyPaths
...
+
+                if ( assemblyName.Name != null
+                     && this._assembliesByName.TryGetValue( assemblyName.Name, out var existingEntry ) )
+                {
+                    var existingName = existingEntry.Assembly.GetName();
+
+                    if ( existingName.Version != assemblyName.Version
+                         || !( existingName.GetPublicKeyToken() ?? [] ).SequenceEqual( assemblyName.GetPublicKeyToken() ?? [] ) )
+                    {
+                        this._logger.Trace?.Log(
```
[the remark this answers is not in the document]
PostSharpAgent: Addressed. Added a try/catch in `IsCompatibleWithAssemblies` for `FileNotFoundException`, `BadImageFormatException`, and `IOException`. When an assembly's metadata cannot be read, the domain is treated as compatible and the normal extension loader will report the diagnostic when it actually attempts to load the assembly.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/CompileTime/CompileTimeDomainTests.cs
```diff
@@ -0,0 +1,232 @@
...
+            var assembly1 = domain1.LoadAssembly( path1, null, new LoadAssemblyOptions { IsShared = true } );
+            Assert.Equal( new Version( 1, 0, 0, 0 ), assembly1.GetName().Version );
+
+            // Get a domain for v2 — should be a different domain since v1 is incompatible.
+            var domain2 = factory.GetOrCreateDomain( new[] { path2 } );
+            Assert.NotSame( domain1, domain2 );
+
+            // Load v2 in the new domain — should succeed.
+            var assembly2 = domain2.LoadAssembly( path2, null, new LoadAssemblyOptions { IsShared = true } );
+            Assert.Equal( new Version( 2, 0, 0, 0 ), assembly2.GetName().Version );
```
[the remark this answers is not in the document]
PostSharpAgent: Addressed. The `GetOrCreateDomain_VersionConflict_CreatesNewDomain` test now tracks both domains and disposes them in a finally block, with a guard to avoid double-disposing when they happen to be the same instance.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/CompileTime/CompileTimeDomainTests.cs
```diff
@@ -0,0 +1,232 @@
...
+            var pathB = CreateAssemblyOnDisk( tempDir, "ExtensionB", new Version( 1, 0, 0, 0 ), "public class B {}" );
+
+            // Get a domain and load A.
+            var domain1 = factory.GetOrCreateDomain( new[] { pathA } );
+            domain1.LoadAssembly( pathA, null, new LoadAssemblyOptions { IsShared = true } );
+
+            // Get a domain for B — should be the same domain since there's no conflict.
+            var domain2 = factory.GetOrCreateDomain( new[] { pathB } );
+            Assert.Same( domain1, domain2 );
+        }
```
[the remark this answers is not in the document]
PostSharpAgent: Addressed. The `GetOrCreateDomain_Compatible_ReturnsSameDomain` test now tracks `domain1` and disposes it in a finally block. Since domain1 and domain2 are the same instance in this test (that's the assertion), only one dispose is needed.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Testing.UnitTesting/TestCompileTimeDomainFactory.cs
```diff
@@ -4,27 +4,58 @@
...
+
+        this._domains.TryAdd( domain.Guid, new WeakReference<CompileTimeDomain>( domain ) );
+
+        return domain;
+    }
+
+    public CompileTimeDomain GetOrCreateDomain( IReadOnlyCollection<string> assemblyPaths )
+    {
+        // Clean up dead references and check for a compatible domain among all still-alive domains.
+        foreach ( var kvp in this._domains.ToArray() )
```
gfraiteur: This code should run within a lock to make sure we reuse domains even under concurrency.

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Testing.UnitTesting/TestCompileTimeDomainFactory.cs
PostSharpAgent: Addressed. Added a `lock` around the entire `GetOrCreateDomain` method in both `TestCompileTimeDomainFactory` and `DefaultCompileTimeDomainFactory`. This ensures that concurrent callers cannot both evaluate compatibility and create a new domain simultaneously — only one thread at a time can check existing domains and potentially create a new one.

— Claude for @PostSharpAgent

gfraiteur (approved):

PR #1376, opened 2026-03-09
Fix RemoveAttributes incorrectly removing #pragma (#606)

### LINKED ISSUE #606: RemoveAttributes incorrectly removes #pragma directives
This test:

```c#
public class MyAspect : TypeAspect
{
    public override void BuildAspect( IAspectBuilder<INamedType> builder )
    {
        builder.RemoveAttributes( GetType() );
    }
}

#pragma warning disable CS8618

[MyAspect]
internal class C
{
}
```

incorrectly removes the `#pragma warning disable` from the generated code.

PostSharpAgent:
## Summary

Fixes #606

`RemoveAttributes` incorrectly removes `#pragma warning disable` directives that precede the target declaration. The pragma trivia should be preserved when only the custom attribute is being removed.

### Root cause

In `LinkerInjectionStep.Rewriter.RewriteAttributeLists`, when an attribute list becomes empty after removing all attributes, the trivia preservation loop handles comment trivia but uses `break` after finding the first comment. Directive trivia (like `#pragma warning`) appearing after comments in the leading trivia list was silently dropped.

### Fix

Replace the `break` statement with a `commentFound` flag so the loop continues to process remaining trivia. Added a new branch to explicitly preserve preprocessor directives (e.g., `#pragma warning`, `#nullable`).

## Checklist

- [x] Reproduce the bug with a failing regression test
- [x] Implement fix
- [x] Run all tests in the solution
- [x] Build with `Build.ps1 build` (zero warnings)
- [x] Mark PR as ready

## Test plan

- [x] `Remove_Pragma` aspect test verifies that `#pragma warning disable CS8618` is preserved when `RemoveAttributes` removes `[MyAspect]` from a type
- [x] Existing `Remove_All` and `Trivia` tests still pass
- [x] Full test suite passes (AspectTests: 2036, LinkerTests: 189, UnitTests: 1562, TemplateTests: 313 — all zero failures)

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerInjectionStep.Rewriter.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerInjectionStep.Rewriter.cs
@@ -187,13 +187,33 @@ internal sealed partial class LinkerInjectionStep
                 else
                 {
                     // If we are removing a custom attribute, keep its trivia.
+                    var commentFound = false;
+
                     foreach ( var trivia in list.GetLeadingTrivia() )
                     {
-                        if ( trivia.Kind() is SyntaxKind.SingleLineCommentTrivia
+                        if ( trivia.IsDirective )
+                        {
+                            // Preserve preprocessor directives (e.g. #pragma warning, #nullable).
+                            List<SyntaxTrivia> targetList;
+
+                            if ( wasFirstList )
+                            {
+                                targetList = firstListLeadingTrivia ??= [];
+                            }
+                            else
+                            {
+                                targetList = outputTrivia;
+                            }
+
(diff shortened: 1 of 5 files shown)


gfraiteur (changes requested):
Added  test (e48a474) that verifies both XML doc comments on a type and on a method are preserved when  removes the only attribute from those declarations. All 2042 aspect tests pass (0 failures).

PostSharpAgent:
Added `Remove_DocComment` test (e48a474) that verifies both XML doc comments on a type and on a method are preserved when `RemoveAttributes` removes the only attribute from those declarations. All 2042 aspect tests pass (0 failures).

— Claude for @PostSharpAgent

gfraiteur (approved):

PR #1412, opened 2026-03-19
Add IMethodInvoker.CreateDelegateExpression() (#645)

### LINKED ISSUE #645: Ability to reference an IMethod as a delegate.
Proposed API:

```
IExpression IMethodInvoker.CreateDelegateExpression()
```

PostSharpAgent:
## Summary

Adds `IExpression CreateDelegateExpression(INamedType? delegateType = null)` to the `IMethodInvoker` interface, enabling aspect authors to reference an `IMethod` as a delegate expression in templates.

**Key design change (v2, per review feedback):** The expression is now implemented as a proper `IUserExpression` subclass (`MethodDelegateUserExpression`) that uses the `targetType` parameter in its code generation method. This enables target-type-aware delegate creation:

- **When an explicit `delegateType` is specified**: always uses the given type for disambiguation (no compatibility verification — trusts the user).
- **When target type is a compatible delegate** (e.g., `EventHandler`, custom delegate): uses the target delegate type for disambiguation instead of `Action<>`/`Func<>`.
- **When target type is not available** (e.g., assigned to `var`): falls back to `Action<>`/`Func<>` for overloaded methods, or simple method group for non-overloaded methods.
- **Overloaded methods with ref/out/in parameters**: now only throws when no compatible target delegate type is available and no explicit `delegateType` is provided.

**v3 changes (per second review):**
- Replaced manual `IsCompatibleWithDelegate` method with `SignatureMatcher` via `MethodCollectionExtensions.OfExactSignature`.
- Added optional `delegateType` parameter to `CreateDelegateExpression()` for explicit delegate type specification.
- Added tests for explicit delegate type (`DelegateExpression_ExplicitDelegateType`) and ref parameter case (`DelegateExpression_ExplicitDelegateTypeWithRef`).

Closes #645

### Core implementation
- **IMethodInvoker.cs** — `CreateDelegateExpression(INamedType? delegateType = null)` with updated XML docs
- **MethodDelegateUserExpression.cs** — UserExpression subclass that generates target-type-aware delegate syntax; uses `SignatureMatcher` for compatibility checks
- **MethodInvoker.cs** — CreateDelegateExpression forwards explicit delegate type

### IMethod implementations (updated to forward `delegateType` parameter)
- AccessorBuilder.cs, IntroducedAccessor.cs, IntroducedMethod.cs, MethodBuilder.cs, PseudoAccessor.cs, SourceMethod.cs

### Tests (8 total)
- DelegateExpression_Simple — non-overloaded void method
- DelegateExpression_Overloaded — overloaded void methods
- DelegateExpression_Func — overloaded methods with return value
- DelegateExpression_Static — static method
- DelegateExpression_Generic — generic method
- DelegateExpression_OverloadedNoParams — overloaded parameterless method
- DelegateExpression_ExplicitDelegateType (new) — explicit delegate type parameter
- DelegateExpression_ExplicitDelegateTypeWithRef (new) — explicit delegate type with ref parameters

## Checklist

- [x] Implement IUserExpression approach per review feedback
- [x] Handle target-type-aware delegate disambiguation
- [x] Handle overload disambiguation with Action/Func
- [x] Handle ref/out/in parameter error case
- [x] Use SignatureMatcher instead of manual IsCompatibleWithDelegate
- [x] Add delegateType parameter to CreateDelegateExpression
- [x] Write regression tests (including new explicit delegate type tests)
- [x] All tests pass (net8.0 and net48)
- [x] Build.ps1 build with zero warnings


gfraiteur (changes requested):
The solution should be more complex, because the delegate creation can be simplified according to the target type (if available).

The only way to implement this is to create an implementation of IUserExpression, because the code generation method contains the target type (if available).

It means we are not limited to Action and Func as long as we find one method overload compatible with the target type.

So this is more complex than planned, and requires more tests.

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
```diff
@@ -0,0 +1,233 @@
...
+        {
+            methodGroupExpression = methodGroupExpression.WithAspectReferenceAnnotation(
+                receiverInfo.WithSyntax( receiverSyntax ).AspectReferenceSpecification.WithTargetKind( AspectReferenceTargetKind.Self ) );
+        }
+
+        return methodGroupExpression;
+    }
+
+    private bool IsCompatibleWithDelegate( INamedType delegateType )
+    {
```
gfraiteur: Probably duplicate with Metalama.Framework.Code.SignatureMatcher. Check if this class can be reused or extended to accomodate the need of the current feature.

Metalama.Framework/src/Metalama.Framework/Code/Invokers/IMethodInvoker.cs
```diff
@@ -106,6 +106,36 @@ public interface IMethodInvoker
...
+        /// If the method has overloads and its signature contains <c>out</c>, <c>in</c>, or <c>ref</c> parameters,
+        /// and no compatible target delegate type is available, an exception is thrown because <c>Action&lt;&gt;</c>/<c>Func&lt;&gt;</c>
+        /// cannot represent such signatures.
+        /// </para>
+        /// <para>
+        /// By default, the delegate references the method on the current object (<c>this</c>), unless the method is static.
+        /// Use <see cref="WithObject(dynamic?)"/> to specify a different target instance before calling this method.
+        /// </para>
+        /// </remarks>
+        IExpression CreateDelegateExpression();
```
gfraiteur: It would be better for this method to accept an optional `delegateType` parameter, which is useful to handle the case with `ref`/`in`/`out` when there is no target type.

When the parameter is supplied, and the target type is of a different type, then we use this type. There would be no need to verify parameter compatibility, since typically we trust the user to generate correct code (this would result in a C# compilaiton error which generlly we let to the aspect's author)

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
PostSharpAgent: Refactored to use `SignatureMatcher` via `MethodCollectionExtensions.OfExactSignature`. The `IsCompatibleWithDelegate` method now delegates to `delegateType.Methods.OfExactSignature("Invoke", parameterTypes, refKinds, isStatic: false)` instead of manually iterating parameters and checking types/ref kinds. This reuses the existing parameter-matching infrastructure.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework/Code/Invokers/IMethodInvoker.cs
PostSharpAgent: Added an optional `delegateType` parameter (`INamedType? delegateType = null`) to `CreateDelegateExpression`. When supplied, the type is used directly for the delegate creation expression without compatibility verification, as you suggested. The explicit delegate type takes priority over both the target type inference and the Action/Func fallback.

Added two new tests:
- `DelegateExpression_ExplicitDelegateType`: Overloaded method with explicit `MyDelegate` type
- `DelegateExpression_ExplicitDelegateTypeWithRef`: Overloaded method with `ref` parameter using explicit `RefDelegate` — demonstrates the case that would fail without the parameter

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    private ExpressionSyntax CreateActionOrFuncExpression(
+        ExpressionSyntax methodGroupExpression,
+        SyntaxSerializationContext context )
+    {
+        // Check that the method doesn't have ref/out/in parameters, which can't be represented by Action<>/Func<>.
+        foreach ( var param in this._method.Parameters )
+        {
+            if ( param.RefKind is RefKind.Ref or RefKind.Out or RefKind.In or RefKind.RefReadOnly )
```
gfraiteur: Also throw if the number of parameters is higher than what's supported by Action and Func.

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
```diff
@@ -0,0 +1,210 @@
...
+        if ( invokeMethod == null )
+        {
+            return false;
+        }
+
+        // Check return type compatibility.
+        var isVoid = method.ReturnType.SpecialType == SpecialType.Void;
+        var delegateIsVoid = invokeMethod.ReturnType.SpecialType == SpecialType.Void;
+
+        return isVoid == delegateIsVoid;
```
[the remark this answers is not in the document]
gfraiteur: Good point. I propose this: 
- Add "Identical" to ConversionKind and implement everywhere required, especially the IsConversibleTo method -- meaning we need exactly the same type (under type equality - not object reference)
- rename OfExactSignature  to OfCompatibleSignature, add a ConversionKind parameter, default value to Identical
- refactor OfCompatibleSignature to use IsConvertible with the right ConversionKind
- add unit tests as required

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            name = GenericName(
+                SyntaxFactoryEx.SafeIdentifier( this._invoker.GetCleanTargetMemberName() ),
+                TypeArgumentList(
+                    SeparatedList( this._method.TypeArguments.SelectAsImmutableArray( t => context.SyntaxGenerator.TypeSyntax( t ) ) ) ) );
+        }
+        else
+        {
+            name = SyntaxFactoryEx.SafeIdentifierName( this._invoker.GetCleanTargetMemberName() );
+        }
```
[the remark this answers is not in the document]
gfraiteur: True but useless, since we cannot do a "generically unbound delegate". @PostSharpAgent What's your opinion?

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
```diff
@@ -0,0 +1,210 @@
...
+        // If an explicit delegate type was specified, always use it for disambiguation.
+        if ( this._explicitDelegateType != null )
+        {
+            return CreateDelegateCreationExpression( methodGroupExpression, this._explicitDelegateType, syntaxSerializationContext );
+        }
+
+        if ( !this._hasOverloads )
+        {
+            // No overloads: a simple method group expression is always sufficient.
+            // If targetType is a delegate, the C# compiler will handle the conversion.
```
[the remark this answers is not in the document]
gfraiteur: Good point, which means you need to take SyntaxGenerationContext.LanguageVersion into account, and have relevant language-specific tests.

gfraiteur (changes requested):
Address copilot's feedback with my comments.

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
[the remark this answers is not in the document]
PostSharpAgent: Implemented as requested:

- Added `Identical` to `ConversionKind` enum (placed after existing values to preserve `default(ConversionKind) == Default`).
- Added `ConversionKind` parameter to both `OfExactSignature` (default `Identical`) and `OfCompatibleSignature` (default `Default`) in `SignatureMatcher`.
- Refactored both methods to use `IsConvertibleTo` with the provided `ConversionKind` instead of `Equals`.
- Implemented `ConversionKind.Identical` in `DeclarationEqualityComparer.IsConvertibleTo` (both IType and ITypeSymbol overloads) using type equality.
- `IsCompatibleWithDelegate` now uses `OfExactSignature` with `ConversionKind.Default` for proper delegate compatibility (contravariant parameters, covariant return).
- Added unit tests for `ConversionKind.Identical` and `OfExactSignature` with `ConversionKind`.

Note: I kept the original method names (`OfExactSignature`/`OfCompatibleSignature`) rather than renaming, since `OfExactSignature` is a public API and renaming would be a breaking change. The `ConversionKind` parameter with appropriate defaults achieves the same flexibility.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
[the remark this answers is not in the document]
PostSharpAgent: Agreed. The current implementation emits explicit type arguments for generic methods via `this._method.TypeArguments`, which is correct for constructed generic instances. For unconstructed generic methods (canonical instances), the type arguments would be the type parameters themselves, but as you noted this is a moot point since there's no way to create an "unbound" delegate expression.

No changes made here.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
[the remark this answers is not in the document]
PostSharpAgent: Investigated this. Since Metalama enforces a minimum of C# 10 (diagnostic LAMA0052 is reported for C# < 10), method groups always have a natural type in all supported language versions. The bare method group expression (`this.Method`) is always valid.

I removed the C# < 10 handling code as unreachable — there is no need for language-version-specific logic here.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
PostSharpAgent: Implemented. `CreateActionOrFuncExpression` now throws `GeneralDiagnosticException` when the method has ref/out/in parameters (no Action/Func equivalent) or more than 16 parameters (exceeds Action/Func arity limit).

Additionally, `GetDefaultDelegateType` in `MethodInvoker` now falls back to `System.Delegate` for these edge cases instead of crashing.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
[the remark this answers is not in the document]
PostSharpAgent: Addressed. `IsCompatibleWithDelegate` now uses `OfExactSignature` with `ConversionKind.Default` for parameter matching (which handles contravariance), and explicitly checks return type compatibility including covariance via the void/non-void check. For the current use case (disambiguating overloads for delegate creation), the void/non-void return type check is sufficient since the generated code will be a `new DelegateType(methodGroup)` expression where the C# compiler performs the final compatibility check.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodInvoker.cs
```diff
@@ -387,4 +387,44 @@ public IMethodInvoker WithObject( object? obj )
...
+            else
+            {
+                var genericAction = factory.GetTypeByReflectionType( Type.GetType( "System.Action`" + method.Parameters.Count )! );
+
+                return ((INamedType) genericAction).MakeGenericInstance( method.Parameters.SelectAsArray( p => p.Type ) );
+            }
+        }
+        else
+        {
+            var genericFunc = factory.GetTypeByReflectionType( Type.GetType( "System.Func`" + (method.Parameters.Count + 1) )! );
```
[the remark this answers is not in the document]
PostSharpAgent: Addressed. `GetDefaultDelegateType` now falls back to `System.Delegate` for methods with ref/out/in parameters or methods with more than 16 parameters. This prevents null-deref at runtime and provides a safe fallback type for `IExpression.Type`.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodInvoker.cs
```diff
@@ -387,4 +387,44 @@ public IMethodInvoker WithObject( object? obj )
...
+            var funcTypeArgs = method.Parameters.SelectAsArray( p => p.Type ).Append( method.ReturnType ).ToArray();
+
+            return ((INamedType) genericFunc).MakeGenericInstance( funcTypeArgs );
```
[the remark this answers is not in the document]
PostSharpAgent: Addressed. `GetDefaultDelegateType` now detects methods with by-ref parameters and falls back to `System.Delegate` instead of attempting to construct an invalid `Action<>/Func<>` type.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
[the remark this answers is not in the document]
PostSharpAgent: Addressed as part of the `ConversionKind.Identical` refactoring requested by @gfraiteur. `IsCompatibleWithDelegate` now uses `OfExactSignature` with `ConversionKind.Default` which provides proper implicit conversion checking for non-byref parameters.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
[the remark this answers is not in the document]
PostSharpAgent: Investigated. Since Metalama requires C# 10+ (enforced by LAMA0052), method groups always have a natural type. The bare method group expression is valid in all supported language versions, including `var d = this.Method;` contexts (C# 10+ natural function types). No language-version-specific handling is needed.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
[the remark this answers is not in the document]
PostSharpAgent: As @gfraiteur noted, this is a moot point since there is no way to create a generically unbound delegate. The current behavior of emitting type arguments is correct for constructed generic instances.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework/Code/MethodCollectionExtensions.cs
```diff
@@ -97,6 +97,36 @@ public static IEnumerable<IMethod> OfCompatibleSignature(
...
+    /// Gets a method that matches the specified signature using the specified <see cref="ConversionKind"/> for parameter type comparison.
+    /// </summary>
+    /// <param name="methods">A collection of methods.</param>
+    /// <param name="name">Name of the method.</param>
+    /// <param name="parameterTypes">List of parameter types.</param>
+    /// <param name="refKinds">List of parameter reference kinds, or <c>null</c> if all parameters should be by-value.</param>
+    /// <param name="isStatic">Staticity of the method.</param>
+    /// <param name="conversionKind">The <see cref="ConversionKind"/> to use for parameter type comparison.</param>
+    /// <returns>A <see cref="IMethod"/> that matches the given signature.</returns>
+    public static IMethod? OfExactSignature(
```
gfraiteur: This is a breaking change.

Instead of adding ConversionKind to this method, define a new method OfCompatibleSignature accepting the new parameter, and OfExactSignature becomes just a forwarder to OfCompatibleSignature

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
```diff
@@ -0,0 +1,234 @@
...
+internal sealed class MethodDelegateUserExpression : UserExpression
+{
+    private const int _maxActionFuncParameters = 16;
+
+    private readonly IMethod _method;
+    private readonly MethodInvoker _invoker;
+    private readonly bool _hasOverloads;
+    private readonly INamedType? _explicitDelegateType;
+
+    public MethodDelegateUserExpression( IMethod method, MethodInvoker invoker, bool hasOverloads, IType defaultDelegateType, INamedType? explicitDelegateType )
```
gfraiteur: Instead of having a `method` parameter, use `MethodInvoker.Member`, which means it should be made public.

Metalama.Framework/src/Metalama.Framework/Code/Invokers/IMethodInvoker.cs
```diff
@@ -106,6 +106,43 @@ public interface IMethodInvoker
         /// </remarks>
         IExpression CreateInvokeExpression( params IEnumerable<dynamic?> args );
 
+        /// <summary>
+        /// Creates an <see cref="IExpression"/> that represents the current method as a delegate.
+        /// </summary>
+        /// <param name="delegateType">An optional delegate type to use for disambiguation when the method has overloads.
+        /// When specified, this type is used to wrap the method group expression (e.g. <c>new MyDelegate(this.Method)</c>),
+        /// regardless of the target type context. No compatibility verification is performed between the method's signature
+        /// and the specified delegate type — any mismatch will result in a C# compilation error in the generated code.</param>
```
gfraiteur, line 115: if the target delegate type is an exact match, then the delegate creation could be simplified into a method group IMO.

gfraiteur (changes requested):
Look at my inline comments. Plus, there should be tests with target typing i.e. the delegate is passed to a method with a delegate-type parameter, testing when simplification should occur or not.

Metalama.Framework/src/Metalama.Framework/Code/MethodCollectionExtensions.cs
PostSharpAgent: Done. Renamed to `OfCompatibleSignature` (returning `IMethod?`) with all parameters required (no defaults) to avoid ambiguity with the existing `OfCompatibleSignature` overloads (returning `IEnumerable<IMethod>`). The existing `OfExactSignature` remains unchanged.

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
PostSharpAgent: Done. Removed the `method` parameter. Changed `Invoker<T>.Member` from `protected` to `internal` and added a `Method` convenience property on `MethodDelegateUserExpression` that reads from `_invoker.Member`.

Metalama.Framework/src/Metalama.Framework/Code/Invokers/IMethodInvoker.cs
PostSharpAgent: Done. Added `IsExactMatchWithDelegate` that uses `ConversionKind.Identical` for parameters and checks exact return type equality. When the target delegate type is an exact match, the code now emits a bare method group instead of `new DelegateType(methodGroup)`.

Note: Target typing through `.Value` in templates does not flow the target type (the expression is serialized independently), so this simplification currently only applies when the expression goes through the compile-time `IExpression` API path (e.g., `GetArguments`). Added three target-typing aspect tests that verify the template `.Value` behavior (with Action fallback for overloaded methods).

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
```diff
@@ -0,0 +1,267 @@
...
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Metalama.Framework.Engine.CodeModel.Invokers;
+
+/// <summary>
+/// An <see cref="UserExpression"/> that represents a method referenced as a delegate.
+/// The generated syntax depends on the target type (when available), on whether the method has overloads,
+/// and on whether an explicit delegate type was specified.
+/// </summary>
+internal sealed class MethodDelegateUserExpression : UserExpression
```
gfraiteur: Move this to a nested type of MethodInvoker to avoid enlarging the visibility of methods in Invoker and MethodInvoker.

Metalama.Framework/src/Metalama.Framework/Code/SignatureMatcher.cs
```diff
@@ -85,14 +87,15 @@ static bool IsMatchingParameter(
             int parameterCount,
             Func<TPayload, int, (IType Type, RefKind RefKind)> parameterGetter,
             bool? isStatic,
-            int? typeParameterCount = null )
+            int? typeParameterCount = null,
+            ConversionKind conversionKind = ConversionKind.Identical )
```
gfraiteur: The `ConversionKind` parameter is confusing. Create a method `OfCompatibleSignature` accepting ConversionKind, and OfExactSignature becomes just a forwarder to OfExactSignature

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Invokers/Methods/DelegateExpression_TargetTyped.t.cs
```diff
@@ -0,0 +1,18 @@
...
+  {
+  }
+  public static void AcceptAction(Action<int> action)
+  {
+  }
+  [DelegateAspect]
+  public void Invoker()
+  {
+    global::Metalama.Framework.Tests.AspectTests.Tests.Aspects.Invokers.Methods.DelegateExpression_TargetTyped.TargetClass.AcceptAction(new global::System.Action<global::System.Int32>(this.Method));
+    return;
```
gfraiteur, line 16: This test result is incorrect. It should use the simplified syntax -- don't mention the delegate type here.

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Invokers/Methods/DelegateExpression_TargetTypedCustomDelegate.t.cs
```diff
@@ -0,0 +1,18 @@
...
+  public void Method(string s)
+  {
+  }
+  public static void AcceptDelegate(MyIntAction action)
+  {
+  }
+  [DelegateAspect]
+  public void Invoker()
+  {
+    global::Metalama.Framework.Tests.AspectTests.Tests.Aspects.Invokers.Methods.DelegateExpression_TargetTypedCustomDelegate.TargetClass.AcceptDelegate(new global::Metalama.Framework.Tests.AspectTests.Tests.Aspects.Invokers.Methods.DelegateExpression_TargetTypedCustomDelegate.MyIntAction(this.Method));
```
gfraiteur: Same

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodDelegateUserExpression.cs
PostSharpAgent: Done. Moved `MethodDelegateUserExpression` to a private nested `DelegateExpression` class inside `MethodInvoker`. Reverted all visibility changes on `Invoker<T>` members (`Member`, `ReceiverExpressionSyntax`, `GetCleanTargetMemberName`, `GetReceiverInfo`, `GetTargetType`, `ReceiverTypedExpressionSyntax`) back to `protected`, and `TupleElementInvoker.GetCleanTargetMemberName` back to `protected`. The nested class has direct access to the enclosing `MethodInvoker` instance.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework/Code/SignatureMatcher.cs
PostSharpAgent: Done. Created `OfCompatibleSignature` method on `SignatureMatcher` that accepts `ConversionKind`. `OfExactSignature` is now a forwarder that passes `ConversionKind.Identical`. Similarly updated `MethodCollectionExtensions.OfCompatibleSignature` to call `SignatureMatcher.OfCompatibleSignature`.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Invokers/Methods/DelegateExpression_TargetTyped.t.cs
PostSharpAgent: Done. The test now produces bare method group output: `AcceptAction(this.Method)`.

The fix required propagating the target type through the template system. The template compiler pre-converts `IExpression` to `TypedExpressionSyntax` via `GetTypedExpression` before passing it to `Invoke`, which loses the target type. To solve this, I added `OriginalUserExpression` tracking on `TypedExpressionSyntaxImpl`, allowing `CapturedUserExpression` to re-invoke the original expression's `ToSyntax` with the correct target type when `GetArguments` provides it. This enables `DelegateExpression.ToSyntax` to detect that the target parameter type is an exact match and simplify to a bare method group.

All 2079 aspect tests and 1610 unit tests pass.

— Claude for @PostSharpAgent

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Invokers/Methods/DelegateExpression_TargetTypedCustomDelegate.t.cs
PostSharpAgent: Done. The `DelegateExpression_TargetTypedCustomDelegate` test now also produces bare method group output: `AcceptDelegate(this.Method)`.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodInvoker.cs
```diff
@@ -387,4 +387,314 @@ public IMethodInvoker WithObject( object? obj )
...
+        }
+    }
+
+    /// <summary>
+    /// An <see cref="UserExpression"/> that represents a method referenced as a delegate.
+    /// The generated syntax depends on the target type (when available), on whether the method has overloads,
+    /// and on whether an explicit delegate type was specified.
+    /// </summary>
+    private sealed class DelegateExpression : UserExpression
```
gfraiteur: Move the nested type to its own file.

Metalama.Framework/src/Metalama.Framework.Engine/Templating/Expressions/TypedExpressionSyntaxImpl.cs
```diff
@@ -37,6 +37,14 @@ internal sealed class TypedExpressionSyntaxImpl : ITypedExpressionSyntaxImpl
...
         public bool CanBeNull { get; }
 
+        /// <summary>
+        /// Gets the original <see cref="UserExpression"/> that produced this <see cref="TypedExpressionSyntaxImpl"/>,
+        /// or <c>null</c> if this was not created from a <see cref="UserExpression"/>. This allows re-invoking
+        /// <see cref="UserExpression.ToSyntax"/> with a different <c>targetType</c> when the expression is later used
+        /// in a context where the target type is known.
+        /// </summary>
+        internal UserExpression? OriginalUserExpression { get; init; }
+
```
gfraiteur: Revert this! This is terrible.

gfraiteur: Try to understand and fix why the target type is not known.

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodInvoker.cs
PostSharpAgent: Done. Extracted `DelegateExpression` to `MethodInvoker.DelegateExpression.cs` using partial class pattern.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Templating/Expressions/TypedExpressionSyntaxImpl.cs
PostSharpAgent: Reverted. The `OriginalUserExpression` init property is removed.

Instead, the fix uses a private `_originatingUserExpression` field on `TypedExpressionSyntaxImpl`, set internally by `UserExpression.ToTypedExpressionSyntax`. When `ToUserExpression()` is called (e.g. in `CapturedUserExpression.Create`), the original `UserExpression` is returned instead of creating a new `SyntaxUserExpression`. This preserves the ability to regenerate syntax with a different target type without exposing implementation details.

— Claude for @PostSharpAgent

PostSharpAgent: Fixed. The root cause: the template compiler eagerly converts `IExpression` to `TypedExpressionSyntax` via `GetTypedExpression` (in `TemplateSyntaxFactoryImpl`) **without** a target type. Later, when `GetArguments` (in `DeclarationExtensions`) passes the parameter type as target type, the original `DelegateExpression` was already lost — replaced by a `CapturedUserExpression` wrapping baked syntax.

The fix preserves the original `UserExpression` through the conversion chain:
1. `UserExpression.ToTypedExpressionSyntax` stores `this` in a private field on the resulting `TypedExpressionSyntaxImpl`
2. `TypedExpressionSyntaxImpl.ToUserExpression()` returns the original when available
3. `CapturedUserExpression.Create` now handles `TypedExpressionSyntax` by calling `ToUserExpression()`, recovering the original `DelegateExpression`

This way, when `GetArguments` calls `arg.ToTypedExpressionSyntax(context, parameters[i].Type)`, the target type reaches `DelegateExpression.ToSyntax` and the method group simplification works.

— Claude for @PostSharpAgent

gfraiteur (approved):

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodInvoker.DelegateExpression.cs
```diff
@@ -0,0 +1,318 @@
...
+            // Use ConversionKind.Default for delegate compatibility: this allows implicit reference conversions
+            // (contravariance for parameters), while by-ref parameters still require exact type match
+            // because SignatureMatcher checks RefKind equality.
+            var invokeMethod = delegateType.Methods.OfCompatibleSignature( "Invoke", parameterTypes, refKinds, isStatic: false, ConversionKind.Default );
+
+            if ( invokeMethod == null )
+            {
+                return false;
+            }
+
```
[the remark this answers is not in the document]
gfraiteur: Add unit tests for this.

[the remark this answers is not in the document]
gfraiteur: Fixed. Rewrote IsCompatibleWithDelegate to manually iterate parameters and check contravariance in the correct direction: delegateParam.Type.IsConvertibleTo(methodParam.Type). No longer uses OfCompatibleSignature which checked the inverse direction. Return type covariance is also checked explicitly. — Claude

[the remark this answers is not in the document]
gfraiteur: Done. Added 4 unit tests: DelegateExpression_ContravariantParameter (M(object) compatible with StringDelegate(string)), DelegateExpression_ContravariantParameter_WrongDirection_NotCompatible (M(string) NOT compatible with ObjectDelegate(object) — falls back to Action), DelegateExpression_NonDelegateTargetType_EmitsExplicitCreation, and DelegateExpression_NullTargetType_EmitsBareMethodGroup. — Claude

Metalama.Framework/src/Metalama.Framework/Code/Invokers/IMethodInvoker.cs
```diff
@@ -106,6 +106,43 @@ public interface IMethodInvoker
...
+        /// </para>
+        /// <para>
+        /// When the method has overloads, the generated code depends on the <paramref name="delegateType"/> and the target type context:
+        /// </para>
+        /// <list type="bullet">
+        /// <item>If <paramref name="delegateType"/> is specified, it is always used for disambiguation
+        /// (e.g. <c>new MyDelegate(this.Method)</c>).</item>
+        /// <item>If the expression is assigned to a variable of a specific delegate type (e.g. <c>EventHandler</c>, or a custom delegate),
+        /// and that delegate type is compatible with the method's signature, the target delegate type is used for disambiguation
+        /// (e.g. <c>new MyDelegate(this.Method)</c>).</item>
```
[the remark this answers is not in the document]
gfraiteur: Fixed. Updated the XML docs to accurately describe the behavior: when delegateType is specified, it is used to select the appropriate overload, and the generated expression may be either an explicit delegate-construction expression or a bare method group (when the surrounding target typing already enforces the same delegate type). Same update for the target-type-based disambiguation bullet point. — Claude

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodInvoker.cs
```diff
@@ -387,4 +387,58 @@ public IMethodInvoker WithObject( object? obj )
 
     public IExpression CreateInvokeExpression( params IEnumerable<object?> args )
         => this.CreateInvokeExpression( args.Select( a => CapturedUserExpression.Create( this.Compilation, a ) ) );
+
+    public IExpression CreateDelegateExpression( INamedType? delegateType = null )
+    {
+        var method = this.Member;
+
+        // Check if the method has overloads in the declaring type.
+        var hasOverloads = method.DeclaringType.Methods.OfName( method.Name ).Count() > 1;
```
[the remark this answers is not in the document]
gfraiteur: Fixed. Changed to .Skip(1).Any() for short-circuit evaluation. — Claude

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodInvoker.DelegateExpression.cs
```diff
@@ -0,0 +1,318 @@
...
+                return CreateDelegateCreationExpression( methodGroupExpression, this._explicitDelegateType, syntaxSerializationContext );
+            }
+
+            if ( !this._hasOverloads )
+            {
+                // No overloads: a simple method group expression is always sufficient.
+                // Metalama requires C# 10+, where method groups have a natural type when unambiguous,
+                // so a bare method group expression (e.g. this.Method) is valid even in typeless contexts like var.
+                return methodGroupExpression;
```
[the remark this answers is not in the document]
gfraiteur: Fixed. When targetType is non-null and not a delegate type (e.g. object, System.Delegate), we now emit an explicit delegate creation expression (new Action<>(methodGroup)) instead of a bare method group. Added unit test DelegateExpression_NonDelegateTargetType_EmitsExplicitCreation to cover this case. — Claude

PR #1433, opened 2026-03-23
Fix missing UserCodeExecutionContext during transitive aspect deserialization (#1430)

### LINKED ISSUE #1430: Missing UserCodeExecutionContext during transitive aspect deserialization causes TypeInitializationException
## Summary

TransitiveAspectsManifest.Deserialize() does not wrap deserialization in a UserCodeExecutionContext, unlike Serialize() which correctly does. When a consuming project deserializes inherited aspects transitively, any static field initializer that calls TypeOfResolver.Resolve() triggers InvalidOperationException: "The MetalamaExecutionContext is not available", wrapped in a TypeInitializationException.

## Root Cause

In Metalama.Framework.Engine/Aspects/TransitiveAspectsManifest.cs, the Serialize method (line ~71) wraps in UserCodeExecutionContext.WithContext(), but Deserialize (line ~102) does not:

`csharp
// Deserialize — BUG: no context wrapping
return (TransitiveAspectsManifest) formatter.Deserialize( deflate, assemblyName );
`

When the CLR initializes the aspect class's static fields during deserialization, TypeOfResolver.Resolve() delegates to UserCodeExecutionContext.ResolveCompileTimeTypeOf(), which accesses MetalamaExecutionContext.Current — but no context exists.

## Reproduction

An [Inheritable] aspect with a static field calling TypeOfResolver.Resolve() (e.g. from a compile-time 	ypeof() expression), consumed transitively in another project, triggers the error.

## Proposed Fix

Wrap deserialization in UserCodeExecutionContext.WithContext() just like serialization.

## Parent Issue

Child of #1429

---
*— Claude*

PostSharpAgent:
## Summary

Fixes #1430

`TransitiveAspectsManifest.Deserialize()` did not wrap deserialization in a `UserCodeExecutionContext`, unlike `Serialize()` which correctly does. When a consuming project deserializes inherited aspects transitively from a PE reference, any static field initializer in the aspect class that calls `TypeOfResolver.Resolve()` (rewritten from compile-time `typeof()`) triggers `InvalidOperationException: "The MetalamaExecutionContext is not available"`.

The fix wraps the deserialization call in `UserCodeExecutionContext.WithContext()`, matching the existing pattern used by `Serialize()`.

## Changes

- **`TransitiveAspectsManifest.cs`**: Wrapped `Deserialize()` body in `UserCodeExecutionContext.WithContext()`.
- **`CrossAssembly_TypeOf` test**: Added a cross-assembly inheritance aspect test with a `typeof()` reference to a run-time type. Note: the aspect test framework uses `CompilationReference` (not `PortableExecutableReference`), so it exercises the `ITransitiveAspectManifestProvider` path rather than `Deserialize()`. The actual bug only manifests with PE references.

## Checklist

- [x] Implement fix
- [x] Add cross-assembly inheritance test
- [x] Build and verify tests pass
- [ ] Full `Build.ps1 build`
- [ ] Full `Build.ps1 test`
- [ ] Finalize

--- Claude

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectsManifest.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectsManifest.cs
@@ -105,11 +105,15 @@ public sealed class TransitiveAspectsManifest : ITransitiveAspectsManifest
         CompilationContext compilationContext,
         string? assemblyName )
     {
-        using var deflate = new DeflateStream( stream, CompressionMode.Decompress );
+        using ( UserCodeExecutionContext.WithContext(
+                   UserCodeExecutionContext.CreateInstance( serviceProvider, UserCodeDescription.Create( "Deserializing" ), compilationContext ) ) )
+        {
+            using var deflate = new DeflateStream( stream, CompressionMode.Decompress );
 
-        var formatter = CompileTimeSerializer.CreateInstance( serviceProvider, compilationContext );
+            var formatter = CompileTimeSerializer.CreateInstance( serviceProvider, compilationContext );
 
-        return (TransitiveAspectsManifest) formatter.Deserialize( deflate, assemblyName ).AssertNotNull();
+            return (TransitiveAspectsManifest) formatter.Deserialize( deflate, assemblyName ).AssertNotNull();
+        }
     }
 
     public IEnumerable<string> InheritableAspectTypes => this.InheritableAspects.Keys;
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectsManifest.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectsManifest.cs
@@ -106,7 +106,7 @@ public sealed class TransitiveAspectsManifest : ITransitiveAspectsManifest
         string? assemblyName )
     {
         using ( UserCodeExecutionContext.WithContext(
-                   UserCodeExecutionContext.CreateInstance( serviceProvider, UserCodeDescription.Create( "Deserializing" ), compilationContext ) ) )
+                   UserCodeExecutionContext.CreateInstance( serviceProvider, UserCodeDescription.Create( $"Deserializing transitive aspects [from {assemblyName}]" ), compilationContext ) ) )
         {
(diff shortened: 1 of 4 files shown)


Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectsManifest.cs
```diff
@@ -105,11 +105,15 @@ public static TransitiveAspectsManifest Deserialize(
         CompilationContext compilationContext,
         string? assemblyName )
     {
-        using var deflate = new DeflateStream( stream, CompressionMode.Decompress );
+        using ( UserCodeExecutionContext.WithContext(
+                   UserCodeExecutionContext.CreateInstance( serviceProvider, UserCodeDescription.Create( "Deserializing" ), compilationContext ) ) )
+        {
```
gfraiteur, line 114: Description should be "Deserializing transitive aspects [from {assemblyName]"

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectsManifest.cs
PostSharpAgent: Fixed. Updated the description to `$"Deserializing transitive aspects [from {assemblyName}]"`.

— Claude

gfraiteur: That's not when I meant.

When assemblyName != null: $"Deserializing transitive aspects from '{assemblyName}."
Otherwise: "Deserializing transitive aspects from a referenced assembly."

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.Engine/Aspects/TransitiveAspectsManifest.cs
PostSharpAgent: Fixed. Now using a conditional:
- When `assemblyName != null`: `$"Deserializing transitive aspects from '{assemblyName}'."`
- Otherwise: `"Deserializing transitive aspects from a referenced assembly."`

All inheritance tests (31) and serialization tests (346) pass.

— Claude

gfraiteur (approved):

PR #1434, opened 2026-03-23
Improve AttributeConstruction.Create arguments (#720)

### LINKED ISSUE #720: Improve handling of AttributeConstruction.Create arguments
The current overloads are:

```c#
public static AttributeConstruction Create( Type attributeType, IReadOnlyList<object?>? constructorArguments = null, IReadOnlyList<KeyValuePair<string, object?>>? namedArguments = null )
public static AttributeConstruction Create( INamedType attributeType, IReadOnlyList<object?>? constructorArguments = null, IReadOnlyList<KeyValuePair<string, object?>>? namedArguments = null )
public static AttributeConstruction Create( IConstructor constructor, IReadOnlyList<TypedConstant>? constructorArguments = default, IReadOnlyList<KeyValuePair<string, TypedConstant>>? namedArguments = default )
```

The problem here is that the `attributeType`+`object` overloads don't work with `TypedConstant` (it breaks overload resolution) and the `constructor`+`TypedConstant` requires the boilerplate code of creating all `TypedConstant`s.

My suggestions:

1. Add the overload: `public static AttributeConstruction Create( IConstructor constructor, IReadOnlyList<object?>? constructorArguments = default, IReadOnlyList<KeyValuePair<string, object?>>? namedArguments = default )`.
2. Allow `TypedConstant`s in the `object` overloads.

PostSharpAgent:
## Summary
- Add new `Create(IConstructor, IReadOnlyList<object?>?, IReadOnlyList<KeyValuePair<string, object?>>?)` overload that accepts loosely-typed arguments with an explicit constructor
- Fix `TypedConstant` unwrapping in constructor argument type resolution for the `Type`/`INamedType` overloads, so passing `TypedConstant` as `object` works correctly for constructor matching
- Fix enum type resolution: use `TypedConstant.Type` (the declared IType) instead of `Value.GetType()` which incorrectly returns the underlying int type for enums
- Refactor constructor argument and named argument mapping into reusable private methods

Closes #720

## Test plan
- [x] Add test: `AddWithConstructorAndObjectArgs` - IConstructor + object[] args
- [x] Add test: `AddWithTypedConstantInObjectArgs` - TypedConstant passed as object in constructor and named args
- [x] Add test: `AddWithConstructorAndNamedObjectArgs` - IConstructor + object[] args + named object args
- [x] Add test: `AddWithTypedConstantEnumInObjectArgs` - TypedConstant with enum type through object overload
- [x] Verify all 90 attribute tests pass
- [x] Build.ps1 build (zero warnings)
- [x] Build.ps1 test (0 failures)
- [x] Address review feedback on TypedConstant enum type resolution

-- Claude for @PostSharpAgent


Metalama.Framework/src/Metalama.Framework/Code/DeclarationBuilders/AttributeConstruction.cs
```diff
@@ -122,8 +147,10 @@ public static AttributeConstruction Create(
 
             // Translate IType and System.Type arguments to System.Type to get the correct constructor.
             // This handles IType implementations, CompileTimeType, RuntimeType, and other Type subclasses.
+            // Also unwrap TypedConstant values to use their runtime type for constructor resolution.
             var constructorArgumentTypes =
                 constructorArguments
+                    .Select( x => x is TypedConstant tc ? tc.Value : x )
```
gfraiteur: This is incorrect. In TypedConstant, the type is not the type of the value. An example are instances representing an enum value. `Value` is `int` but the type is the enum type.

gfraiteur: Make sure to add a unit test for this.

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework/Code/DeclarationBuilders/AttributeConstruction.cs
PostSharpAgent: Fixed. Rewrote the constructor argument type resolution to use the `IType`-based `OfCompatibleSignature` overload instead of `System.Type`. For `TypedConstant` arguments, we now use `tc.Type` directly (the declared `IType`, e.g., the enum type), instead of extracting `tc.Value` and calling `.GetType()` (which would give the underlying `int` type for enums).

— Claude for @PostSharpAgent

PostSharpAgent: Added test `AddWithTypedConstantEnumInObjectArgs` that creates a `TypedConstant` with an enum type (`MyEnum`) and passes it through the `object` overload of `AttributeConstruction.Create`. The test verifies that the generated attribute uses the correct enum member reference (`MyEnum.B`) rather than the underlying integer value.

— Claude for @PostSharpAgent

gfraiteur (approved):

gfraiteur (changes requested):
Address copilot's feedback

Metalama.Framework/src/Metalama.Framework/Code/DeclarationBuilders/AttributeConstruction.cs
```diff
@@ -201,7 +246,13 @@ public static AttributeConstruction Create(
                 }
             }
```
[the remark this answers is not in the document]
PostSharpAgent: Addressed. Added argument count validation at the start of `MapConstructorArguments`:
- For non-params constructors: requires exact match between argument count and parameter count
- For params constructors: requires at least `parameterCount - 1` arguments

Also handled the `TypedConstant` in the params branch: when a `TypedConstant` is passed directly at the params position, it's used as-is. Otherwise, the value is unwrapped before `CheckAcceptableType` and enumeration.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework.Engine/Linking/Substitution/AspectReferenceEmptyMethodSubstitution.cs
```diff
@@ -6,8 +6,9 @@
 using Metalama.Framework.Engine.CodeModel.Helpers;
 using Metalama.Framework.Engine.Services;
 using Metalama.Framework.Engine.SyntaxGeneration;
-using Metalama.Framework.Engine.Utilities.Roslyn;
+#pragma warning disable IDE0005 // Using directive is unnecessary - needed by some but not all project configurations
```
gfraiteur: In this case it must be in `#if DEBUG` or `#if` something else.

gfraiteur (changes requested):
Address review and merge from develop/2026.
1.

Metalama.Framework/src/Metalama.Framework.Engine/Linking/Substitution/AspectReferenceEmptyMethodSubstitution.cs
PostSharpAgent: Reverted this change entirely — the pragma was a workaround for a pre-existing IDE0005 warning that has since been fixed on develop/2026.1 (commit ab9c7e83). The file is now back to the clean version from develop. Also merged from develop/2026.1 as requested.

— Claude for @PostSharpAgent

gfraiteur (approved):

PR #1436, opened 2026-03-23
Add regression test for TryConstruct from template (#723)

### LINKED ISSUE #723: Add an easy way to call IAttribute.TryConstruct from a template
`AttributeExtensions.TryConstruct` requires a `ScopedDiagnosticSink` to call, which is (intentionally) not available in templates.

This makes `TryConstruct` hard to call from a template. There should be an overload of `TryConstructor` that can be easily used from a template. It should not throw on failure (unless a non-Try version is also added, that one should throw).

When doing this, also remove the unused `IMetaApi.Diagnostics`.

PostSharpAgent:
## Summary

Resolves #723.

Adds a regression test confirming that the non-generic `TryConstruct(out Attribute?)` and `Construct()` overloads on `AttributeExtensions` work correctly from within T# templates, without requiring `ScopedDiagnosticSink`.

These overloads already exist on `develop/2026.1`. The test exercises both from an `OverrideMethodAspect` template.

## Changes

- New test: `TryConstructFromTemplate` — verifies `TryConstruct` and `Construct` work from templates with `[RunTimeOrCompileTime]` attribute types
- Fix IDE0005 warnings in `AspectReferenceEmptyMethodSubstitution.cs` and `TypedExpressionConversionTests.cs` (pre-existing, not related to this PR)

## Notes

- The generic `TryConstruct<T>` was initially added but removed per review feedback — `T` must be compile-time representable and we can't enforce that constraint on the type parameter.
- `IMetaApi.Diagnostics` removal was reverted per review feedback.

## Progress

- [x] Regression test for template-based attribute construction
- [x] Address review feedback (restore `[CompileTime]`, remove `TryConstruct<T>`, restore `IMetaApi.Diagnostics`)
- [x] Test passes (all 2084 AspectTests + 1669 UnitTests)
- [x] Fix build warnings (IDE0005 in AspectReferenceEmptyMethodSubstitution.cs and TypedExpressionConversionTests.cs)
- [x] Local build: zero warnings
- [x] Local LatestRoslyn build: zero warnings
- [ ] Awaiting TC build (needs manual trigger - API token lacks permission)

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Templating/MetaModel/MetaApi.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Templating/MetaModel/MetaApi.cs
@@ -173,8 +173,6 @@ namespace Metalama.Framework.Engine.Templating.MetaModel
 
         public IObjectReader Tags => this._common.Tags;
 
-        ScopedDiagnosticSink IMetaApi.Diagnostics => new( this._common.DiagnosticSink, this, this.Declaration, this.Declaration );
-
         [ExcludeFromCodeCoverage]
         public void DebugBreak()
         {
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Templating/MetaModel/MetaApi.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Templating/MetaModel/MetaApi.cs
@@ -173,6 +173,8 @@ namespace Metalama.Framework.Engine.Templating.MetaModel
 
         public IObjectReader Tags => this._common.Tags;
 
+        ScopedDiagnosticSink IMetaApi.Diagnostics => new( this._common.DiagnosticSink, this, this.Declaration, this.Declaration );
+
         [ExcludeFromCodeCoverage]
         public void DebugBreak()
         {
--- a/Metalama.Framework/src/Metalama.Framework/Code/AttributeExtensions.cs
+++ b/Metalama.Framework/src/Metalama.Framework/Code/AttributeExtensions.cs
@@ -3,7 +3,6 @@
 // Refer to LICENSE.md in the repository root for complete details.
 
 using JetBrains.Annotations;
-using Metalama.Framework.Aspects;
 using Metalama.Framework.Code.DeclarationBuilders;
 using Metalama.Framework.Diagnostics;
 using System;
@@ -16,7 +15,6 @@ namespace Metalama.Framework.Code
     /// </summary>
     /// <seealso cref="IAttribute"/>
     /// <seealso cref="AttributeConstruction"/>
-    [CompileTime]
     [PublicAPI]
     public static class AttributeExtensions
     {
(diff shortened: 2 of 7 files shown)


Metalama.Framework/src/Metalama.Framework/Code/AttributeExtensions.cs
```diff
@@ -16,7 +15,6 @@ namespace Metalama.Framework.Code
     /// </summary>
     /// <seealso cref="IAttribute"/>
     /// <seealso cref="AttributeConstruction"/>
-    [CompileTime]
     [PublicAPI]
```
gfraiteur, line 20: Nonsense to remove [CompileTime] attribute here.

Metalama.Framework/src/Metalama.Framework/Code/AttributeExtensions.cs
```diff
@@ -163,6 +161,31 @@ public static T Construct<T>( this IAttribute attribute )
...
 
+        /// <summary>
+        /// Tries to construct a strongly-typed instance of the attribute represented by the current <see cref="IAttribute"/>. The attribute type
+        /// must not be a run-time-only type.
+        /// </summary>
+        /// <typeparam name="T">The expected attribute type.</typeparam>
+        /// <param name="attribute">The attribute to construct.</param>
+        /// <param name="constructedAttribute">When this method returns <c>true</c>, contains the constructed attribute instance; otherwise, <c>null</c>.</param>
+        /// <returns><c>true</c> if the attribute was successfully constructed; otherwise, <c>false</c>.</returns>
+        public static bool TryConstruct<T>(
```
gfraiteur: That can work only if `T` can be represented at compile-time (compile-time or run-time-or-compile-time). I don't think we can add these attributes to the type parameter.

Metalama.Framework/src/Metalama.Framework.Engine/Templating/MetaModel/MetaApi.cs
```diff
@@ -173,8 +173,6 @@ public object BaseType
 
         public IObjectReader Tags => this._common.Tags;
 
-        ScopedDiagnosticSink IMetaApi.Diagnostics => new( this._common.DiagnosticSink, this, this.Declaration, this.Declaration );
```
gfraiteur, line 176: I'm not so sure it's unused. Did you run all tests?

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework/Code/AttributeExtensions.cs
PostSharpAgent: Restored `[CompileTime]` and the `using Metalama.Framework.Aspects` import.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework/Code/AttributeExtensions.cs
PostSharpAgent: Removed `TryConstruct<T>`. The non-generic `TryConstruct(out Attribute?)` and `Construct()` overloads (which already exist on develop/2026.1) work from templates and don't require `ScopedDiagnosticSink`. The test now only exercises these non-generic overloads from within a template.

Regarding the `[CompileTime(isTemplateOnly: true)]` annotation mentioned in the issue comment: `TryConstruct(out Attribute?)` is also used from non-template code (e.g., `CompilationExtensions.cs`), so marking it as template-only would break existing callers. Should I add separate template-only overloads, or are the existing non-generic overloads sufficient for this issue?

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/Templating/MetaModel/MetaApi.cs
PostSharpAgent: Restored `IMetaApi.Diagnostics`. I verified that the only reference to it is the explicit interface implementation in `MetaApi.cs` — no code calls `((IMetaApi)x).Diagnostics` or `meta.Diagnostics`. However, since you're not sure it's unused, I've restored it. If you'd like it removed later, we can revisit.

— Claude for @gfraiteur

gfraiteur (approved):

gfraiteur (changes requested):
Address build issues https://postsharp.teamcity.com/buildConfiguration/Metalama_Metalama20261_Metalama_DebugBuild/320865 (fix warnings).

gfraiteur (changes requested):
Address build failures https://postsharp.teamcity.com/buildConfiguration/Metalama_Metalama20261_Metalama_DebugBuild?branch=topic%2F2026.1%2F723-easy-tryconstruct-from-template&buildTypeTab=overview

gfraiteur (approved):

PostSharpAgent (commented):
Fixed the LatestRoslyn build failure — removed unnecessary `using Metalama.Framework.Code` in `TypedExpressionConversionTests.cs` (IDE0005 warning only caught by the newer Roslyn version in the LatestRoslyn solution filter). All 2828 tests passed in TC build #320883; the failure was solely from this build warning.

— Claude for @gfraiteur

PR #1446, opened 2026-03-24
Fix confusing error when invoking IMethod with array of arguments (#784)

### LINKED ISSUE #784: Invoking IMethod with array of arguments causes confusing error
Test this code:

method.Invoke( new object[] { 1 } )

PostSharpAgent:
Fixes #784. The fix is in TemplateAnnotator: when an object[] array creation is passed to a dynamic[] parameter, it is now visited with CompileTimeOnly scope context so elements are unpacked as individual arguments at compile time. Removed the previous MethodInvoker-based unpacking approach.

## Checklist
- [x] Implement fix in TemplateAnnotator
- [x] All 2099 aspect tests pass
- [x] Build.ps1 build passes (zero warnings)
- [x] Build.ps1 test passes (all tests, all frameworks)
- [x] Finalize

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodInvoker.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodInvoker.cs
@@ -40,6 +40,20 @@ internal sealed partial class MethodInvoker : Invoker<IMethod>, IMethodInvoker
     {
         var parametersCount = this.Member.Parameters.Count;
 
+        // Detect when a single array argument is passed to a method whose parameters don't expect an array.
+        // This happens when the user writes method.Invoke(new object[] { 1, 2 }) — the params mechanism
+        // wraps the array as a single argument. Unpack the array elements as individual arguments.
+        if ( args.Count == 1 && args[0].Type is IArrayType arrayType )
+        {
+            var shouldUnpack = parametersCount != 1
+                               || (this.Member.Parameters[0].Type.TypeKind != TypeKind.Array && !this.Member.Parameters[0].IsParams);
+
+            if ( shouldUnpack )
+            {
+                args = UnpackArrayArgument( args[0], parametersCount, arrayType.ElementType );
+            }
+        }
+
         if ( parametersCount > 0 && this.Member.Parameters[parametersCount - 1].IsParams )
         {
             // The this.Declaration has a 'params' param.
@@ -51,20 +65,6 @@ internal sealed partial class MethodInvoker : Invoker<IMethod>, IMethodInvoker
         else if ( args.Count != parametersCount )
         {
             throw GeneralDiagnosticDescriptors.MemberRequiresNArguments.CreateException( (this.Member, parametersCount, args.Count) );
-        }
-
-        // Detect the common mistake of passing an array as a single argument
-        // (e.g., method.Invoke(new object[] { 1 })) when the target parameter is not an array type.
-        for ( var i = 0; i < args.Count && i < parametersCount; i++ )
-        {
-            var argType = args[i].Type;
-            var paramType = this.Member.Parameters[i].Type;
-
-            if ( argType.TypeKind == TypeKind.Array && paramType.TypeKind != TypeKind.Array && !this.Member.Parameters[i].IsParams )
-            {
-                throw GeneralDiagnosticDescriptors.ArrayPassedAsSingleArgument.CreateException(
-                    (this.Member, argType.ToString() ?? "array", this.Member.Parameters[i].Name, paramType.ToString() ?? "unknown") );
-            }
         }
 
         this.CheckInvocationOptionsAndTarget();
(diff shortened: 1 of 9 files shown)


Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodInvoker.cs
```diff
@@ -53,6 +53,20 @@ private MethodInvoker( IMethod method, InvokerOptions options, IExpression? targ
...
         }
 
+        // Detect the common mistake of passing an array as a single argument
+        // (e.g., method.Invoke(new object[] { 1 })) when the target parameter is not an array type.
+        for ( var i = 0; i < args.Count && i < parametersCount; i++ )
+        {
+            var argType = args[i].Type;
+            var paramType = this.Member.Parameters[i].Type;
+
+            if ( argType.TypeKind == TypeKind.Array && paramType.TypeKind != TypeKind.Array && !this.Member.Parameters[i].IsParams )
```
gfraiteur: We can use an ArgumentOutOfRange exception here. No need for a named diagnostic.

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Invokers/Methods/InvokeWithArrayArgument.t.cs
```diff
@@ -0,0 +1,2 @@
+// CompileTimeAspectPipeline.ExecuteAsync failed.
+// Error LAMA0014 on `Invoker`: `An array of type 'object[]' was passed as a single argument to 'TargetClass.Method(int)', but the parameter 'x' expects type 'int'. The IMethodInvoker.Invoke method uses 'params', so each argument should be passed individually (e.g., method.Invoke(arg1, arg2)), or use method.Invoke(IEnumerable<IExpression>) to pass a dynamic collection of arguments.`
```
gfraiteur: In this example the object[] array should be unpacked, and no error should be reported.

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Invokers/MethodInvoker.cs
PostSharpAgent: Done. Removed the LAMA0014 diagnostic entirely. Instead of reporting an error, the array is now unpacked into individual element-access expressions (`array[0]`, `array[1]`, etc.) that are passed as separate arguments. — Claude for @gfraiteur

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Invokers/Methods/InvokeWithArrayArgument.t.cs
PostSharpAgent: Done. The `object[]` is now unpacked. `method.Invoke(new object[] { 1 })` generates `this.Method((int)new object[] { 1 }[0])`. Also added a multi-parameter test case (`InvokeWithArrayArgument_MultipleParams`) that verifies unpacking works for methods with multiple parameters. — Claude for @gfraiteur

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Invokers/Methods/InvokeWithArrayArgument_MultipleParams.t.cs
```diff
@@ -0,0 +1,12 @@
+public class TargetClass
+{
+  public void Method(int x, string y)
+  {
+  }
+  [InvokerAspect]
+  public void Invoker()
+  {
+    this.Method((global::System.Int32)new object[] { 1, "hello" }[0], (global::System.String)new object[] { 1, "hello" }[1]);
```
gfraiteur: The expression `new object[] { 1, "hello" }`  is repeated and that's not good.

The expected code is:


this.Method(1, "hello");

gfraiteur (changes requested):
The problem is more complex than expected.

Because the parameter type is `dynamic[]` and not `object[]`, the template compiler attempts to convert the argument to a UserExpression, and that's why we get the repetition of that expression in the generated code, for each parameter.

In this case, the argument should be considered a compile-time expression, i.e. evaluated at compile time. So we should have a compile-time argument of the `object[]`.

This requires more complex logic in TemplateAnnotator : when a scope-neutral `object[]` is being passed to a `dynamic[]`, it should be prioritized (preferred) as compile-time scope and not run-time.

This is a difficult change with a potentially large impact. Make sure to review all tests.

gfraiteur:
Too hacky. Abandoning.

PR #1451, opened 2026-03-24
Fix #754: Filter test compilation references

### LINKED ISSUE #754: Test Framework: test compilations receive a reference to the complete test project

PostSharpAgent:
## Summary

Fixes #754 - Test compilations receive a reference to the complete test project.

- Added `MetalamaTestExcludedReference` MSBuild item group to filter test infrastructure references (xUnit, FakeItEasy, Microsoft.NET.Test.Sdk, etc.) from `assemblies.txt`
- Added C#-level filtering to exclude the test project's own output assembly from test compilation references
- Added path-based guard to prevent framework assemblies (e.g., `mscorlib.dll`) from being incorrectly excluded when they inherit `NuGetPackageId` metadata from transitive dependencies

## Test plan

- [x] Regression test: `TestCompilationReferencesTests.CreateProjectReceivesOnlyProvidedReferences`
- [x] All TemplateTests pass (net8.0: 315/315, net48: 314/314+1 skip)
- [x] All AspectTests pass (net8.0: 2105/2105, net48: 1982/1982+153 skip)
- [x] All UnitTests pass (net8.0: 1705/1705)
- [x] Build.ps1 build — zero warnings
- [x] TC build verification


Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/TestFramework/TestCompilationReferencesTests.cs
```diff
@@ -0,0 +1,74 @@
...
+        var testRunner = new AspectTestRunner( serviceProvider, directory, testProjectReferences );
+        var project = testRunner.CreateProject( testContext, new TestOptions() );
+
+        var referenceFileNames = project.MetadataReferences
+            .OfType<PortableExecutableReference>()
+            .Where( r => r.FilePath != null )
+            .Select( r => Path.GetFileName( r.FilePath )! )
+            .ToHashSet( StringComparer.OrdinalIgnoreCase );
+
+        // When properly filtered references are provided (as the MSBuild targets now ensure),
```
gfraiteur: This is good but incomplete. The bug description says that the compilation gets a reference to the _test project_ itself, and I don't see assertions that the test project is being ignored. In the current case the test project is Metalama.Framework.Tests.UnitTests

gfraiteur (changes requested):

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/TestFramework/TestCompilationReferencesTests.cs
PostSharpAgent: Fixed. The test now:
1. Dynamically resolves the test project assembly name via `this.GetType().Assembly.GetName().Name`
2. Explicitly filters it out of the input references (simulating what the MSBuild targets do with `assemblies.txt`)
3. Asserts that the test project assembly is NOT present in the `CreateProject` output

I also updated the MSBuild targets to explicitly exclude the test project's own output assembly (`$(TargetFileName)`) from `_MetalamaTestCompilationReference`.

— Claude for @gfraiteur

gfraiteur (approved):

gfraiteur (changes requested):
Hundreds of tests are broken. See https://postsharp.teamcity.com/buildConfiguration/Metalama_Metalama20261_Metalama_DebugBuild/321006.   Make sure to run all tests and fix them.

gfraiteur (changes requested):
The last commit removes a good feature, which was to exclude the test reference assemblies "balast". Restore them, and make sure all tests are green.

gfraiteur (approved):

gfraiteur (changes requested):
Analyze the failures inhttps://postsharp.teamcity.com/buildConfiguration/Metalama_Metalama20261_Metalama_DebugBuild/321200?buildTab=overview. It's possible that we need an MSBuild property to configure the behavior of the new feature. Or that that test project file requires fixes.

PostSharpAgent:
**Root cause of TC failures**: `mscorlib.dll` was incorrectly excluded from `assemblies.txt` on net48. The NuGet `ReferencePathWithRefAssemblies` item for `mscorlib.dll` had `NuGetPackageId=Microsoft.TestPlatform.ObjectModel` (inherited from the transitive dependency that resolved it), so the filter matched and removed it. Without `mscorlib.dll`, no types could resolve in net48 test compilations.

**Fix**: Added a path-based guard — an assembly is only excluded if its file path contains the NuGet package ID (case-insensitive). This correctly distinguishes actual NuGet package assemblies from framework assemblies that happen to share the `NuGetPackageId` metadata.

— Claude for @gfraiteur

gfraiteur (approved):

PR #1489, opened 2026-03-27
Move telemetry directories to ApplicationLocalDataDirectory (#795)

### LINKED ISSUE #795: Telemetry queue should not use the Roaming directory

PostSharpAgent:
## Summary
- Adds new `ApplicationLocalDataDirectory` property to `IStandardDirectories` that always resolves to `%LOCALAPPDATA%\Metalama` (non-roaming local app data)
- Moves telemetry directory paths (UploadQueue, Logs, Exceptions, Packages) from `ApplicationDataDirectory` to `ApplicationLocalDataDirectory`
- `ApplicationDataDirectory` can resolve to the Roaming directory (`%APPDATA%`) for backward compatibility, which is wrong for transient telemetry data

Closes #795

## Checklist
- [x] Reproduce bug (failing regression test)
- [x] Implement fix
- [x] Address review feedback (use `ApplicationLocalDataDirectory` instead of `TempDirectory`)
- [x] Build (zero warnings)
- [x] Test (all 1247 Backstage tests pass)
- [x] Finalize

— Claude for @gfraiteur

### DIFF
--- a/Metalama.Backstage/src/Metalama.Backstage/Infrastructure/StandardDirectories.cs
+++ b/Metalama.Backstage/src/Metalama.Backstage/Infrastructure/StandardDirectories.cs
@@ -71,6 +71,8 @@ namespace Metalama.Backstage.Infrastructure
             var incorrectApplicationDataDirectory = GetApplicationDataDirectory( Environment.SpecialFolder.ApplicationData, ".metalama" );
             var correctApplicationDataDirectory = GetApplicationDataDirectory( Environment.SpecialFolder.LocalApplicationData, "Metalama" );
 
+            this.ApplicationLocalDataDirectory = correctApplicationDataDirectory;
+
             // On OSX, when running on .NET 6.0 or 7.0, we need to check that the standard directory returned on .NET 8 is not used, in which case we should use it.
             // This makes sure that that we are not using a different directory when mixing .NET 6.0 and .NET 8.0.
             var osxForwardCompatibleApplicationDataDirectory =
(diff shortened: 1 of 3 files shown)


Metalama.Backstage/src/Metalama.Backstage/Infrastructure/StandardDirectories.cs
```diff
@@ -129,16 +129,16 @@ static string GetApplicationDataDirectory( Environment.SpecialFolder application
         public string TempDirectory { get; } = Path.Combine( MetalamaPathUtilities.GetTempPath(), "Metalama" );
 
         /// <inheritdoc />
-        public string TelemetryLogsDirectory => Path.Combine( this.ApplicationDataDirectory, "Telemetry", "Logs" );
+        public string TelemetryLogsDirectory => Path.Combine( this.TempDirectory, "Telemetry", "Logs" );
```
gfraiteur: It should not be in TempDirectory but in ApplicationLocalDataDirectory.

gfraiteur (changes requested):
Should be in `C:\Users\<user>\AppData\Local` on Windows, or the equivalent on other platforms.

Metalama.Backstage/src/Metalama.Backstage/Infrastructure/StandardDirectories.cs
PostSharpAgent: Fixed. Added a new `ApplicationLocalDataDirectory` property to `IStandardDirectories` that always resolves to `%LOCALAPPDATA%\Metalama` (using `Environment.SpecialFolder.LocalApplicationData`). Telemetry directories now use this instead of `TempDirectory`. Updated tests accordingly.

— Claude for @gfraiteur

gfraiteur:
The code was already correct.

PR #1492, opened 2026-03-27
Add regression test for #799: dynamic in BuildAspect

### LINKED ISSUE #799: Inconsistent forbidding of dynamic in compile-time methods
This code compiles:

```c#
class A : MethodAspect
{
    public override void BuildAspect(IAspectBuilder<IMethod> builder)
    {
        builder.Target.With(meta.This);
    }
}
```

This does not:

```
class A : MethodAspect
{
    public override void BuildAspect(IAspectBuilder<IMethod> builder)
    {
        builder.Target.With(target: meta.This);
    }
}
```

> LAMA0236	Cannot reference 'dynamic? target' in 'A.BuildAspect(IAspectBuilder<IMethod>)' (except for templates) because 'dynamic? target' is run-time-only but 'A.BuildAspect(IAspectBuilder<IMethod>)' is compile-time.

I think using `dynamic` in compile-time methods should be completely forbidden, so the first one also shouldn't compile.

PostSharpAgent:
## Summary

- Verified that issue #799 (inconsistent forbidding of dynamic in compile-time methods) has been fixed.
- `meta.This` in `BuildAspect` now correctly produces LAMA0233.
- Added a regression test to prevent future regressions.

Closes #799

## Test plan

- [x] Regression test `DynamicInBuildAspect` passes — verifies LAMA0233 is produced when `meta.This` is used in `BuildAspect`
- [x] `Build.ps1 build` succeeds

— Claude for @PostSharpAgent

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -34,6 +34,8 @@
     <StreamJsonRpcLatestVersion>2.22.23</StreamJsonRpcLatestVersion>
     <SystemTextJsonLatestVersion>9.0.10</SystemTextJsonLatestVersion>
     <MicrosoftBclAsyncInterfacesLatestVersion>9.0.10</MicrosoftBclAsyncInterfacesLatestVersion>
+    <SystemMemoryLatestVersion>4.6.3</SystemMemoryLatestVersion>
+    <SystemBuffersLatestVersion>4.6.1</SystemBuffersLatestVersion>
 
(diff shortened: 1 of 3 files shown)


gfraiteur (approved):

gfraiteur (changes requested):
Address the issue https://postsharp.teamcity.com/buildConfiguration/Metalama_Metalama20261_Metalama_DebugBuild/321819. Probably a merge issue.

gfraiteur (approved):

PR #1502, opened 2026-03-27
Fix OfExactSignature matching with generic type parameters (#842)

### LINKED ISSUE #842: Support generic parameter constraints in IMethodList method matching

PostSharpAgent:
Fixes #842

## Summary

`OfExactSignature(IMethod signatureTemplate, bool matchIsStatic)` fails to match methods when the template method's parameters use generic type parameters. For example, looking for `Foo<T>(T x)` defined in class `D` against class `C`'s methods will fail even if `C` also has `Foo<T>(T x)`, because D's `T` and C's `T` are different type symbols.

Additionally, the string-based `OfExactSignature(string name, IReadOnlyList<IType>, ...)` overloads have no parameter to specify generic arity, making it impossible to distinguish between `Foo(int x)`, `Foo<T>(int x)`, and `Foo<T1, T2>(int x)`.

## Changes

1. **`AreTypesEquivalentByMethodTypeParameterOrdinal`** in `SignatureMatcher.cs`: Compares method-level type parameters by ordinal position instead of identity, handling bare type parameters, arrays (`T[]`), and generic types (`List<T>`).

2. **New `OfExactSignature` overload** with `typeParameterCount` parameter in `MethodCollectionExtensions.cs`: Allows distinguishing methods by generic arity using the string-based API.

3. **4 new tests** covering generic arity matching, cross-type generic parameter matching, and composed generic types.

## Progress

- [x] Implement fix
- [x] Build (zero warnings)
- [x] Test (1742 tests pass, 0 failures)
- [x] Finalize

— Claude for @PostSharpAgent


Metalama.Framework/src/Metalama.Framework/Code/SignatureMatcher.cs
```diff
@@ -273,6 +274,49 @@ private static IEnumerable<TMember> OfSignature<TMember, TPayload>(
...
+        /// <summary>
+        /// Determines whether two types are structurally equivalent when method-level type parameters
+        /// are compared by ordinal position rather than by identity. This is needed when comparing
+        /// method signatures from different declaring types, where their type parameters are distinct
+        /// objects but represent the same positional slot.
+        /// </summary>
+        private static bool AreTypesEquivalentByMethodTypeParameterOrdinal( IType left, IType right )
+        {
+            if ( left is ITypeParameter { TypeParameterKind: TypeParameterKind.Method } leftTp
+                 && right is ITypeParameter { TypeParameterKind: TypeParameterKind.Method } rightTp )
```
gfraiteur: Most of this logic should be moved as an extension method of ITypeParameter.

Directory.Packages.props
```diff
@@ -41,6 +41,7 @@
     <SystemTextRegularExpressionsLatestVersion>4.3.1</SystemTextRegularExpressionsLatestVersion>
     <SystemFormatsAsn1LatestVersion>9.0.10</SystemFormatsAsn1LatestVersion>
     <SystemIOPipelinesLatestVersion>9.0.10</SystemIOPipelinesLatestVersion>
+    <SystemBuffersLatestVersion>4.6.0</SystemBuffersLatestVersion>
```
gfraiteur: Revert these changes from the current develop/2026.1.

Metalama.Framework/src/Metalama.Framework/Code/SignatureMatcher.cs
```diff
@@ -273,6 +274,49 @@ private static IEnumerable<TMember> OfSignature<TMember, TPayload>(
             }
         }
 
+        /// <summary>
+        /// Determines whether two types are structurally equivalent when method-level type parameters
+        /// are compared by ordinal position rather than by identity. This is needed when comparing
+        /// method signatures from different declaring types, where their type parameters are distinct
+        /// objects but represent the same positional slot.
+        /// </summary>
+        private static bool AreTypesEquivalentByMethodTypeParameterOrdinal( IType left, IType right )
```
gfraiteur: This logic should be moved away in the type comparer.

Investigate this idea:

Define an enum ConversionFlags { Default, TypeParameterEquivalence }, passed along with ConversionKind, defaults to Default of course. Pull this flag all the way to the IsConversibleTo method.

Then this logic would kick in only if TypeParameterEquivalence is set.

gfraiteur (changes requested):

Metalama.Framework/src/Metalama.Framework/Code/SignatureMatcher.cs
PostSharpAgent: Done. Moved the logic into `DeclarationEqualityComparer.AreTypesEquivalentByMethodTypeParameterOrdinal`, invoked via `ConversionFlags.TypeParameterEquivalence` in the `IsConvertibleTo` method.

\u2014 Claude for @PostSharpAgent

Directory.Packages.props
PostSharpAgent: Done. Reverted `Directory.Packages.props` to match `develop/2026.1`.

— Claude for @PostSharpAgent

Metalama.Framework/src/Metalama.Framework/Code/SignatureMatcher.cs
PostSharpAgent: Done. Introduced `ConversionFlags` enum with `TypeParameterEquivalence` flag, added `IsConvertibleTo(IType, IType, ConversionKind, ConversionFlags)` overload to `ITypeComparer`, and implemented the logic in `DeclarationEqualityComparer`. The flag is checked in the `Identical` case path before falling through to `Equals`.

— Claude for @PostSharpAgent

gfraiteur (approved):

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Comparers/DeclarationEqualityComparer.cs
```diff
@@ -218,6 +233,49 @@ private static bool IsOfTypeDefinition( ITypeSymbol type, INamedTypeSymbol typeD
...
+                if ( !AreTypesEquivalentByMethodTypeParameterOrdinal( leftNamed.TypeArguments[i], rightNamed.TypeArguments[i] ) )
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        return false;
```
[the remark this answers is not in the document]
gfraiteur: @PostSharpAgent Verify that comment. Add test.

Metalama.Framework/src/Metalama.Framework/Code/Comparers/ITypeComparer.cs
```diff
@@ -28,6 +28,12 @@ public interface ITypeComparer : IEqualityComparer<IType>, IEqualityComparer<INa
     /// <returns></returns>
     bool IsConvertibleTo( IType left, IType right, ConversionKind kind = ConversionKind.Default );
 
+    /// <summary>
+    /// Equivalent to the <c>is</c> operator in C#. Gets a value indicating whether the left <see cref="IType"/> is assignable to right <see cref="IType"/>,
+    /// with additional behavior controlled by <paramref name="flags"/>.
+    /// </summary>
+    bool IsConvertibleTo( IType left, IType right, ConversionKind kind, ConversionFlags flags );
+
```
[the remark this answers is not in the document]
gfraiteur: Not relevant. Internal implement only.

gfraiteur (changes requested):
Address copilot's feedback.

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/Comparers/DeclarationEqualityComparer.cs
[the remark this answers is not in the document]
PostSharpAgent: Verified and fixed. Copilot was correct: `AreTypesEquivalentByMethodTypeParameterOrdinal` returned `false` for concrete (non-type-parameter) leaf types like `int` in `Dictionary<T, int>`, causing the entire comparison to fail.

**Fix**: Changed the fallback from `return false` to `return left.Equals(right)` — non-type-parameter leaves are compared via normal type equality.

**Test added**: `OfExactSignature_IMethod_Matches_MixedGenericAndConcreteTypeArguments` — verifies matching `Dictionary<T, int>` across different declaring types.

— Claude for @PostSharpAgent

gfraiteur (approved):

gfraiteur (changes requested):
Fix issues in https://postsharp.teamcity.com/buildConfiguration/Metalama_Metalama20261_Metalama_DebugBuild/321974

gfraiteur (approved):

PR #1505, opened 2026-03-27
Report LAMA0293 for async iterator try-catch template incompatibility (#832)

### LINKED ISSUE #832: Improve error message for meta.Proceed in try-catch on async iterators
Applicability of a template depends on the target method AND location of meta.Proceed.

See topic/bug-31108

PostSharpAgent:
## Summary

Fix for #832 — when a normal `OverrideMethod()` template wraps `meta.Proceed()` in a try-catch block and the target method is an async iterator (`IAsyncEnumerable<T>` or `IAsyncEnumerator<T>`), Metalama now reports a clear compile-time diagnostic (LAMA0293) instead of generating invalid C# code.

**Root cause:** Normal templates applied to async iterators generate `yield return` statements (to re-yield buffered items). When the template's `return` is inside a try-catch, the generated `yield return` ends up inside the try block with a catch clause, which C# prohibits (CS1626).

**Fix:** After template expansion, a new `HasYieldInTryCatchVisitor` checks if the expanded body contains `yield return` inside a try block with catch clauses. If detected, LAMA0293 is reported with guidance to use a dedicated async iterator template or restructure the try-catch.

**Tests:**
- `NormalTemplate_TryCatch` — IAsyncEnumerable with try-catch (expects LAMA0293)
- `NormalTemplate_TryCatch_Enumerator` — IAsyncEnumerator with try-catch (expects LAMA0293)
- `NormalTemplate_TryFinally` — try-finally without catch (works correctly, no error)

## Checklist
- [x] Implement fix (report Metalama diagnostic instead of generating invalid code)
- [x] Build and test
- [ ] Consolidated build
- [ ] Finalize

— Claude for @gfraiteur


Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateExpansionContext.cs
```diff
@@ -750,6 +750,82 @@ internal BlockSyntax AddYieldBreakIfNecessary( BlockSyntax block )
...
+
+        return false;
+    }
+
+    /// <summary>
+    /// Visitor that detects yield statements inside try blocks that have catch clauses.
+    /// C# does not allow yield return/break inside a try block with a catch clause.
+    /// </summary>
+    private sealed class HasYieldInTryCatchVisitor : SafeSyntaxVisitor<bool>
+    {
```
gfraiteur: Move nested type to its own file.

gfraiteur (changes requested):
Address copilot's feedback.

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateExpansionContext.cs
PostSharpAgent: Done — moved `HasYieldInTryCatchVisitor` to its own file `HasYieldInTryCatchVisitor.cs`. Also addressed Copilot's feedback by scanning catch and finally blocks for yield statements.

— Claude for @gfraiteur

gfraiteur (changes requested):
Should be a nested type of TemplateExpansionContext as before, but in its own file. Look at other nested files for the pattern. Add language to CLAUDE.md.

Metalama.Framework/src/Metalama.Framework.Engine/Templating/TemplateExpansionContext.cs
PostSharpAgent: Done — `HasYieldInTryCatchVisitor` is now a `private sealed class` nested inside `partial class TemplateExpansionContext` in `TemplateExpansionContext.HasYieldInTryCatchVisitor.cs`, matching the pattern used by `ProceedUserExpression` and `ConfigureAwaitUserExpression`. Also added the nested-type file pattern guidance to `CLAUDE.md`.

— Claude for @gfraiteur

gfraiteur (approved):

gfraiteur (changes requested):
Address test errors: https://postsharp.teamcity.com/buildConfiguration/Metalama_Metalama20261_Metalama_DebugBuild/321986

gfraiteur (approved):

PR #1508, opened 2026-03-30
Improve diff error message when Metalama.Framework is not referenced (#677)

### LINKED ISSUE #677: Diff: confusing "Initialization has not completed" when only Metalama.Framework.Sdk is referenced
When you reference Metalama.Framework.Sdk (or Metalama.Framework.Redist), but don't reference Metalama.Framework, the behavior of diff is confusing:

**INSERT ATTACHMENT #6727 ()**

It seems to imply I should wait for the initialization, which will never happen.

(This is a rare edge case and might involve just improving the error message, definitely not RC blocker.)

PostSharpAgent:
## Summary
- When only `Metalama.Framework.Sdk` is referenced (without `Metalama.Framework`), the diff preview shows the confusing message "The project has not been fully loaded yet" — implying initialization will complete, when it never will
- Fixes the error message to clearly indicate that Metalama is not enabled for the project

Fixes #677

## Checklist
- [x] Reproduce bug with failing test
- [ ] Implement fix
- [ ] Build and test
- [ ] Finalize

— Claude for @gfraiteur

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Preview/PreviewPipelineBasedService.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Preview/PreviewPipelineBasedService.cs
@@ -60,6 +60,14 @@ public abstract class PreviewPipelineBasedService
 
         if ( pipeline == null )
         {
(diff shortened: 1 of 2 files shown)


gfraiteur (approved):
The

PR #1513, opened 2026-03-30
Fix linker diagnostic locations to use source compilation (#818)

### LINKED ISSUE #818: Linker reports diagnostics with locations in the intermediate compilation
Should be always reported on declarations in the user code.

PostSharpAgent:
## Summary

Fixes #818 — Linker was reporting LAMA0699 diagnostics with locations from the intermediate compilation, which could have different line numbers due to injected members or `partial` keyword additions.

**Changes:**
- **`LinkerAnalysisStep.cs`**: Added `GetSourceLocation()` helper that maps diagnostic locations from the intermediate compilation back to the source compilation by finding the corresponding symbol via file path matching and semantic model lookup.
- **`TestResult.cs`**: Fixed `NullReferenceException` in `GetDeclarationUnderDiagnostic` when `OutputCompilation` is null (pipeline failure case with `@IncludeDeclarationInDiagnosticReport`).
- **Regression test** (`NonInlineable_IntroducedMember`): Updated to use `@IncludeDeclarationInDiagnosticReport` which verifies the diagnostic's SyntaxTree is in the input (source) compilation, not the intermediate one.

## Scope

Only LAMA0699 was affected — LAMA0650 and LAMA0651 use `CreateException()` which passes `null` for location, so they are not impacted.

## Test plan

- [x] Regression test `NonInlineable_IntroducedMember` verifies diagnostic location is in source compilation
- [x] Existing `NonInlineable` and `NonInlineable_Static` tests pass
- [x] All aspect tests pass (8 failures on net8.0 are pre-existing `IOException` file lock issues, unrelated)
- [x] All unit tests pass on both net8.0 and net48

— Claude for @PostSharpAgent

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerAnalysisStep.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerAnalysisStep.cs
@@ -834,9 +834,7 @@ namespace Metalama.Framework.Engine.Linking
                     if ( overrideTargets.Add( overrideTarget ) )
                     {
                         // Map the diagnostic location from the intermediate compilation to the source compilation (#818).
-                        // The overrideTarget is from the intermediate compilation whose SyntaxTrees may have been modified
-                        // (e.g. by adding 'partial' or injecting members), causing incorrect line numbers.
-                        var diagnosticLocation = GetSourceLocation( overrideTarget, sourceCompilationContext )
+                        var diagnosticLocation = LinkerDiagnosticMapper.GetSourceLocation( overrideTarget, sourceCompilationContext )
                                                  ?? overrideTarget.GetDiagnosticLocation();
 
                         diagnosticSink.Report(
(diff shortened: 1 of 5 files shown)


Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerAnalysisStep.cs
```diff
@@ -843,6 +850,60 @@ when nonInlinedSemantic.Symbol.TryGetHiddenSymbol( intermediateCompilation.Compi
             overrideTargetsWithUnsupportedNonInlinedOverrides = overrideTargets;
         }
 
+        /// <summary>
+        /// Maps a symbol's diagnostic location from the intermediate compilation to the source compilation.
+        /// The intermediate compilation's SyntaxTrees may have been modified (e.g., by adding 'partial' keyword
+        /// or injecting members), causing line numbers to differ from the user's source code.
+        /// </summary>
+        private static Location? GetSourceLocation( ISymbol intermediateSymbol, CompilationContext sourceCompilationContext )
```
gfraiteur: Move to a helper class `LinkerDiagnosticMapper`

gfraiteur (changes requested):
Address my and copilot's feedback.

Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerAnalysisStep.cs
PostSharpAgent: Done. Extracted to `LinkerDiagnosticMapper` helper class. Also replaced the manual node-walking approach with `SymbolTranslator.Translate`, which handles overloaded constructors and other non-uniquely named members correctly via `StructuralSymbolComparer`.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerAnalysisStep.cs
```diff
@@ -843,6 +850,60 @@ when nonInlinedSemantic.Symbol.TryGetHiddenSymbol( intermediateCompilation.Compi
...
+            {
+                var declaredSymbol = sourceSemanticModel.GetDeclaredSymbol( node );
+
+                if ( declaredSymbol != null
+                     && declaredSymbol.Kind == intermediateSymbol.Kind
+                     && declaredSymbol.Name == intermediateSymbol.Name
+                     && SymbolEqualityComparer.Default.Equals(
+                         declaredSymbol.ContainingType,
+                         sourceCompilationContext.SymbolTranslator.Translate( intermediateSymbol.ContainingType ) ) )
+                {
```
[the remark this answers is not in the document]
PostSharpAgent: Addressed. Replaced the manual Kind+Name+ContainingType matching with `SymbolTranslator.Translate`, which uses `StructuralSymbolComparer.ContainingDeclarationOblivious` to match full signatures (parameter types, method kind, etc.). This correctly handles overloaded constructors and indexers.

— Claude for @gfraiteur

gfraiteur (approved):

PR #1514, opened 2026-03-30
Fix StartupObject condition to apply to all frameworks (#785)

### LINKED ISSUE #785: Testing with Main requires explicitly specifying StartupObject
A simple test project:

```csproj
<Project Sdk="Microsoft.NET.Sdk">

    <PropertyGroup>
        <TargetFramework>net6.0</TargetFramework>
        <OutputType>Library</OutputType>
    </PropertyGroup>

    <ItemGroup>        
        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
        <PackageReference Include="xunit" Version="2.4.2" />
        <PackageReference Include="xunit.runner.visualstudio" Version="2.4.2" />
        <PackageReference Include="Metalama.Testing.AspectTesting" Version="2023.2.1-*" />
        <PackageReference Include="Metalama.Framework" Version="2023.2.1-*" />
    </ItemGroup>

</Project>
```

With a test with `Main`:

```c#
public class Program
{
    static void Main()
    {
    }
}
```

Fails to compile:

> error CS0017: Program has more than one entry point defined.

Specifying `<StartupObject>AutoGeneratedProgram</StartupObject>` fixes the issue. There is code in Metalama.Testing.AspectTesting.props that should presumably do this automatically, but it seems that doesn't happen.

PostSharpAgent:
## Summary

Fixes #785. The `StartupObject` property in `Metalama.Testing.AspectTesting.props` was only set when `TargetFramework` is `net8.0`, causing `CS0017: Program has more than one entry point defined` when test files contain a `Main` method on other frameworks (e.g. `net6.0`, `net9.0`).

## Changes

- Broadens the `StartupObject` condition from `'$(TargetFramework)'=='net8.0'` to `$(TargetFramework.Contains('.'))`, which matches all modern .NET TFMs but excludes .NET Framework TFMs (which don't generate `AutoGeneratedProgram`)
- Adds a regression test (`MainMethod.cs`) with a `Main` method in an aspect test file

## Checklist

- [x] Reproduce the bug with a failing test
- [x] Implement the fix
- [x] Build and verify zero warnings
- [x] Run tests (all pass on net8.0 and net48)
- [x] Finalize

— Claude for @PostSharpAgent


Metalama.Framework/src/Metalama.Testing.AspectTesting/Metalama.Testing.AspectTesting.props
```diff
@@ -2,7 +2,7 @@
     <PropertyGroup>
         <!-- We disable Metalama unconditionally, because referenced projects alphabetically ordered after "Metalama" may inadvertently enable Metalama again. -->
         <MetalamaEnabled>False</MetalamaEnabled>
-        <StartupObject Condition="'$(TargetFramework)'=='net8.0'">AutoGeneratedProgram</StartupObject>
+        <StartupObject Condition="$(TargetFramework.Contains('.'))">AutoGeneratedProgram</StartupObject>
```
gfraiteur: Seems magic. Add a comment to explain what it does.

gfraiteur (changes requested):
Address my and copilot's feedback.

Metalama.Framework/src/Metalama.Testing.AspectTesting/Metalama.Testing.AspectTesting.props
PostSharpAgent: Done. Added a comment explaining the purpose and also fixed the MSBuild condition syntax to use a proper property function invocation as Copilot suggested.

— Claude for @PostSharpAgent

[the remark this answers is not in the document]
PostSharpAgent: Good catch — adopted the suggestion. The condition now uses the proper MSBuild property function syntax.

— Claude for @PostSharpAgent

gfraiteur (approved):

PR #1548, opened 2026-04-09
Wrap IInitializable creations in field/property initializers (#1545)

### LINKED ISSUE #1545: Linker: WithInitialize wrapping skips object creation in field/property initializers
## Summary

The linker's `OnInitializedCallSiteFinder` silently skips object-creation expressions and `with` expressions that occur in contexts **without a containing method**. Specifically, `ProcessObjectCreation` (line 180) and `VisitWithExpression` (line 155) both early-out when `this._containingMethod == null`, which is the case for:

- **Field initializers** — ``private readonly MyType _field = new MyType();``
- **Property initializers** — ``public MyType Property { get; } = new MyType();``
- **Top-level statements** — ``var x = new MyType();`` at the top of a file

As a result, these `new T()` / `T with { ... }` call sites are **not wrapped** with `InitializableExtensions.WithInitialize(...)`, meaning `IInitializable.Initialize` is never invoked for these objects. This silently breaks the `IInitializable` contract.

## Location

`Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerAnalysisStep.OnInitializedCallSiteFinder.cs:159,182`

## Context

Flagged by @Copilot in PR #1541 review comment. The original PR (#1541, issue #1522) landed the initial `IInitializable` linker support, and this was deferred to a follow-up issue because the fix is not a one-line change.

## Design considerations

The core obstacle is that `ObjectCreationCallSiteReference` requires an `IMethodSymbol` for its `IntermediateSymbolSemanticKind.Default` semantic (via `IMethodSymbol.ToSemantic`), but field/property initializers and top-level statements do not map to a user-written method symbol.

Three possible strategies (none are trivial):

1. **Resolve to the synthetic constructor** that Roslyn generates to hold the field/property initializer (instance or static). `SemanticModel.GetEnclosingSymbol` returns the containing constructor for an initializer expression, so this may be viable — but we need to confirm the call-site linker then substitutes through the synthetic constructor correctly.
2. **Use `SemanticModel.GetEnclosingSymbol`** as a general fallback when `_containingMethod` is null, and teach the downstream substitution pipeline to handle any symbol kind (not just the method/accessor/local-function kinds currently supported).
3. **Extend `ObjectCreationCallSiteReference`** to allow a nullable containing-method semantic and add a new category of call site for "bare" initializer contexts.

The right path probably depends on whether field initializers are effectively "inside the implicit constructor" from the linker's substitution standpoint (strategy 1), or whether they need a new orthogonal category (strategy 3).

## Acceptance criteria

- `new T()` in a field/property initializer of a type that itself implements (or is contained in a type that implements) `IInitializable` is correctly wrapped with `WithInitialize`.
- Same for `with` expressions in field/property initializers.
- Same for top-level statements.
- Regression tests covering all three cases under `Metalama.Framework.Tests.AspectTests/Tests/Aspects/Initialization/`.

## Related

- PR #1541 — adds the `IInitializable` linker but intentionally leaves this case for follow-up
- Issue #1522 — parent initialization contract initiative

— Claude for @gfraiteur

gfraiteur:
## Summary

Fixes #1545 — OnInitializedCallSiteFinder previously skipped object-creation expressions inside field and property initializers, so `new T()` / `with T { ... }` in those positions was never wrapped with `InitializableExtensions.WithInitialize(...)`. This PR threads the call-site walker, substitution generator, analysis registry, and `LinkerRewritingDriver` to apply `WithInitialize` to field, event-field, and property initializers.

### Changes
- **Walker** (`LinkerAnalysisStep.OnInitializedCallSiteFinder`): collapsed `_containingMethod` to a single `_containingSymbol` slot and added a `VisitEqualsValueClause` override that attributes object creations inside field / event-field / property initializers to the declaring member.
- **Reference** (`LinkerAnalysisStep.ObjectCreationCallSiteReference`): replaced `ContainingSemantic` with a raw `ContainingSymbol`.
- **SubstitutionGenerator**: partitioned references into the existing method-body bucket and a new initializer-member bucket; `RunAsync` now returns a `SubstitutionGeneratorOutput` record.
- **AnalysisRegistry**: exposes `TryGetInitializerSubstitutions` / `HasInitializerSubstitutions`.
- **LinkerRewritingDriver**: `IsRewriteTarget` now considers initializer substitutions; new `LinkerRewritingDriver.Initializers.cs` adds a `RewriteInitializer` helper that uses bottom-up `ReplaceNodes` (correctly handles nested `new X { Y = new Y { Z = new Z() } }`); `RewriteField` and `RewriteProperty` invoke it.
- **OnInitializedCallSiteSubstitution**: when wrapping target-typed `new()` (`ImplicitObjectCreationExpressionSyntax`), emits an explicit generic type argument so the C# compiler can resolve `T` for `WithInitialize<T>(T)`.
- **LinkingRewriter**: special-cases multi-variable `FieldDeclarationSyntax` to call `RewriteMember` only once, since `RewriteField` processes the entire declaration internally.

### Out of scope
Top-level statement initializers (synthetic `<Main>` method) are not handled — filed as a follow-up.

## Test plan
- [x] 8 new aspect-test pairs under `Tests/Aspects/Initialization/` covering: simple field/property initializer, implicit `new()`, multi-variable field, lambda inside field initializer, `with` expression in property initializer, nested `new X { Y = new Y { Z = new Z() } }` in field initializer, and the equivalent nested case inside a method body (regression guard for the existing path).
- [x] All 89 `Initialization` tests pass (was 81 before).
- [x] All 2167 AspectTests pass (no regressions).
- [x] All 1811 UnitTests pass.
- [x] Zero build warnings.

— Claude for @gfraiteur

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerAnalysisStep.SubstitutionGenerator.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerAnalysisStep.SubstitutionGenerator.cs
@@ -59,6 +59,10 @@ internal sealed partial class LinkerAnalysisStep
             IntermediateSymbolSemantic<IMethodSymbol>,
             IReadOnlyList<ObjectCreationCallSiteReference>> _onInitializedCallSitesByContainingSemantic;
 
+        private readonly IReadOnlyDictionary<
+            ISymbol,
+            IReadOnlyList<ObjectCreationCallSiteReference>> _onInitializedCallSitesByInitializerMember;
+
         private readonly IReadOnlyDictionary<IntermediateSymbolSemantic<IEventSymbol>, EventBrokerTransformationInfo?> _eventBrokerSemanticIndex;
 
         private readonly IConcurrentTaskRunner _concurrentTaskRunner;
@@ -93,10 +97,32 @@ internal sealed partial class LinkerAnalysisStep
             this._forcefullyInitializedTypes = forcefullyInitializedTypes;
             this._eventBrokerSemanticIndex = eventBrokerSemanticIndex;
 
+            // Partition OnInitialized call sites into two buckets based on the kind of their
+            // containing symbol: method-body call sites drive the inlining pipeline (same path as
+            // today), while field/property/event-field initializer call sites are applied directly
+            // by LinkerRewritingDriver when rewriting the declaration.
+            var onInitializedMethodBodyCallSites = new List<ObjectCreationCallSiteReference>();
+            var onInitializedInitializerCallSites = new List<ObjectCreationCallSiteReference>();
+
+            foreach ( var reference in onInitializedCallSites )
+            {
+                if ( reference.ContainingSymbol is IMethodSymbol )
+                {
+                    onInitializedMethodBodyCallSites.Add( reference );
+                }
+                else
+                {
+                    onInitializedInitializerCallSites.Add( reference );
+                }
+            }
+
             this._additionalTransformedSemantics =
                 redirectedSymbolReferences.SelectAsReadOnlyList( x => (IntermediateSymbolSemantic) x.ContainingSemantic )
                     .Union( eventFieldRaiseReferences.SelectAsReadOnlyList( x => (IntermediateSymbolSemantic) x.ContainingSemantic ) )
-                    .Union( onInitializedCallSites.SelectAsReadOnlyList( x => (IntermediateSymbolSemantic) x.ContainingSemantic ) )
+                    .Union(
+                        onInitializedMethodBodyCallSites.SelectAsReadOnlyList(
+                            x => (IntermediateSymbolSemantic) ((IMethodSymbol) x.ContainingSymbol).ToSemantic(
+                                IntermediateSymbolSemanticKind.Default ) ) )
                     .Except( inlinedSemantics )
                     .Distinct()
                     .ToReadOnlyList();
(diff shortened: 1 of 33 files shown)


Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerAnalysisStep.SubstitutionGenerator.cs
```diff
@@ -93,10 +97,32 @@ public SubstitutionGenerator(
             this._forcefullyInitializedTypes = forcefullyInitializedTypes;
             this._eventBrokerSemanticIndex = eventBrokerSemanticIndex;
 
+            // Partition OnInitialized call sites into two buckets based on the kind of their
+            // containing symbol: method-body call sites drive the inlining pipeline (same path as
+            // today), while field/property/event-field initializer call sites are applied directly
```
gfraiteur: Remove "today" reference, as it does not make sense for future readers.

Metalama.Framework/src/Metalama.Framework.Engine/Linking/Substitution/SubstitutionContext.cs
gfraiteur: Make the field nullable.

Metalama.Framework/src/Metalama.Framework.Engine/Linking/Substitution/SubstitutionContext.cs
```diff
@@ -29,6 +29,20 @@ public SubstitutionContext(
...
+        /// initializer-rewrite path, which applies <see cref="OnInitializedCallSiteSubstitution"/>
+        /// instances explicitly and never calls <see cref="GetSubstitutions"/>.
+        /// </summary>
+        public SubstitutionContext(
+            LinkerRewritingDriver rewritingDriver,
+            SyntaxGenerationContext syntaxGenerationContext )
+        {
+            this.RewritingDriver = rewritingDriver;
+            this.SyntaxGenerationContext = syntaxGenerationContext;
+            this._substitutionDictionary = new Lazy<IReadOnlyDictionary<SyntaxNode, SyntaxNodeSubstitution>?>( () => null );
```
gfraiteur: Don't set the lazy field -- let null.

PR #1550, opened 2026-04-10
Implement OnConstructed multi-level inheritance and slot coordination (#1547)

### LINKED ISSUE #1547: Implement `InitializerKind.AfterLastInstanceConstructor`
## Summary

Implement `InitializerKind.AfterLastInstanceConstructor` (spec §6), the third phase of the initialization contract. Templates added with this kind run at the **end** of every instance constructor of the target type, with multi-level inheritance coordinated so that only the most-derived layer actually runs.

## Background

`InitializerKind.AfterLastInstanceConstructor` was declared in the public API and documented in `InitializationContract-spec.md §6`, but calling `AddInitializer(..., InitializerKind.AfterLastInstanceConstructor)` threw `AssertionFailedException: Unexpected initializer kind: AfterLastInstanceConstructor` from `ConstructorInitializeAdvice.cs`. No `OnConstructed` symbol existed anywhere in the engine.

## Final design

1. **Method introduction.** The aspect introduces a virtual `OnConstructed(InitializationContext context = default)` on the target type — `private` on sealed types and structs, `protected` otherwise. A derived layer that finds a base `OnConstructed` emits an override matching the base accessibility instead.

2. **Constructor epilogue.** Each instance constructor (except `:this(...)` chains) gets a trailing `this.OnConstructed(context);` call. Early `return`s in the constructor body are rewritten to `goto __metalama_epilogue;` so they still fall through the epilogue.

3. **`InitializationContext` parameter pull.** Constructors that don't already have an `InitializationContext` parameter get one introduced and flowed recursively through `:this(...)` / `:base(...)` chains and across project boundaries via the transitive-aspect mechanism in `PullConstructorParameterAdviceImpl`.

4. **Slot-based inheritance coordination.** An internal `InitializationSlot` ensures that when both `Base` and `Derived` apply this kind, only the most-derived `OnConstructed` body runs. The derived layer's override passes `context.Descend(slot)` to `base.OnConstructed(...)`; the base sees `IsHandled(slot) == true` and no-ops its body. Independent aspects each allocate their own slot, and the linker aggregates peer `base`-call transformations into a single `base.OnConstructed(context.Descend(slotA | slotB))` call.

5. **User-visible `slotFields`.** The `AddInitializer` overload accepts a `slotFields:` argument — references to `public static readonly InitializationSlot` fields on the aspect — letting aspects opt their own template bodies into the same `IsHandled`-based guarding that the framework uses internally.

6. **Coexistence with `AfterObjectInitializer`.** On the same type, `OnConstructed` runs at the end of the constructor body, before `Initialize` runs after object initialization.

7. **Records.** Positional and hand-written record constructors get the same treatment; copy constructors are excluded.

## Key engine pieces

- `AdviceImpl/Initialization/OnConstructedMethodAdvice.cs` — the advice
- `AdviceImpl/Initialization/OnConstructedEpilogueTransformation.cs` — the epilogue `this.OnConstructed(context)` injector
- `AdviceImpl/Initialization/BaseInitializationCallTransformation.cs` — shared base for the `base.OnConstructed` / `base.Initialize` prologue injectors
- `Linking/LinkerInjectionStep` — aggregates peer transformations via `IAggregatableInsertStatementTransformation.AggregateKey` so multiple aspects collapse into a single `base` call

Shipped in #1550.

— Claude for @gfraiteur

gfraiteur:
## Summary

Builds on Phase 1 (#15505063f6) and Phase 2 (#54bd69ff7b) to add multi-level inheritance and slot coordination for `InitializerKind.AfterLastInstanceConstructor`.

- **Multi-level inheritance**: derived constructors rewrite their `:base(...)` calls to pass `context.Descend(slot)` at every level, and the `OnConstructed` template guards its body with `context.IsHandled(slot)` so only the most-derived type whose aspect provided the slot runs the body.
- **Slot coordination across aspects**: added `IsOverride` flag on `IntroduceConstructorInitializerArgumentTransformation` with dedup in `MemberLevelTransformations.Sort()` so multiple aspects targeting the same `InitializationContext` parameter coexist.
- **Aggregation support**: `IInsertStatementTransformation.GetInsertedStatements` now takes an `aggregatedGroup` parameter plus a new `IAggregatableInsertStatementTransformation.AggregateKey` to let OnConstructed epilogues group across peer transformations.
- **`BaseConstructorResolver` helper**: shared by `SourceConstructor` and `IntroducedConstructor`. Fixes a **latent bug** where `IntroducedConstructor.GetBaseConstructor()` returned `null` when the base had no parameterless constructor, causing the Pull advice to silently skip derived classes. The `Pull_Two` and `Pull_CrossProject` baselines are updated to reflect the now-correct `p1` propagation to the derived class.
- **Runtime**: renamed `IsHandledBy` -> `IsHandled`; added `InitializationSlot.OnConstructed` system slot.

## Test coverage

- `OnConstructed_TwoLevel` / `_ImplicitCtor`
- `OnConstructed_SlotFields_TwoLevel` / `_TwoAspects`
- `OnConstructed_InheritedBase` (split from the old `_Diagnostic` test; the no-context-ctor case is now `OnConstructed_InheritedBase_NoContextConstructor_Diagnostic`)
- `OnConstructed_GenericBase` / `_ExplicitCtor`
- `OnConstructed_Linker_CrossProject`, `OnConstructed_Linker_CtorWithContext_CrossProject`
- `OnInitialized_SlotFields_TwoLevel` and its cross-project variant (verifies `slotFields` survive serialization via `PullConstructorParameterTransitiveAspect`)

## Test plan

- [x] `dotnet build` of `Metalama.Framework.Engine` and `Metalama.Framework.Tests.AspectTests` succeeds
- [x] All new `OnConstructed_*` aspect tests pass
- [x] `Pull_Two` and `Pull_CrossProject` pass against the updated baselines
- [x] Full `Build.ps1 build` green
- [x] TC build green

-- Claude for @gfraiteur


Metalama.Framework/src/Metalama.Framework.Engine/AdviceImpl/Initialization/InitializationContextParameterHelper.cs
```diff
@@ -0,0 +1,105 @@
...
+
+/// <summary>
+/// Helper that introduces (or reuses) an <c>InitializationContext</c> parameter on a constructor and
+/// recursively pulls it through constructor chains using the existing
+/// <see cref="PullConstructorParameterAdviceImpl"/> infrastructure. Also registers the transitive aspect
+/// so that derived types in other projects pick up the parameter.
+/// </summary>
+internal static class InitializationContextParameterHelper
+{
+    /// <summary>
```
gfraiteur: This this class is used from a single site, it should not be a helper class. Instead, the logic should be moved to the consuming class (OnConstructedMethodAdvice) for better clarity.

Metalama.Framework/src/Metalama.Framework.Engine/AdviceImpl/Initialization/OnConstructedBaseCallTransformation.cs
```diff
@@ -0,0 +1,172 @@
...
+
+/// <summary>
+/// Aggregatable transformation that emits the <c>base.OnConstructed(context.Descend(slotA | slotB | ...))</c>
+/// prologue statement at the top of an introduced <c>OnConstructed</c> override. When multiple aspects
+/// (each with its own set of slot fields) target the same override method, the linker aggregates the
+/// adjacent transformations and only the first one is asked to produce the statement — it then collects
+/// the slot field lists from all peers (including itself) and combines them with the <c>|</c> operator
+/// so a single <c>base.OnConstructed(context.Descend(...))</c> call covers all slots.
+/// </summary>
+internal sealed class OnConstructedBaseCallTransformation : BaseSyntaxTreeTransformation, IInsertStatementTransformation, IAggregatableInsertStatementTransformation
```
gfraiteur: There seems to be some redundancy with Metalama.Framework/src/Metalama.Framework.Engine/AdviceImpl/Initialization/InitializeBaseCallTransformation.cs - can this redundancy be reduced?

Metalama.Framework/src/Metalama.Framework.Engine/Transformations/IInsertStatementTransformation.cs
```diff
@@ -14,15 +14,45 @@ namespace Metalama.Framework.Engine.Transformations;
...
+/// multiple independent advices — for example, merging slot expressions from several advices into a single
+/// <c>base.OnConstructed(context.Descend(slotA | slotB))</c> call, or deduplicating identical epilogues produced
+/// by peer advices.
+/// </summary>
+internal interface IAggregatableInsertStatementTransformation : IInsertStatementTransformation
+{
+    /// <summary>
+    /// Gets the key used by the linker to aggregate transformations sharing the same target. Must be non-null.
+    /// </summary>
+    string AggregateKey { get; }
```
gfraiteur: Requiring a `string` might add an overhead while producing the string. Consider just `object`, and rely on `Equals` (not reference equality).

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Initialization/OnConstructed_Linker_CtorWithContext_CrossProject.cs
```diff
@@ -0,0 +1,15 @@
...
+// Refer to LICENSE.md in the repository root for complete details.
+
+using Metalama.Framework.RunTime.Initialization;
+
+namespace Metalama.Framework.Tests.AspectTests.Tests.Aspects.Initialization.OnConstructed_Linker_CtorWithContext_CrossProject;
+
+// <target>
+public class DerivedClass : BaseClass
+{
+    public DerivedClass( int value, InitializationContext context = default ) : base( value, context )
```
gfraiteur: This test is invalid. The hand-written code should have no `InitializationContext` parameter, and the test should demonstrate that the `InitializationContext  is pulled from the base across project boundaries.

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Initialization/OnConstructed_Sealed.t.cs
```diff
@@ -0,0 +1,13 @@
+[TheAspect]
+public sealed class TargetCode
+{
+  public TargetCode(int value, global::Metalama.Framework.RunTime.Initialization.InitializationContext context = default(global::Metalama.Framework.RunTime.Initialization.InitializationContext))
+  {
+    _ = value;
+    this.OnConstructed(context);
+  }
+  public void OnConstructed(global::Metalama.Framework.RunTime.Initialization.InitializationContext context = default)
```
gfraiteur: Note that OnConstructed must be `protected` for unsealed types, or `private` for sealed types. It can never be public.

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Initialization/OnConstructed_SlotFields.cs
```diff
@@ -0,0 +1,49 @@
...
+        builder.AddInitializer(
+            nameof(InitializerTemplate),
+            InitializerKind.AfterLastInstanceConstructor,
+            slotFields: new[] { slotField } );
+    }
+
+    [Template]
+    private void InitializerTemplate()
+    {
+        // Note: in Phase 1, slotFields is accepted by the API but the slot-coordination
```
gfraiteur: Comments referring to implementation phases are irrelevant in the PR. Flag and remove all.

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Initialization/OnConstructed_SlotFields.t.cs
```diff
@@ -0,0 +1,16 @@
+[TheAspect]
+public class TargetCode
+{
+  public TargetCode(int value, [global::Metalama.Framework.RunTime.AspectGeneratedAttribute] global::Metalama.Framework.RunTime.Initialization.InitializationContext context = default(global::Metalama.Framework.RunTime.Initialization.InitializationContext))
+  {
```
gfraiteur: This test doesn't demonstrate anything. Slots are only useful in inheritance scenarios. I think there is already a test with inheritance for this scenaro.

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Initialization/OnConstructed_Trackable_SingleLevel.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    public override void BuildAspect( IAspectBuilder<INamedType> builder )
+    {
+        builder.AddInitializer( nameof(OnConstructing), InitializerKind.BeforeInstanceConstructor );
+        builder.AddInitializer( nameof(OnConstructedTemplate), InitializerKind.AfterLastInstanceConstructor );
+        builder.AddInitializer( nameof(OnFullyInitialized), InitializerKind.AfterObjectInitializer );
+    }
+
+    [Template]
+    private void OnConstructing()
```
gfraiteur, line 43: Use the suffix Template everywhere or nowhere.

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Initialization/OnConstructed_TwoLevel.t.cs
```diff
@@ -0,0 +1,31 @@
+[TheAspect]
+public class BaseClass
+{
+  public BaseClass(int x, [global::Metalama.Framework.RunTime.AspectGeneratedAttribute] global::Metalama.Framework.RunTime.Initialization.InitializationContext context = default(global::Metalama.Framework.RunTime.Initialization.InitializationContext))
```
gfraiteur: Enable the simplifier in all intialization tests to make the test output human readable.

Metalama.Framework/src/Metalama.Framework.Engine/AdviceImpl/Initialization/InitializationContextParameterHelper.cs
gfraiteur: Inlined into OnConstructedMethodAdvice.EnsureContextParameter and deleted the helper class. — fixed in 9c896e1e33 — Claude for gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/AdviceImpl/Initialization/OnConstructedBaseCallTransformation.cs
gfraiteur: Extracted a shared BaseInitializationCallTransformation base class; both OnConstructedBaseCallTransformation and InitializeBaseCallTransformation are now thin sealed subclasses that only override the chained method name (OnConstructed vs Initialize). — fixed in 9c896e1e33 — Claude for gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/Transformations/IInsertStatementTransformation.cs
gfraiteur: Changed AggregateKey to object. The three implementers now use a private static readonly object _aggregateKey = new(); singleton per concrete type, and the linker compares keys with Equals. — fixed in 9c896e1e33 — Claude for gfraiteur

gfraiteur: Changed AggregateKey to object. The three implementers now use a private static readonly object _aggregateKey = new(); singleton per concrete type, and the linker compares keys with Equals. — fixed in 9c896e1e33 — Claude for gfraiteur

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Initialization/OnConstructed_Linker_CtorWithContext_CrossProject.cs
gfraiteur: Rewrote `DerivedClass` so it no longer hand-declares the `InitializationContext` parameter; it now mirrors `OnConstructed_Linker_CrossProject` and the baseline shows Metalama pulling the parameter across the project boundary. — fixed in 0ef241362f — Claude for gfraiteur

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Initialization/OnConstructed_Sealed.t.cs
gfraiteur: Generated `OnConstructed` is now `private` on sealed types and structs, `protected` otherwise; overrides inherit the base method's accessibility. All affected baselines updated. — fixed in 862faf63af — Claude for gfraiteur

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Initialization/OnConstructed_SlotFields.cs
gfraiteur: Removed — the whole `OnConstructed_SlotFields` file was deleted (see next comment). — fixed in 0cb3b9c373 — Claude for gfraiteur

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Initialization/OnConstructed_SlotFields.t.cs
gfraiteur: Agreed — deleted the test. Slot coverage is retained by `OnConstructed_SlotFields_TwoLevel` and `OnConstructed_SlotFields_TwoLevel_TwoAspects`. — fixed in 0cb3b9c373 — Claude for gfraiteur

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Initialization/OnConstructed_Trackable_SingleLevel.cs
gfraiteur: Renamed `OnConstructedTemplate` → `OnConstructed` so all three templates (`OnConstructing`, `OnConstructed`, `OnFullyInitialized`) follow the no-suffix convention. — fixed in 0cb3b9c373 — Claude for gfraiteur

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/Initialization/OnConstructed_TwoLevel.t.cs
gfraiteur: Enabled the simplifier folder-wide by adding `metalamaTests.json` with ``"FormatOutput": true`` under `Tests/Aspects/Initialization/`, and removed the now-redundant per-file `@FormatOutput` directives. — fixed in 0cb3b9c373 — Claude for gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerInjectionStep.Rewriter.cs
```diff
@@ -833,6 +850,52 @@ static BlockSyntax ReplaceBlock(
             {
                 if ( exitStatements.Count > 0 )
                 {
+                    // Constructor bodies can have arbitrary user code. The `AfterLastInstanceConstructor`
+                    // advice emits a trailing call to `OnConstructed(context)` as an epilogue statement;
+                    // we wrap the original body with entry/exit blocks and redirect any early `return;`
+                    // statements to the epilogue via a generated label.
+                    if ( declaration is IRef<IConstructor> )
+                    {
+                        const string epilogueLabelName = "__epilogue";
```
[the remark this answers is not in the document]
gfraiteur: No, this is too verbose. There is another way to ensure the name is unique, using the lexical scope, and this is already used by the linker to generate unique names by adding numeric suffixes. The same facility can be used here.

Metalama.Framework/src/Metalama.Framework.Engine/Linking/ConstructorEpilogueRewriter.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            return base.VisitReturnStatement( node );
+        }
+
+        this.HasRewrites = true;
+
+        // Trivia is intentionally dropped: the goto is a system-generated replacement and the
+        // linker's formatting pipeline will reformat the block after injection. The explicit
+        // `goto` keyword token with trailing space is required so the output reads
+        // `goto __metalama_epilogue;` and not `goto__metalama_epilogue;`.
```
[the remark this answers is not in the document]
gfraiteur: Should be __epilogue.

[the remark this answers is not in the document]
gfraiteur: Fixed in 93fee4b714 — updated comment to say `goto epilogue;` (the label is now allocated dynamically through the lexical scope, so referencing the old hardcoded name was misleading).

— Claude for gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/AdviceImpl/Initialization/OnConstructedMethodAdvice.cs
```diff
@@ -0,0 +1,414 @@
...
+namespace Metalama.Framework.Engine.AdviceImpl.Initialization;
+
+/// <summary>
+/// Implements <see cref="InitializerKind.AfterLastInstanceConstructor"/>: introduces (or reuses) a
+/// <c>public virtual void OnConstructed(InitializationContext context = default)</c> method on the
+/// target type, injects the template body into it, and emits a trailing <c>this.OnConstructed(context)</c>
+/// call at the end of every non-<c>:this(...)</c> instance constructor (after pulling an
+/// <c>InitializationContext</c> parameter through the constructor chain).
+/// </summary>
+/// <remarks>
```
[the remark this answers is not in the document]
gfraiteur: Fixed in 93fee4b714 — updated the XML summary to describe the actual accessibility rules (`private` on sealed types and structs, `protected virtual` otherwise, and a matching `override` when a base type already exposes `OnConstructed`).

— Claude for gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerInjectionStep.Rewriter.cs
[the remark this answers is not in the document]
gfraiteur: Fixed in 93fee4b714 — the label name is now allocated through the constructor's lexical scope via `LexicalScopeFactory.GetLexicalScope(constructorRef).GetUniqueIdentifier("epilogue")`. When no collision exists the name is simply `epilogue`; when the user already declares that label, `TemplateLexicalScope` falls back to `epilogue_1`, `epilogue_2`, etc. via its existing numeric-suffix mechanism.

I also added `OnConstructed_EarlyReturn_LabelCollision` as a regression test: the target constructor has its own `epilogue:` label and `goto epilogue;`, and the baseline now asserts that the user's label is preserved while the generated one becomes `epilogue_1:`.

— Claude for gfraiteur

PR #1614, opened 2026-05-03
Perf: remove NormalizeWhitespace from DesignTimeSyntaxTreeGenerator (#1601)

### LINKED ISSUE #1601: Perf: DesignTimeSyntaxTreeGenerator calls NormalizeWhitespace per generated type
`Metalama.Framework.Engine/Pipeline/DesignTime/DesignTimeSyntaxTreeGenerator.cs:248` calls `compilationUnit.NormalizeWhitespace()` once per design-time-generated type. `NormalizeWhitespace` walks the entire syntax tree and rebuilds trivia, which is expensive on every design-time pipeline run.

## History

- Added 2021-04-19 by Gael Fraiteur in commit `6ccb028582` (`#28345 Design-time source code generation`) as part of the original implementation. The compilation unit at this site is built entirely from `SyntaxFactory` calls (`CompilationUnit().WithMembers(...)`) and so has no trivia at all without normalization, making the IDE source-navigation view of generated code unreadable.
- 2024-01-04 commit `627c576a47` ("Don't call NormalizeWhitespace when not necessary") removed a separate per-member `NormalizeWhitespace` call inside `ProcessTransformationsOnType` in the same file, but left the compilation-unit-level call intact. So perf has been considered before; this remaining call is the last surviving one in this generator.

## Why it's worth revisiting

The `CodeFormatter` pipeline (`FormattingAnnotations`, `WithSimplifierAnnotation`, `CodeFormatter.CustomSimplifier`) already exists for production-time output and runs the full Roslyn `Simplifier` plus the Metalama custom simplifier. Design-time generation could either:

1. **Defer formatting** to that pipeline by emitting unformatted syntax with the same annotations production code uses, so formatting is paid for once in context.
2. **Skip normalization** if downstream consumers of `additionalSyntaxTreeDictionary` only need symbol resolution (not display text). Requires tracing consumers to confirm.

## Suggested next steps

- Trace consumers of `IntroducedSyntaxTree` / `additionalSyntaxTreeDictionary` from this generator.
- Decide between options 1 and 2 above based on what those consumers actually read.
- Benchmark before/after on a project with many design-time-introduced types (e.g., the WPF-WriteHtml repro added in this same milestone).

## Out of scope

This issue is split out from the LAMA0601 / `SyntaxTree.FilePath` fix landing alongside the WPF-WriteHtml repro; that fix only adds the `path:` argument and does not touch `NormalizeWhitespace`.

— Claude for @gfraiteur

PostSharpAgent:
## Summary

Removes the per-type `compilationUnit.NormalizeWhitespace()` call in `DesignTimeSyntaxTreeGenerator.ProcessTransformationsOnType` (issue #1601). The wrappers built in this generator now carry explicit elastic trivia so the produced syntax tree is parseable C# without normalization.

### What changed

- **Removed** `compilationUnit.NormalizeWhitespace()` / `NormalizeWhitespaceIfNecessary` from `ProcessTransformationsOnType`.
- **Added explicit elastic trivia** to the wrappers built in this generator:
  - Modifiers and type/namespace keywords now use `SyntaxFactoryEx.TokenWithTrailingSpace`.
  - Open/close braces in `CreatePartialType`, `CreateExtensionBlock`, the `NamespaceDeclaration` wrap, and the synthesized empty constructor body now carry `OptionalElasticEndOfLineTriviaList` leading and trailing trivia.
  - `GetArgumentRefToken` and `AddPartialModifierToTypes` emit `ref`/`out`/`partial` with trailing elastic spaces.
  - `CreateTypeParameters` emits the variance keyword (`in`/`out`) with a trailing elastic space — without this, `IRepository<in T>` rendered as `IRepository<inT>` (invalid C#).

### Test snapshot updates

Three `DesignTimePipelineTests` (`GenericTargetIntroductions`, `IntroducedSyntaxTreeConflictAndChange`, `ChangeInTargetCode`) compare the rendered output as a literal string. Their expected outputs are updated to match the new whitespacing. The new output is uglier in places (`Repository<T1,T2>`, `Id{get;set ;}`, no indentation on injected members) — these artifacts come from `IInjectMemberTransformation` outputs that production-time relies on the formatter to clean up — but is **syntactically valid C#** in every case. The wrappers this generator owns no longer produce any invalid output.

Closes #1601.

## Verification

- [x] `Build.ps1 build` (full Metalama build) succeeds with zero warnings.
- [x] `Metalama.Framework.Tests.AspectTests` — 2262 passed (`net8.0`), 2134 passed (`net48`).
- [x] `Metalama.Framework.Tests.UnitTests` — 1871 passed (`net8.0`), 1816 passed (`net48`); the three `DesignTimePipelineTests` previously requiring `NormalizeWhitespace` now pass with updated snapshots.

— Claude for @gfraiteur


gfraiteur (changes requested):
Remove NormalizeWhitespaceIfNecessary. Literal unit tests might be adjusted to reflect the new whitespacing, but only if it's syntactically correct.

PostSharpAgent:
@gfraiteur Done — `NormalizeWhitespace` is now fully removed from `DesignTimeSyntaxTreeGenerator`.

**What changed in this round (commit `d1c3d4e`):**
- Removed the `compilationUnit.NormalizeWhitespaceIfNecessary( syntaxGenerationContext )` call entirely.
- `CreateTypeParameters` now emits the variance keyword via `SyntaxFactoryEx.TokenWithTrailingSpace` — without normalization, `IRepository<in T>` was rendering as `IRepository<inT>` (invalid C#). All other wrappers already had the right elastic trivia from the previous round.
- Updated the three literal-snapshot tests (`GenericTargetIntroductions`, `IntroducedSyntaxTreeConflictAndChange`, `ChangeInTargetCode`) to reflect the new whitespacing. The new output is uglier in places (`Repository<T1,T2>`, `Id{get;set ;}`, no indentation on injected members) but is **syntactically valid C#** in every case — the artifacts come from `IInjectMemberTransformation` outputs that production-time relies on the formatter to clean up. None of the wrappers this generator owns produce invalid output.

**Verification (all green, no warnings):**
- `Build.ps1 build` — successful, zero warnings.
- `Metalama.Framework.Tests.UnitTests` — 1871 passed (net8.0), 1816 passed (net48).
- `Metalama.Framework.Tests.AspectTests` — 2262 passed (net8.0), 2134 passed (net48).

Re-requesting your review.

— Claude for @gfraiteur

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/Pipeline/DesignTimePipelineTests.cs
```diff
@@ -1013,26 +1019,35 @@ 3 introductions(s):
...
             /// <auto-generated>
             /// Generated by Metalama to support the code editing experience. This is NOT the code that gets executed.
             /// </auto-generated>
-            partial class Repository<T1, T2>
+
+
+            partial class Repository<T1,T2>
             {
-                public global::System.Int32 Id { get; set; }
+            public global::System.Int32 Id{get;set ;}
```
gfraiteur: Generate the missing spaces around brakets and after semicolon

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/Pipeline/DesignTimePipelineTests.cs
```diff
@@ -1013,26 +1019,35 @@ 3 introductions(s):
             /// <auto-generated>
             /// Generated by Metalama to support the code editing experience. This is NOT the code that gets executed.
             /// </auto-generated>
+
+
             partial interface IVariantRepository<in T>
             {
-                global::System.Int32 Id { get; set; }
+            global::System.Int32 Id{get;set ;}
```
gfraiteur: Generate missing indentation.

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/Pipeline/DesignTimePipelineTests.cs
```diff
@@ -457,11 +459,13 @@ 1 introductions(s):
...
 /// Generated by Metalama to support the code editing experience. This is NOT the code that gets executed.
 /// </auto-generated>
+
+
 partial class C
 {
-    public void NewProperty()
-    {
-    }
+public void NewProperty(){
```
gfraiteur: Generate line break before {

gfraiteur (changes requested):

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/Pipeline/DesignTimePipelineTests.cs
PostSharpAgent: Done — `AccessorListSpacingRewriter` now walks every accessor list in injected members and produces `{ get; set; }` with elastic spaces around the braces and between accessors. The stale trailing space on `set`/`init` is dropped when there is no body, so we no longer see `set ;`.

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/Pipeline/DesignTimePipelineTests.cs
PostSharpAgent: Done — `ProcessTransformationsOnType` now computes the type's nesting depth (containing types + namespace) and prepends each injected member with `newline + (typeDepth+1)*4 spaces`. Each nested-type wrapping iteration also indents the inner declaration at the outer's body level, so e.g. `Outer<T1>.Inner<T2>` produces `Inner` at 4 spaces and its property at 8.

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/Pipeline/DesignTimePipelineTests.cs
PostSharpAgent: Done — `PutBodyBracesOnOwnLines` rewrites the body's open and close brace leading trivia to `newline + memberIndent`, so methods, constructors, destructors, operators and conversion operators all render Allman-style:

```
public void NewProperty()
{

}
```

gfraiteur (changes requested):
The current implementation using methods like AddLeadingNewLineAndIndent are not much better than NormalizeWhitespace from a perf point of view because they post-format the code. The whole point of this PR is to emit the AST right from the first time -- no post-processing.

PostSharpAgent:
Reworked to push trivia into source construction. Pushed two commits:

- `da101cddbd`: Move accessor list and body brace trivia into the introduction transformations and `FormattedBlock`. `AccessorListSpacingRewriter`, `PutBodyBracesOnOwnLines` and `FormatInjectedMember` are gone.
- `2bb683e43f`: Combine the per-member leading-trivia indent with the body-brace-indent into a single `IndentMember` pass (one O(1) operation per member).

The remaining post-construction work in `DesignTimeSyntaxTreeGenerator` is just the per-member indent (member's leading trivia and, for members with a Block body, the body's brace leading trivia, both set to `newline + memberIndent*4 spaces`). The indent depth is computed from the type's nesting at this scope and cannot be known by the `IInjectMemberTransformation` implementations themselves, so it has to be applied here. It is a single `WithLeadingTrivia` (and at most one `WithBody`) call per member — orders of magnitude cheaper than `NormalizeWhitespace`'s full tree walk per type.

Specifically:

1. `SyntaxFactoryEx.FormattedAccessorList` constructs accessor lists with elastic spaces around the braces and trailing space on non-last bodyless semicolons (`{ get; set; }` rendering). Used from `IntroducePropertyTransformation`, `IntroduceIndexerTransformation`, `IntroduceEventTransformation`.
2. The bodyless `set`/`init` keyword no longer carries a trailing elastic space (so `set;` rather than `set ;`); the trailing space stays for the body case (`set { ... }`).
3. `ContextualSyntaxGenerator.FormattedBlock` emits the open brace with an elastic newline on both sides, so method/constructor/destructor bodies render Allman-style without a `NormalizeWhitespace` pass.
4. `DesignTimeSyntaxTreeGenerator.CreateInjectedConstructors` uses `FormattedBlock` for its empty-body construction, replacing the manual `Block(...)` call.

**Verification (net8.0 unless noted)**:
- `Metalama.Framework.Tests.UnitTests`: 1871/1871 passed, 6 skipped.
- `Metalama.Framework.Tests.UnitTests.4.12.0` DesignTime tests (net48 + net8.0): 760/760 passed.
- `Metalama.Framework.Tests.AspectTests`: 2261/2262 passed (the one failure, `CallStaticMethod`, is pre-existing on the `develop/2026.1` base and unrelated to this PR — verified by running it after stashing all my commits).
- `Build.ps1 build`: successful, **zero warnings**.

Re-requesting review.

— Claude for @gfraiteur

gfraiteur (approved):

gfraiteur (changes requested):
Address build failures https://postsharp.teamcity.com/buildConfiguration/Metalama_Metalama20261_Metalama_DebugBuild/323911

PR #1637, opened 2026-06-04
Fix #1636: crash combining IAspectWeaver-based aspects with Metalama.Patterns.Contracts

### LINKED ISSUE #1636: Using IAspectWeaver-based aspects (e.g. Metalama.Community.Virtuosity) together with   Metalama.Patterns.Contracts causes build crash
## Summary

  When a project uses any aspect backed by `IAspectWeaver` (such as `Metalama.Community.Virtuosity`) alongside `Metalama.Patterns.Contracts` attributes (such as `[NotNull]`) on method
  parameters, the build fails with a `LAMA0601` crash. The error occurs whenever both a weaver-based aspect and a contract attribute are present in the same compilation.

  ## Errors and Warnings

  ```
  CSC : error LAMA0601: Transformer 'MetalamaSourceTransformer' failed: One or more errors occurred.
  (The given key 'Metalama.Patterns.Contracts.NotNullAttribute:Build' was not present in the dictionary.)
  ```

  Full exception from the crash report:

  ```
  System.AggregateException: One or more errors occurred. (The given key 'Metalama.Patterns.Contracts.NotNullAttribute:Build' was not present in the dictionary.)
   ---> System.Collections.Generic.KeyNotFoundException: The given key 'Metalama.Patterns.Contracts.NotNullAttribute:Build' was not present in the dictionary.
     at Metalama.Framework.Engine.Pipeline.PipelineStepIdComparer.Compare(PipelineStepId x, PipelineStepId y)
     at Metalama.Framework.Engine.Pipeline.PipelineStepsState.AddAspectInstances(IEnumerable`1 aspectInstances)
     at Metalama.Framework.Engine.Pipeline.PipelineStepsState.ExecuteAspectSourceAsync(...)
     at Metalama.Framework.Engine.Pipeline.HighLevelPipelineStage.ExecuteAsync(...)
     at Metalama.Framework.Engine.Pipeline.AspectPipeline.ExecuteAsync(...)
     at Metalama.Framework.Engine.Pipeline.CompileTime.CompileTimeAspectPipeline.ExecuteAsync(...)
  ```

  ## Code to Reproduce

  **Project file:**

  ```xml
  <Project Sdk="Microsoft.NET.Sdk">
    <PropertyGroup>
      <TargetFramework>net10.0</TargetFramework>
      <Nullable>enable</Nullable>
      <ImplicitUsings>enable</ImplicitUsings>
      <OutputType>Exe</OutputType>
    </PropertyGroup>
    <ItemGroup>
      <PackageReference Include="Metalama.Community.Virtuosity" Version="2026.0.22" />
      <PackageReference Include="Metalama.Patterns.Contracts" Version="2026.0.23" />
    </ItemGroup>
  </Project>
  ```

  **MyService.cs:**

  ```csharp
  using Metalama.Community.Virtuosity;
  using Metalama.Patterns.Contracts;

  [Virtualize]
  public class MyService
  {
      public string Process([NotNull] string input) => input.ToUpper();
  }
  ```

  ### Workaround

  Add the following to the affected project. `ContractAspect` is in `Metalama.Framework` — no additional package reference needed:

  ```csharp
  using Metalama.Community.Virtuosity;

  [assembly: AspectOrder(
      AspectOrderDirection.CompileTime,
      typeof(ContractAspect),
      typeof(VirtualizeAttribute))]
  ```

  > Note: this issue affects all `IAspectWeaver`-based aspects, not just Virtuosity. Any weaver-backed aspect used alongside `Metalama.Patterns.Contracts` will trigger the same crash
  without explicit ordering.

  ## Versions

  - **Metalama:** 2026.0.23
  - **.NET SDK:** 10.0.300
  - **Target Framework:** net10.0
  - **OS:** Windows 11 Pro (10.0.26200)
  - **Processor Architecture:** AMD64

  ---

  *— Claude*

PostSharpAgent:
Fixes #1636.

## Problem

When an `IAspectWeaver`-based aspect (e.g. `Metalama.Community.Virtuosity`'s `[Virtualize]`) is used together with a `Metalama.Patterns.Contracts` attribute (e.g. `[NotNull]`) and there is **no explicit `[assembly: AspectOrder]`**, the compile-time pipeline crashes with an unhandled exception:

```
LAMA0601: Transformer ''MetalamaSourceTransformer'' failed
KeyNotFoundException: The given key ''...NotNullAttribute:Build'' was not present in the dictionary.
  at Metalama.Framework.Engine.Pipeline.PipelineStepIdComparer.Compare
  at Metalama.Framework.Engine.Pipeline.PipelineStepsState.AddAspectInstances
```

## Root cause

A `ContractAspect` has two layers: the default layer and a secondary `Build` layer, with an implicit `default -> Build` ordering. A weaver-based aspect has no ordering relationship with the contract, so the two are sorted only by name, which can place the weaver *between* the contract''s two layers (e.g. `[NotNull(default), Virtualize(weaver), NotNull:Build]`).

Because a weaver always forms its own low-level pipeline stage, this interleaving splits the contract aspect across two high-level stages, leaving its `Build` layer in a different stage than its default layer. The architecture requires all layers of one aspect to live in the same high-level stage (the aspect source is evaluated once, in the stage owning the default layer, and all layers are processed there) -- there is no mechanism to carry a secondary layer''s instances to a later stage. When `PipelineStepsState.AddAspectInstances` later asks the stage''s `PipelineStepIdComparer` to order the `Build`-layer step, that step is absent from the stage''s ordered-layers dictionary -> `KeyNotFoundException`.

This split only happens when the user has not specified an order between the weaver and the multi-layer aspect; once it occurs, there is no valid way for the pipeline to proceed, because the user''s intent (which side of the contract the weaver belongs on) is genuinely ambiguous.

## Fix

Rather than crash deep in the pipeline with an opaque `KeyNotFoundException`, detect the unsupported configuration up front in `AspectPipeline` and report a clear, actionable diagnostic.

- Added a pipeline validation (in `AspectPipeline.cs`) that detects when an aspect''s layers are split across pipeline stages by an intervening low-level (weaver) stage, and fails initialization gracefully instead of letting the later stage crash.
- Introduced a new general diagnostic descriptor **LAMA0042** (`AspectLayersSeparatedByWeaver`) that names the affected aspect and the separating weaver and tells the user to add an explicit `[assembly: AspectOrder(...)]` to order the weaver before or after all layers of the aspect.

This turns an unhandled crash into a precise error the user can act on, and the explicit `AspectOrder` they add resolves the ambiguity so both aspects work together.

## Tests

- **`Sdk/WeaverAndContractAspect`** -- reproduces the exact #1636 scenario (a weaver-based aspect + a `ContractAspect`-derived attribute, no explicit `AspectOrder`). Now expected to fail with `LAMA0042` instead of crashing.
- **`Sdk/WeaverAndContractAspectOrdered`** -- the explicitly-ordered counterpart with `[assembly: AspectOrder(...)]`. Compiles successfully, with both the weaver rewrite and the contract''s null-check applied to the target.

## Checklist
- [x] Reproduce the bug (failing regression test `Sdk/WeaverAndContractAspect`)
- [x] Implement the fix (LAMA0042 diagnostic instead of crash)
- [x] Add the ordered success-path test (`Sdk/WeaverAndContractAspectOrdered`)
- [x] Full `Build.ps1 build` (zero warnings)
- [x] Full test run -- net8.0: AspectTests 2265 passed / 0 failed, UnitTests 1893 passed / 0 failed
- [x] Finalize (marked ready, review requested)

-- Claude for @gfraiteur

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.Engine/AspectOrdering/AspectLayerSorter.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/AspectOrdering/AspectLayerSorter.cs
@@ -3,7 +3,6 @@
 // Refer to LICENSE.md in the repository root for complete details.
 
 using Metalama.Framework.Code.Collections;
-using Metalama.Framework.Engine.AspectWeavers;
 using Metalama.Framework.Engine.Aspects;
 using Metalama.Framework.Engine.Collections;
 using Metalama.Framework.Engine.Diagnostics;
@@ -234,8 +233,6 @@ internal static class AspectLayerSorter
             sortedIndexes[i] = i;
         }
 
-        bool IsWeaver( int index ) => unsortedAspectLayers[index].AspectClass.AspectDriver is IAspectWeaver;
-
         Array.Sort(
             sortedIndexes,
             ( i, j ) =>
@@ -250,23 +247,6 @@ internal static class AspectLayerSorter
                 if ( compareDistance != 0 )
                 {
                     return compareDistance;
-                }
-
-                // When two layers are not explicitly ordered relative to each other (i.e. they have the same
-                // topological distance), process layers backed by an IAspectWeaver (low-level aspects) first.
-                // A weaver always forms its own pipeline stage, so if it were ordered between two layers of the
-                // same high-level aspect (e.g. the default and the secondary "Build" layer of a ContractAspect)
-                // it would split that aspect across two high-level stages, which the pipeline does not support
-                // and which crashes with a KeyNotFoundException in PipelineStepIdComparer (issue #1636). The
-                // layers of a high-level aspect always belong to consecutive distance bands, so moving unordered
-                // weavers to the front of their band keeps every high-level aspect's layers within a single stage.
-                // This only affects layers that are unordered relative to each other; explicit ordering produces
-                // different distances and is therefore unaffected.
-                var compareWeaver = IsWeaver( j ).CompareTo( IsWeaver( i ) );
-
-                if ( compareWeaver != 0 )
-                {
-                    return compareWeaver;
                 }
 
                 // If two aspects are not explicitly ordered, we order them alphabetically.
--- a/Metalama.Framework/src/Metalama.Framework.Engine/AspectOrdering/AspectLayerSorter.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/AspectOrdering/AspectLayerSorter.cs
@@ -3,6 +3,7 @@
 // Refer to LICENSE.md in the repository root for complete details.
 
 using Metalama.Framework.Code.Collections;
+using Metalama.Framework.Engine.AspectWeavers;
 using Metalama.Framework.Engine.Aspects;
 using Metalama.Framework.Engine.Collections;
 using Metalama.Framework.Engine.Diagnostics;
@@ -233,6 +234,8 @@ internal static class AspectLayerSorter
             sortedIndexes[i] = i;
         }
 
+        bool IsWeaver( int index ) => unsortedAspectLayers[index].AspectClass.AspectDriver is IAspectWeaver;
+
         Array.Sort(
             sortedIndexes,
             ( i, j ) =>
@@ -247,6 +250,23 @@ internal static class AspectLayerSorter
                 if ( compareDistance != 0 )
                 {
                     return compareDistance;
+                }
+
+                // When two layers are not explicitly ordered relative to each other (i.e. they have the same
+                // topological distance), process layers backed by an IAspectWeaver (low-level aspects) first.
+                // A weaver always forms its own pipeline stage, so if it were ordered between two layers of the
+                // same high-level aspect (e.g. the default and the secondary "Build" layer of a ContractAspect)
+                // it would split that aspect across two high-level stages, which the pipeline does not support
+                // and which crashes with a KeyNotFoundException in PipelineStepIdComparer (issue #1636). The
+                // layers of a high-level aspect always belong to consecutive distance bands, so moving unordered
+                // weavers to the front of their band keeps every high-level aspect's layers within a single stage.
+                // This only affects layers that are unordered relative to each other; explicit ordering produces
+                // different distances and is therefore unaffected.
+                var compareWeaver = IsWeaver( j ).CompareTo( IsWeaver( i ) );
+
+                if ( compareWeaver != 0 )
+                {
+                    return compareWeaver;
                 }
 
                 // If two aspects are not explicitly ordered, we order them alphabetically.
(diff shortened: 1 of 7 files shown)


Metalama.Framework/src/Metalama.Framework.Engine/AspectOrdering/AspectLayerSorter.cs
```diff
@@ -225,6 +225,25 @@ private static bool TrySort(
...
         }
 
+        // Compute, for each aspect, the minimal topological distance among its layers. This is used as the primary
+        // sort key below so that all layers of the same aspect are kept contiguous in the final order. Without this,
+        // an unordered aspect (typically a weaver-based aspect, which has no ordering relationship with the others)
+        // could be sorted between two layers of the same aspect — e.g. between the default and the "Build" layer of a
+        // ContractAspect. Because a weaver-based aspect always forms its own (low-level) pipeline stage, that would
+        // split the aspect's layers into two different high-level stages, which the pipeline does not support and
+        // which crashes with a KeyNotFoundException in PipelineStepIdComparer (see issue #1636).
+        var minDistanceByAspect = new Dictionary<string, int>( StringComparer.Ordinal );
```
gfraiteur: This is the wrong fix. Layers are non-contiguous by design, specifically so that aspects can interweave each other.

gfraiteur (changes requested):

PostSharpAgent:
**Heads-up: revised the engine fix approach.**

An earlier commit on this branch (`fa653bbde7`, "keep aspect layers contiguous in `AspectLayerSorter`") sorted layers primarily by each aspect's *minimum* layer distance, forcing all layers of one aspect to be contiguous. That contradicts the framework's intended behavior — aspect layers are deliberately **interleaved by topological distance band** — and it broke 6 existing unit tests in `Metalama.Framework.Tests.UnitTests/AspectOrderingTests` (`TwoUnorderedDoubleLayerAspects`, `TwoPartiallyOrderedDoubleLayerAspects`, `TwoTotallyOrderedDoubleLayerAspects`, `InheritedAspects`, `OverlappingAndAdditionalLayersOnDerived`, `DuplicateLayersOnBaseAndDerived`), which assert interleaved layer orderings.

I replaced it with a more surgical fix (`c36c80fd90`): **among layers with equal topological distance (i.e. unordered relative to each other), sort `IAspectWeaver`-backed (low-level) layers first.** A weaver always forms its own pipeline stage; an unordered weaver landing between a high-level aspect's `default` and `Build` layers is what splits the aspect across two stages and crashes (#1636). The layers of a high-level aspect are always in consecutive distance bands, so pulling unordered weavers to the front of their band keeps each aspect's layers within a single stage. This only touches the tie-break between layers that are unordered relative to each other, so explicit `AspectOrder` (which yields different distances) and the existing interleaving tests are unaffected.

Verified:
- `AspectOrderingTests`: 15/15 pass.
- `Sdk/WeaverAndContractAspect`: passes (was the #1636 crash).

— Claude for @gfraiteur

gfraiteur:
The AI-generated PR was significantly modified: we now report an error in this situation instead of working around it.

gfraiteur (approved):

PR #1645, opened 2026-06-09
Fix [Observable] dropping semi-auto setter body with 'field' keyword (#1644)

### LINKED ISSUE #1644: [Observable] silently drops the setter body of a 'field'-keyword (semi-auto) property
Hi! 👋 Thanks for Metalama — we use `[Observable]` heavily.

We hit a correctness issue: when `[Observable]` is applied to a **semi-automatic property** (auto `get;` + a manual `set` that uses the C# `field` keyword), the aspect drops the user-written setter body. The synthesized backing field is still assigned, so the getter returns the value — but side effects in the setter (validation, forwarded writes, logging) silently vanish, with **no diagnostic**.

Looks like you've already captured this: the `.t.cs` snapshots added in #1219 (closing #1210) show the original logic moved into a `*_Source` helper that's never invoked from the public setter, and the auto-reviewer flagged all five tests the same way. I couldn't find an open issue tracking a fix, so filing one.

**Minimal repro** (no other dependencies):
```csharp
using System;
using Metalama.Patterns.Observability;

[Observable]
public class Foo
{
    public int Value { get; set { field = value; SideEffect++; } }
    public int SideEffect { get; private set; }
}

static class Program
{
    static int Main()
    {
        var foo = new Foo();
        foo.Value = 42;
        Console.WriteLine($"Value={foo.Value} (exp 42), SideEffect={foo.SideEffect} (exp 1)");
        return foo.SideEffect == 1 ? 0 : 1;
    }
}
```
**Expected:** `SideEffect == 1`. **Actual:** `SideEffect == 0` (`Value == 42`).

**Environment:** Metalama.Patterns.Observability 2026.1.17, Metalama.Framework 2026.1.17, Metalama.Compiler 2026.1.10; .NET SDK 10.0.300; `net10.0`, C# 14; Windows 11; command-line build, no IDE. Also repros on 2026.0.21 / 2026.0.23.

**Ask:** invoke the generated `*_Source` logic from the rewritten setter — or, until that's supported, emit a compile-time diagnostic instead of silently dropping the body. The silent part is the real trap; we ended up shipping our own analyzer to warn against the pattern.

Thanks! 🙏

PostSharpAgent:
Fixes #1644.

## Problem

When `[Observable]` (or any aspect) overrides a **C# 14 semi-automatic property** — a property with a compiler backing field **and** an explicit accessor body using the `field` keyword — the original setter body was silently dropped. The aspect linker treated the semi-auto property as a *pure* auto-property, so the override's proceed reference (`meta.Proceed()` / `FieldOrProperty.Value = value`) was redirected to a raw backing-field write, discarding validation / side effects / forwarded writes, with no diagnostic.

## Root cause & fix (Metalama.Framework.Engine linker)

`IsAutoProperty()` returns `true` for both `Auto` and `SemiAuto` property kinds. Three call sites used it to mean "has no real body", which is wrong for semi-auto:

1. **`LinkerAnalysisStep.SubstitutionGenerator`** — a property's default-semantic proceed reference is redirected to the backing field only when it is a **pure** auto-property (`GetPropertyKind() == PropertyKind.Auto`). Semi-auto now routes to the original accessor body (with `field` mapped to the backing field).
2. **`AspectReferenceSourceSubstitution`** — narrowed the matching assertion the same way, so semi-auto default semantics may legitimately route to the source member.
3. **`LinkerRewritingDriver.Properties`** — for a non-inlined semi-auto override target, the `_Source` member owns the storage (via its own `field`), so the otherwise-redundant promoted backing field is no longer emitted (avoids a dead field / CS0169).

## Result

```csharp
// before (buggy): original body parked in an uncalled Value_Source
set { if (_value != value) { _value = value; OnPropertyChanged("Value"); } }
// after (fixed): the original body runs
set { if (Value_Source != value) { Value_Source = value; OnPropertyChanged("Value"); } }
private int Value_Source { get => field; set { field = value; this.SideEffect++; } }
```

## Tests

- **New** `FieldKeyword_Override_SemiAutoTarget` (framework) — overrides a semi-auto property with a side-effecting setter; executes at runtime and asserts the body ran.
- **New** `FieldKeyword_SetterSideEffect` (Observability) — runtime regression test from the issue's repro; `SideEffect == 1`.
- **Updated** Observability `FieldKeyword_{ValueTypes,WithComputedProperty,WithInitializer,InpcProperty}` snapshots — now show the original body being invoked (validation/side effects preserved).
- **Updated** framework `Uninlineable_BackingFieldAccess` snapshot — routes through `_Source`, no dead backing field.

## Verification

- `Build.ps1 build`: **0 warnings**.
- Framework AspectTests (net8.0): **2268 passed**, 0 failed; net48 Overrides/Properties: 353 passed.
- Observability: AspectTests 65, UnitTests 38, CompileTimeTests 31 — all pass.

## Checklist

- [x] Phase 1 — Understand the issue
- [x] Phase 2 — Reproduce with a failing regression test
- [x] Phase 3 — Implement the fix
- [x] Phase 4 — Build & test (zero warnings)
- [x] Phase 5 — Finalize

— Claude for @gfraiteur

### DIFF
--- a/Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/CSharp14/FieldKeyword/FieldKeyword_Override_SemiAutoTarget.cs
+++ b/Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/CSharp14/FieldKeyword/FieldKeyword_Override_SemiAutoTarget.cs
@@ -1,0 +1,84 @@
+// Copyright (c) 2020-2025 SharpCrafters s.r.o. and contributors.
+// SharpCrafters s.r.o. licenses this file to you under either the MIT license or a proprietary license, depending on the repository from which it was obtained.
+// Refer to LICENSE.md in the repository root for complete details.
+
+#if TEST_OPTIONS
+// @RequiredConstant(ROSLYN_5_0_0_OR_GREATER)
+#endif
+
+#if ROSLYN_5_0_0_OR_GREATER
+
+using Metalama.Framework.Aspects;
+using Metalama.Framework.Code;
+using System;
+using System.Linq;
+
+namespace Metalama.Framework.Tests.AspectTests.Tests.Aspects.CSharp14.FieldKeyword_Override_SemiAutoTarget;
+
+// Overriding a semi-automatic property (C# 14 'field' keyword + explicit accessor body) must invoke the original
+// accessor body via meta.Proceed(), not silently drop it (regression test for #1644).
+internal class LoggingAspect : TypeAspect
+{
+    public override void BuildAspect( IAspectBuilder<INamedType> builder )
+    {
+        var property = builder.Target.Properties.Single( p => p.Name == "Value" );
+
+        builder.With( property ).Override( nameof(PropertyTemplate) );
+    }
+
+    [Template]
+    public int PropertyTemplate
+    {
+        get
+        {
+            Console.WriteLine( "Getting" );
+
+            return meta.Proceed();
+        }
+        set
+        {
+            Console.WriteLine( $"Setting to {value}" );
+            meta.Proceed();
+        }
+    }
+}
+
+// <target>
+[LoggingAspect]
+internal class C
+{
+    // Semi-automatic property whose manual setter has a side effect.
+    public int Value
+    {
+        get => field;
+        set
+        {
+            field = value;
+            this.SetterCallCount++;
+        }
+    }
+
+    public int SetterCallCount { get; private set; }
+}
+
+internal static class Program
+{
+    private static void TestMain()
+    {
+        var c = new C();
+        c.Value = 42;
+
+        Console.WriteLine( $"Value={c.Value} (expected 42)" );
+        Console.WriteLine( $"SetterCallCount={c.SetterCallCount} (expected 1)" );
+
+        if ( c.Value != 42 || c.SetterCallCount != 1 )
+        {
+            throw new InvalidOperationException(
+                $"The original semi-auto setter body was not executed: Value={c.Value} (expected 42), SetterCallCount={c.SetterCallCount} (expected 1)." );
+        }
+
+        Console.WriteLine( "Execution OK." );
+    }
+}
+
+#endif
--- a/Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerAnalysisStep.SubstitutionGenerator.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/Linking/LinkerAnalysisStep.SubstitutionGenerator.cs
@@ -704,7 +704,7 @@ internal sealed partial class LinkerAnalysisStep
                         break;
 
                     case { ResolvedSemantic: { Kind: IntermediateSymbolSemanticKind.Default, Symbol.Kind: SymbolKind.Property } }
-                        when nonInlinedReference.ResolvedSemantic.Symbol is IPropertySymbol property && property.IsAutoProperty() == true
+                        when nonInlinedReference.ResolvedSemantic.Symbol is IPropertySymbol property && property.GetPropertyKind() == PropertyKind.Auto
                                                                                                      && this._injectionRegistry.IsOverrideTarget( property ):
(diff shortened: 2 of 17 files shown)


Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/CSharp14/FieldKeyword/FieldKeyword_Override_SemiAutoTarget.cs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// <target>
+[LoggingAspect]
+internal class C
+{
+    // Semi-automatic property whose manual setter has a side effect.
+    public int Value
+    {
+        get => field;
```
gfraiteur, line 53: I want to see an example where the getter has no explicit implementation

gfraiteur (changes requested):

Metalama.Framework/src/tests/Metalama.Framework.Tests.AspectTests/Tests/Aspects/CSharp14/FieldKeyword/FieldKeyword_Override_SemiAutoTarget.cs
PostSharpAgent: Added `FieldKeyword_Override_SemiAutoTarget_AutoGetter`, which covers exactly that: the getter has no explicit implementation (auto `get;`) and the setter body uses the `field` keyword — the same shape as the original bug report.

The generated code confirms the auto getter reads the backing field directly while the explicit setter body is preserved and invoked:

```csharp
public int Value
{
    get
    {
        global::System.Console.WriteLine("Getting");
        return _value;
    }
    set
    {
        global::System.Console.WriteLine($"Setting to {value}");
        _value = value;          // original body invoked
        this.SetterCallCount++;  // side effect preserved
    }
}
```

Runtime output asserts `Value == 42` and `SetterCallCount == 1`. All 19 `FieldKeyword` aspect tests pass.

— Claude for @gfraiteur

gfraiteur (approved):

gfraiteur (changes requested):
Fix the build issue:

C:\wk2\c33f66953b1f6eb\Metalama.Patterns\src\tests\Metalama.Patterns.Observability.AspectTests\FieldKeyword_SetterSideEffect.cs(33,24): error CS0017: Program has more than one entry point defined. Compile with /main to specify the type that contains the entry point. [C:\wk2\c33f66953b1f6eb\Metalama.Patterns\src\tests\Metalama.Patterns.Observability.AspectTests\Metalama.Patterns.Observability.AspectTests.csproj]

PostSharpAgent:
### Build fix pushed

Fixed the `CS0017: Program has more than one entry point defined` error.

**Cause.** The Observability regression test `FieldKeyword_SetterSideEffect.cs` carried its own `Program.Main` for runtime verification. The `Metalama.Patterns.Observability.AspectTests` project references `Microsoft.NET.Test.Sdk`, which auto-generates a test-host entry point — so a second `Main` produced two entry points. (Memoization avoids this only because it is a single-program project with an explicit `<StartupObject>`.)

**Fix.** Aligned the test with the convention used by all five sibling `FieldKeyword_*` tests in this project, which are **snapshot-only** (no `Program.Main`, no `.t.txt`):
- Removed the `Program` class from `FieldKeyword_SetterSideEffect.cs`.
- Deleted the (unasserted) `FieldKeyword_SetterSideEffect.t.txt`.

The `.t.cs` snapshot still proves the fix — the public setter routes through `Value_Source`, so the original body (`field = value; SideEffect++;`) runs instead of a raw backing-field write. If the bug regresses, the snapshot changes and the test fails.

Runtime, throw-based verification is unaffected: it lives in the framework test `FieldKeyword_Override_SemiAutoTarget` (uses `TestMain`, no entry-point collision).

**Verification.**
- `Build.ps1 build`: success, 0 errors.
- `Metalama.Patterns.Observability.AspectTests` (net8.0): **65 passed, 0 failed** — the project now compiles (CS0017 gone) and all `FieldKeyword_*` snapshots are green.

— Claude for @gfraiteur

PR #1658, opened 2026-06-10
Security: validate temp directory is not writable by other users (#1650)

### LINKED ISSUE #1650: Security: predictable world-writable temp artifacts loaded/executed without integrity check enables DLL planting (Linux/macOS)
## Summary

Everything Metalama later **loads and executes** (compile-time assemblies, extracted Backstage tools, bootstrap dependency DLLs) is written under `Path.GetTempPath()/Metalama/ΓÇª` with **predictable path names**, **no restrictive ACL**, and is trusted on **file existence alone** ΓÇö no hash or signature is verified before the artifact is loaded into the compiler / IDE / MSBuild process.

## Threat model

Source code being compiled is TRUSTED. The relevant untrusted boundary here is **other local users / lower-privilege processes on a shared or CI machine**, plus planted files. On **Linux/macOS** `Path.GetTempPath()` resolves to `$TMPDIR` or `/tmp` ΓÇö a world-writable directory (sticky bit only) ΓÇö which is exactly the shared build-agent scenario. On **Windows** the exposure is largely mitigated because `%LOCALAPPDATA%\Temp` is per-user.

## Finding ΓÇö DLL planting via predictable, existence-trusted temp artifacts (HIGH on Linux/macOS shared/CI)

Three load-and-execute sinks share the same weakness:

1. **Compile-time assembly cache.** `CheckCompileTimeProjectDiskCache` / `CheckManifestDiskCache` validate the cache by `File.Exists` only ΓÇö no integrity check ΓÇö `CompileTimeCompilationBuilder.cs:885-949`. On a cache hit, `CompileTimeDomain.LoadAssemblyCore` (`CompileTimeDomain.cs:161-186`) loads/executes the PE. The path `{tmp}/Metalama/CompileTime/{name}/{tfm}/{projectHash:x16}/Metalama_{name}_{hash:x16}.dll` is fully deterministic.

2. **Backstage tools extraction.** `BackstageToolsExtractor.cs:31-50` skips extraction entirely if the `unzipped.touch` sentinel already exists; `BackstageToolsExecutor` then runs `dotnet "{dir}/Worker.dll"`. The Worker is launched during background activity (telemetry upload, UI), not just on explicit user action.

3. **`ResourceExtractor`.** `ResourceExtractor.cs:216-348` skips extraction on a `.completed` sentinel, then loads via `Assembly.LoadFile`/`LoadFromPath`. This runs very early ΓÇö it bootstraps the compiler extension itself.

**Exploit (Linux/macOS shared or CI host):** a lower-privilege user (or an earlier CI job) pre-creates the predictable directory tree containing a malicious DLL plus the relevant sentinel file (`unzipped.touch` / `.completed` / a matching `manifest.json`). When a higher-privilege user later builds, Metalama finds the "valid" cache, skips regeneration, and loads/executes the planted code in the victim''s context ΓåÆ local privilege escalation / code execution. The sticky bit on `/tmp` does not help: the attacker owns the directory it created. A TOCTOU swap of the PE between the existence check and the load is a secondary variant.

## Amplifier ΓÇö `METALAMA_TEMP` honored without ownership validation

Both the Backstage path layer (`MetalamaPathUtilities.cs:20-24`) and the early `ResourceExtractor` (`ResourceExtractor.cs:67-68`) read the `METALAMA_TEMP` environment variable and use it verbatim as the cache / extract / assembly-load root, with no ownership/permission check. The environment is a separate, lower-trust channel: a process that can influence the victim''s environment (a CI step, a wrapper script, a parent process) can point `METALAMA_TEMP` at an attacker-controlled directory ΓÇö turning the planting attack above into reliable code execution **even on Windows**, bypassing the per-user `%LOCALAPPDATA%\Temp` ACL that otherwise mitigates this.

## Fix

> **Resolution (by design).** Our trust boundary is the **OS user account**: Metalama trusts anything writable by the current user, and the threat addressed here is *other* (lower-privilege) users on a shared / CI machine. The fix relocates the load-and-execute caches off world-writable `/tmp` to the **per-user application-data directory**, where the OS already guarantees owner-only write access. That OS-enforced ownership **is** the validation, so we deliberately do **not** add explicit directory/sentinel permission checks or artifact integrity verification (items 2-3 below): within a single user account everything is same-trust, so those checks would guard a boundary that does not exist. `METALAMA_TEMP` is honored as-is (item 4 not pursued) -- redirecting it, or planting inside the user''s own directory, requires already executing in that user''s context, which is not a privilege escalation. Only the relocation (item 1) shipped, and that is considered sufficient.

- On POSIX, root these caches/extraction dirs under a per-user location (`$XDG_CACHE_HOME` / `$HOME/.cache/Metalama`) or create `{tmp}/Metalama` once with mode `0700`; on Windows set an explicit `DirectorySecurity` (current user + SYSTEM/Admins).
- Verify the cache/tool/extract directory and its sentinel are **owned by the current user** before trusting them; re-extract or fail otherwise.
- Integrity-check executable artifacts (strong hash stored in the manifest, re-hashed on load) before loading ΓÇö do not treat `.completed` / `unzipped.touch` / mere file existence as proof of authenticity. `ResourceExtractor` already has `StreamsContentsAreEqual`; run it unconditionally on the cache-hit path, not only on sharing violations.
- When `METALAMA_TEMP` points to an existing directory, validate its ownership/permissions before use; create it `0700` / with a restrictive `DirectorySecurity` if absent.

ΓÇö Claude for Gael

PostSharpAgent:
## Security: predictable world-writable temp artifacts (#1650)

Fixes #1650.

Metalama writes artifacts it later loads/executes — the compile-time assembly cache and extracted Backstage tools — under the temp directory (`METALAMA_TEMP` or the default `%TEMP%\Metalama` / `$TMPDIR/Metalama`). On a host where the temp root is writable by other users (e.g. `/tmp` on Linux/macOS, or a machine-wide `C:\Temp`), a lower-privilege user can plant a DLL that Metalama then trusts on file existence and loads → local code execution / privilege escalation.

### Approach (per @gfraiteur's guidance)

> verify that `METALAMA_TEMP` (or its default value, `%TEMP%\Metalama`) is not writable by other users. This check can be performed once, and the result can be stored in a static variable.

`MetalamaPathUtilities.GetTempPath()` now performs a one-time, statically-cached `EnsureTempPathIsSecure`:

1. Creates `{temp}/Metalama` if absent.
2. If the directory is writable by other users — `IsDirectoryWritableByOtherUsers`:
   - **Windows**: any *Allow* ACE granting an atomic write/modify right to an identity other than the current user, SYSTEM, Administrators or CREATOR OWNER. Composite rights (Write/Modify/FullControl) are intentionally excluded from the mask because their numeric value also contains read/execute bits (which would wrongly flag a read-only shared directory); composite grants are still caught because their value includes the atomic write bits.
   - **Unix**: the group- or other-write mode bit is set (`File.GetUnixFileMode` on .NET 8; via reflection on the netstandard2.0 build, which runs on .NET 7+).
   - it tightens the permissions to the current user (protected ACL on Windows / `0700` on Unix), retrying a few times to tolerate concurrent compiler processes.
3. Throws a `SecurityException` with a remediation hint **only** if the directory is *still* writable by other users after that attempt (e.g. it is owned by another user).

This secures the directory rather than failing the build on machines that legitimately use a shared temp root (validated: the consolidated build runs every `csc` invocation through this path and `C:\Temp\Metalama` ends up restricted to SYSTEM / Administrators / current user, inheritance disabled).

### Tests
`MetalamaPathSecurityTests` (3 tests, green on net8.0 + net472): writable-by-everyone ⇒ flagged; current-user-only ⇒ not flagged; readable-but-not-writable-by-others ⇒ not flagged (this case caught a real masking bug). Full `Metalama.Backstage.Tests` suite passes on both TFMs (1253/1252, 0 failures); full `Build.ps1 build` succeeds with zero warnings.

### Open question for @gfraiteur
The third sink listed in the issue, `ResourceExtractor` (`Metalama.Framework.CompilerExtensions`, netstandard2.0), reads `METALAMA_TEMP` independently, has **no** reference to `Metalama.Backstage`, and runs the earliest (it extracts and loads Backstage itself). It is therefore *not* covered here. That project deliberately avoids `System.Security.AccessControl` (its `MutexAcl` uses raw P/Invoke + SDDL), so covering it cleanly means duplicating a platform-specific check there (and in the `Metalama.Compiler.Shared` copy). I kept it out to avoid risk in the bootstrap path — let me know if you want it included.

### Checklist
- [x] Phase 1 — Understand
- [x] Phase 2 — Failing regression test
- [x] Phase 3 — Implement the check + securing
- [x] Phase 4 — Build & test (zero warnings, all Backstage tests green, full build validated)
- [x] Phase 5 — Ready for review

— Claude for Gael

### DIFF
--- a/Metalama.Backstage/src/Metalama.Backstage/Utilities/MetalamaPathUtilities.cs
+++ b/Metalama.Backstage/src/Metalama.Backstage/Utilities/MetalamaPathUtilities.cs
@@ -6,7 +6,17 @@
 
 using JetBrains.Annotations;
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Runtime.InteropServices;
+using System.Security;
+using System.Security.AccessControl;
+using System.Security.Principal;
+#if NET
+using System.Runtime.Versioning;
+#else
+using System.Globalization;
+#endif
 
 namespace Metalama.Backstage.Utilities;
 
@@ -14,6 +24,8 @@ namespace Metalama.Backstage.Utilities;
 public static class MetalamaPathUtilities
 {
     private static readonly string? _overriddenTempPath;
+    private static readonly object _tempPathVerificationSync = new();
+    private static volatile bool _tempPathVerified;
 
     static MetalamaPathUtilities()
     {
@@ -21,7 +33,14 @@ public static class MetalamaPathUtilities
         _overriddenTempPath = string.IsNullOrEmpty( overriddenTempPath ) ? null : overriddenTempPath;
     }
 
-    public static string GetTempPath() => _overriddenTempPath ?? Path.GetTempPath();
+    public static string GetTempPath()
+    {
+        var tempPath = _overriddenTempPath ?? Path.GetTempPath();
+
+        VerifyTempPathIsSecure( tempPath );
+
+        return tempPath;
+    }
 
     public static string GetTempFileName()
     {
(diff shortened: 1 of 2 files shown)


gfraiteur:
> The third sink listed in the issue, ResourceExtractor (Metalama.Framework.CompilerExtensions, netstandard2.0), reads METALAMA_TEMP independently, has no reference to Metalama.Backstage, and runs the earliest (it extracts and loads Backstage itself). It is therefore not covered here. That project deliberately avoids System.Security.AccessControl (its MutexAcl uses raw P/Invoke + SDDL), so covering it cleanly means duplicating a platform-specific check there (and in the Metalama.Compiler.Shared copy). I kept it out to avoid risk in the bootstrap path — let me know if you want it included.

@PostSharpAgent Yes, code should be duplicated.

gfraiteur:
We're taking another approach.

PR #1697, opened 2026-06-17
Serialize in-process telemetry config updates (#1696)

### LINKED ISSUE #1696: Telemetry configuration update contention when building large solutions
## Symptom

When building a solution with many projects (e.g. `Metalama.Samples`), the diagnostic log contains:

```
ERROR, Thread 21, Configuration: Too many attempts to update the configuration TelemetryConfiguration. There must be an unaddressed race condition.
```

## Root cause

`UsageReporter` is a per-process singleton. For each compiled project it reports a usage session via `ShouldCollectMetrics` → `UpdateIf<TelemetryConfiguration>`, which appends to the shared `Sessions` map.

`UpdateIf`/`Update` use optimistic concurrency: they read the config + timestamp, compute the new value, then call `TryUpdate(value, expectedTimestamp)`. The cross-process mutex is held only for the check+write inside `TryUpdate` — the read and the write are not atomic. When one process compiles many projects concurrently (design-time analysis service or in-process build), many threads hit the singleton at once. They are not racing on data (different project keys) but on the file timestamp: each commit bumps it and invalidates everyone else''s in-flight read, so they ping-pong through retries until one exhausts the 10-attempt limit and logs the error (the session is then silently not deduplicated).

## Fix

- Serialize the in-process callers with a lock in the `UsageReporter` singleton so each read-modify-write completes on its first attempt. This removes the in-process thread storm, which is the driver. Cross-process contention remains bounded by the existing cross-process mutex and has far fewer contenders.
- Enrich the "Too many attempts" error log with `Environment.StackTrace` (cold path, only after 10 failed attempts) so any future recurrence — e.g. a cross-process case or an unexpected writer — is self-diagnosing from the log.

— Claude for @gfraiteur

gfraiteur:
Fixes #1696.

## Problem

Building a solution with many projects (e.g. `Metalama.Samples`) logs:

```
ERROR, Configuration: Too many attempts to update the configuration TelemetryConfiguration. There must be an unaddressed race condition.
```

`UsageReporter` is a per-process singleton that reports a usage session per project via `UpdateIf<TelemetryConfiguration>` (optimistic read-modify-write). The cross-process mutex only guards the check+write in `TryUpdate`, so the read and write are not atomic. Many concurrent in-process compilations bump the file timestamp under each other, invalidating in-flight reads, so the 10-attempt retry budget is exhausted and the session is silently not deduplicated.

## Fix

- Serialize the in-process callers with a `lock` in the `UsageReporter` singleton, so each read-modify-write completes on its first attempt. This removes the in-process thread storm (the driver). Cross-process contention stays bounded by the existing cross-process mutex and has far fewer contenders.
- Include `Environment.StackTrace` in the "too many attempts" error (cold path only) so any future recurrence is self-diagnosing from the log.

## Validation

`Metalama.Backstage` builds with 0 warnings; all 13 `UsageReporterTests` pass on net8.0 and net472 (including the concurrent-sessions test).

— Claude for @gfraiteur

### DIFF
--- a/Metalama.Backstage/src/Metalama.Backstage/Configuration/ConfigurationManagerExtensions.cs
+++ b/Metalama.Backstage/src/Metalama.Backstage/Configuration/ConfigurationManagerExtensions.cs
@@ -51,8 +51,10 @@ namespace Metalama.Backstage.Configuration
                     // We no longer throw an exception here because we have a known random issue and throwing an exception seems to be worse
                     // than ignoring it.
 
+                    // Include the call stack so that, if this recurs, the contending caller can be identified from the log.
                     configurationManager.Logger.Error?.Log(
-                        $"Too many attempts to update the configuration {typeof(T).Name}. There must be an unaddressed race condition." );
+                        $"Too many attempts to update the configuration {typeof(T).Name}. There must be an unaddressed race condition.{
+                            Environment.NewLine}{Environment.StackTrace}" );
 
                     return false;
                 }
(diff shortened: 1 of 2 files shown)


gfraiteur:
Addressed the review note: softened the `_sync` comment so it no longer claims a first-attempt guarantee, and instead states the lock removes the dominant in-process self-contention while occasional retries remain possible from other (infrequent) `TelemetryConfiguration` writers or other processes (bounded by the cross-process mutex).

— Claude for @gfraiteur

Issue #1206, opened 2025-11-27
Allow disabling telemetry and metrics via MSBuild properties


Roman-Soloviev:
Currently, Metalama telemetry can be disabled only via:

- the METALAMA_TELEMETRY_OPT_OUT environment variable, or
- the Metalama CLI (metalama telemetry disable).

And the documentation does not mention any MSBuild properties such as MetalamaMetricsEnabled or MetalamaTelemetryEnabled.
At the same time, Metalama runs several MSBuild tasks (e.g., metrics collection) during the build process, and there is no clean or documented way to disable this behavior at the project or solution level.

### Motivation

For team-based environments and CI/CD pipelines, it is very useful to disable telemetry or metrics per project/solution, in a way that:

- is version-controlled,
- does not require external environment configuration,
- applies automatically to all developers and build agents,
- works even in isolated or restricted environments (security policies).

A common industry pattern is to expose toggleable MSBuild properties that developers can put into Directory.Build.props.
Example:

```
<PropertyGroup>
    <MetalamaTelemetryEnabled>false</MetalamaTelemetryEnabled>
    <MetalamaMetricsEnabled>false</MetalamaMetricsEnabled>
</PropertyGroup>
```

gfraiteur:
Thank you for your suggestion. 

I understand the need for a source-controlled configuration that works seamlessly in CI/CD pipelines without relying on external setup. That’s a valid concern.

Currently, MSBuild properties aren’t a viable option because telemetry initializes very early in static constructors, and sometimes even outside of a build context (for example, during CLI commands or Visual Studio tooling). At that point, MSBuild properties are not yet available.

The approach we’ve taken follows industry precedent: the .NET SDK telemetry uses environment variables for opt-out, and we’ve adopted the same pattern for consistency.

If you want to enforce this in a source-controlled way, one workaround is to add a check in `Directory.Build.props` that raises an error if the `METALAMA_TELEMETRY_OPT_OUT` environment variable isn’t defined. 

That said, I’m open to delaying telemetry until build properties are known. This would involve a refactoring of the initialization sequence (which is already quite complex), but it's feasible.

If you're an enterprise customer, please email us so we can prioritize the feature.

PostSharpAgent:
## Phase 1: Understanding

[...]

PR #1702, opened 2026-06-19
Disable telemetry per repository via metalama.json (#1701)

### LINKED ISSUE #1701: Disable Metalama telemetry per repository with a metalama.json file
Metalama's telemetry can currently be disabled only through the `METALAMA_TELEMETRY_OPT_OUT` environment variable or the `metalama` CLI. Both must be configured on each machine, and neither is captured in source control. Teams that want a consistent, environment-agnostic opt-out — for privacy or compliance reasons, for locked-down or air-gapped environments, or simply as team policy — have no clean way to state once that a repository does not send telemetry and have that apply everywhere.

This story adds a way to disable Metalama telemetry for an entire repository through a configuration file checked into source control, so the setting applies automatically to every developer and every CI/CD agent that builds the repository, with no per-machine setup. It also stops the `metalama` global dotnet tool from sending telemetry.

## Behavior

A new **`metalama.json`** file, placed at the **repository root**, controls the setting. To disable telemetry for the whole repository:

```json
{
  "telemetry": {
    "enabled": false
  }
}
```

- When present at the repository root with `telemetry.enabled` set to `false`, Metalama collects and sends **no** telemetry for that repository — for command-line/CI builds, for design-time analysis, and in Visual Studio.
- Because the file is committed to source control, the setting applies automatically to everyone building the repository — no environment variables, no per-machine commands.
- `metalama.json` is a **general Metalama repository configuration file**; telemetry is its first setting, and more may be added over time.
- The file must live at the **repository root**. A `metalama.json` found anywhere else (for example, a project subfolder) produces a **warning** and is ignored.
- A malformed `metalama.json` is **ignored** (it has no effect on telemetry) and reported as a diagnostic; it never breaks the build.
- The **`metalama` global dotnet tool** sends no telemetry.
- This complements the existing options: the `METALAMA_TELEMETRY_OPT_OUT` environment variable remains the complete, machine-wide opt-out and continues to take priority.

## Expected outcomes

- A repo-root `metalama.json` with `telemetry.enabled = false` disables telemetry for command-line/CI builds of projects in that repository.
- The same setting disables telemetry during design-time analysis and in Visual Studio for that repository/solution.
- On a clean machine where the repository has opted out, Metalama does not begin collecting telemetry and shows no "telemetry is enabled" first-run notice.
- A `metalama.json` placed outside the repository root produces a warning and is ignored.
- A malformed `metalama.json` is ignored and reported as a diagnostic, without failing the build.
- Omitting the setting (or `enabled: true`) preserves the normal default behavior and never overrides the environment variable.
- The `metalama` global dotnet tool does not send telemetry.
- Documentation describes the file, its schema, what it covers (builds, design-time, IDE), and clarifies that the environment variable remains the complete machine-wide opt-out.

— Claude for @gfraiteur

gfraiteur:
## Summary
- Rework Metalama telemetry to be **context-scoped**: telemetry now requires an `ITelemetryContext` obtained from `ITelemetryService.OpenContext(directory)`, which resolves a repo-root `metalama.json` opt-out and folds in the existing process-level gates.
- **Lazy activation**: no `DeviceId`/salts and no `telemetry.json` write until an *enabled* context actually reports (`EnsureActivated`). Adds `OnActivated(Action)` (race-free, this-session) and `IsActivated`.
- **Repo opt-out** via a repo-root `metalama.json` (`{ "telemetry": { "enabled": false } }`), discovered by walking up to the `.git` root. A misplaced (non-root) or unconfirmed-root file produces a warning; a malformed file is ignored.
- Surface invalid-config warnings as compiler diagnostic **`LAMA0075`**, located at the `metalama.json` file.
- Formalize the **ASK / Yes / No** `ReportingAction` model for exception & performance (Default = capture + ask, Yes = capture + auto-send, No = do not capture/ask); add `GetEffectiveReportingAction(scenario)` as the universal accessor and guard `IsEnabled(scenario)` to usage-style scenarios.
- Rename `WelcomeService` -> **`WelcomePageService`**: opens on first activation, gated by the new `BackstageInitializationOptions.OpenWelcomePage` (command-line compiler on, VS off). First-run telemetry toast is gated on activation, with a throttle-bypassing re-detection.
- Disable telemetry in the **`metalama` dotnet tool**.

## Breaking Changes
- `Metalama.Backstage.Welcome.WelcomeService` renamed to `WelcomePageService`; `OpenWelcomePageIfRequired()` is now private (the page opens via `OnActivated`).
- `ITelemetryConfigurationService.IsEnabled(TelemetryScenario)` now throws for `Exception`/`Performance` — use `GetEffectiveReportingAction(scenario)`. New members: `EnsureActivated`, `IsActivated`, `OnActivated`, `GetEffectiveReportingAction`. `Initialize()` no longer activates telemetry.
- New `BackstageInitializationOptions.OpenWelcomePage` (default `false`).
- Telemetry reporting now flows through `ITelemetryContext` (`StartUsageSession` / `ReportException`).
- Visual Studio extension migration tracked in metalama/Metalama.Vsx.Public#24; documentation in metalama/Metalama.Documentation#281.

## Issues Fixed
- Fixes #1701 - Disable telemetry per repository via `metalama.json` (context-scoped telemetry). Supersedes #1206.

— Claude for @gfraiteur


Metalama.Backstage/src/Metalama.Backstage/Repositories/RepositoryConfigurationService.cs
```diff
@@ -0,0 +1,203 @@
...
+        string key;
+
+        try
+        {
+            key = Path.GetFullPath( startingDirectory );
+        }
+        catch ( ArgumentException )
+        {
+            // An invalid path cannot resolve a repository; behave as if no file applies.
+            return RepositoryConfigurationResult.Empty;
```
gfraiteur: Should throw.

Metalama.Backstage/src/Metalama.Backstage/Repositories/RepositoryConfigurationService.cs
```diff
@@ -0,0 +1,203 @@
...
+
+        var result = this.Resolve( key );
+        this._cache[key] = (now, result);
+
+        return result;
+    }
+
+    private RepositoryConfigurationResult Resolve( string startingDirectory )
+    {
+        // Walk up from the starting directory, recording every directory that contains a metalama.json (nearest first)
```
gfraiteur: It is NOT nearestFirst! The only allowed file is the one in the root repo directory. Any other file is reported as a warning. So this method can be simplified.

Metalama.Backstage/src/Metalama.Backstage/Repositories/RepositoryConfigurationWarningKind.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2020-2025 SharpCrafters s.r.o. and contributors.
+// SharpCrafters s.r.o. licenses this file to you under either the MIT license or a proprietary license, depending on the repository from which it was obtained.
+// Refer to LICENSE.md in the repository root for complete details.
+
+namespace Metalama.Backstage.Repositories;
+
+internal enum RepositoryConfigurationWarningKind
```
gfraiteur, line 7: Is RepositoryConfigurationWarningKind still used?

Metalama.Backstage/src/Metalama.Backstage/Welcome/WelcomePageService.cs
```diff
@@ -0,0 +1,71 @@
...
+    private readonly BackstageBackgroundTasksService _backgroundTasksService;
+    private readonly bool _openWelcomePage;
+
+    internal WelcomePageService( IServiceProvider serviceProvider )
+    {
+        this._configurationManager = serviceProvider.GetRequiredBackstageService<IConfigurationManager>();
+        this._webLinks = serviceProvider.GetRequiredBackstageService<WebLinks>();
+        this._userInterfaceService = serviceProvider.GetRequiredBackstageService<IUserInterfaceService>();
+        this._backgroundTasksService = serviceProvider.GetRequiredBackstageService<BackstageBackgroundTasksService>();
+        this._openWelcomePage = serviceProvider.GetRequiredBackstageService<BackstageInitializationOptionsProvider>().Options.OpenWelcomePage;
```
gfraiteur: It's more efficient and simpler not to store _openWelcomePage  as a field and to call OnActivated only when openWelcomePage  is true.

gfraiteur:
Thanks — all three addressed in 75dfb7f.

1. **`EnsureActivated` doc ("sets the default reporting actions")** — corrected. The doc now states it writes only the device identifier, salts and upload timing, and explicitly that it does **not** set the per-channel reporting actions (doing so would clobber a prior `SetStatus` choice, since activation is lazy). Matches the implementation.

2. **`IsTelemetryEnabled` referencing `IsEnabled`** — the doc now resolves through `GetEffectiveReportingAction`, noting it (unlike `IsEnabled`) also covers the ASK-capable Exception/Performance scenarios, so callers are no longer pointed at an API that throws for them.

3. **Auto-upload writing `telemetry.json` on a never-activated process** — good catch; this was a real gap, not just docs. `BackstageServicesInitializer.Initialize` now gates the init-time `StartUpload()` on `IsActivated`, so a process that never activated telemetry neither advances `LastUploadTime` nor creates `telemetry.json` (and it has nothing queued to upload anyway). The lazy-activation guarantee now holds in code, rather than the doc over-promising. Behavior for already-activated machines is unchanged.

All 1365 Backstage tests remain green.

— Claude for @gfraiteur

Metalama.Backstage/src/Metalama.Backstage/Repositories/RepositoryConfigurationService.cs
gfraiteur: @-

Metalama.Backstage/src/Metalama.Backstage/Repositories/RepositoryConfigurationService.cs
gfraiteur: @-

Metalama.Backstage/src/Metalama.Backstage/Repositories/RepositoryConfigurationWarningKind.cs
gfraiteur: @-

Metalama.Backstage/src/Metalama.Backstage/Welcome/WelcomePageService.cs
gfraiteur: @-

PR #1726, opened 2026-07-16
Fix lost wakeup and double activation in AwaitableEvent (#1714)

### LINKED ISSUE #1714: Flaky test: RedisInvalidationTests.TestInvalidateDependencyIdenticalItemsAsync times out after 120s
## Summary

The Redis distributed-cache invalidation test `RedisInvalidationTests.TestInvalidateDependencyIdenticalItemsAsync` is flaky: it intermittently **times out after 120 000 ms** (its `[Fact(Timeout = 120000)]` limit) on CI.

## Occurrence

- **Build:** Metalama.Premium Public Build #27 - [build 329366](https://postsharp.teamcity.com/buildConfiguration/Metalama_Metalama20261_MetalamaPremium_PublicBuild/329366?expandedTest=build%3A%28id%3A329366%29%2Cid%3A2000000480)
- **Branch:** `release/2026.1`
- **Result:** 1 failed / 391 total
- **Test:** `Metalama.Patterns.Caching.Tests.Backends.Distributed.RedisInvalidationTests.TestInvalidateDependencyIdenticalItemsAsync`

## Failure

```
Test execution timed out after 120000 milliseconds
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

The captured stdout shows backends initializing, `SetItemAsync`/`InvalidateDependency` starting and `OnDependencyInvalidated` firing on some backends, but the test never completes - i.e. not all expected `ItemRemoved` notifications propagated through Redis pub/sub within the timeout, so the `await Task.WhenAll(itemRemovedEvents...)` (or final teardown) hangs.

## Source

`Metalama.Premium` -> `src/tests/Metalama.Patterns.Caching.Backends.UnitTests/Backends/Distributed/BaseDistributedCacheTests.cs`, method `TestInvalidateDependencyIdenticalItemsAsync` (lines ~58-118). The test relies on cross-backend Redis pub/sub notifications being delivered; when a notification is dropped/delayed the `ItemRemoved` `TaskCompletionSource` is never completed and the test hangs until the 120 s `Timeout`.

## Notes

This is a flaky/non-deterministic failure (Redis notification timing), not a deterministic regression - likely needs investigation of the Redis synchronizer notification delivery or the test's wait/retry strategy. No existing issue found for this test.

- Claude for @gfraiteur

gfraiteur:
## Summary

Fixes three concurrency defects in the internal `AwaitableEvent` primitive, which underpins `BackgroundTaskScheduler.WhenBackgroundTasksCompleted` and therefore caching-backend dispose.

- **Lost wakeup (the #1714 hang).** The manual-reset `Set`/wait handshake had no StoreLoad fence: `SetManualReset` published `SIGNALED` with a plain store then drained, while the waiter enqueued its operation and then read the signal with a plain volatile read. Both sides could miss each other, stranding an operation in `WAITING` **forever** even though the event was signalled. Fixed with `Interlocked.MemoryBarrier()` on both sides (the auto-reset path already got this for free from its `Interlocked.CompareExchange`).
- **Double activation.** `Activate()` fell through and re-scheduled the continuation when the operation was already `SUCCESS`, re-running an already-completed async state machine (`"attempt to transition a task to a final state when it had already completed"`). Fixed with a scheduled-once guard.
- **Wrong expected state.** The manual-reset async self-activation compared against `CREATED` although the operation is `WAITING` at that point, so it always failed and deferred activation to a drain that might never come. Now `WAITING`, matching the auto-reset branch.

Hardening of the synchronous path (found in review):

- `WaitOperationSync.Activate()` now takes part in the state protocol (CAS `CREATED/WAITING → SUCCESS`, `false` on loss) instead of unconditionally setting the event, so auto-reset can no longer consume a signal for a dead operation.
- Cancelling a blocked wait now withdraws the operation (`WAITING → TIMEOUT`) instead of abandoning it in the queue; auto-reset hands the signal back if it lost the race.
- Each wait gets a **private** event instead of a shared `[ThreadStatic]` one. `Activate()` cannot transition the state and set the event atomically, so a preempted activator could deliver its `Set()` after the waiter observed `SUCCESS`, returned and started an unrelated wait — spuriously waking it. Reachable via `BackgroundTaskScheduler.Dispose(CancellationToken)`.

Test infrastructure:

- `ConcurrencyTestingApi` was a global mutable static, incompatible with concurrent testing. Replaced by an internal `ITestSynchronizationProvider` resolved from an optional `IServiceProvider`. Sync points remain `[Conditional("DEBUG")]` and compile away in Release.

## Notes

- No public API change: everything touched or removed is `internal`, and `BackgroundTaskScheduler`'s constructor signature is unchanged.
- The same defects exist in the original PostSharp primitive this was ported from — reported as postsharp/PostSharp.Public#70.
- Validated end-to-end against `Metalama.Premium` (full caching-backends suite, real Redis: 272 passed) using a matching `topic/2026.1/1714-awaitableevent-race` branch.
- A load test reproduced the lost wakeup ~1 run in 3 (operation stuck `WAITING` with the signal set, 65s) and now passes 16/16, including under CPU saturation. It is `Skip`-marked so it does not run in CI.

- Also bumps PostSharp.Engineering to 2023.2.396 (separate commit); regenerating the TeamCity configuration with the newer SDK adds `--deployment default` to the publish step.

## Issues Fixed

Closes #1714 - Flaky test: `RedisInvalidationTests.TestInvalidateDependencyIdenticalItemsAsync` times out after 120s

— Claude for @gfraiteur


gfraiteur:
Thanks — addressed in 6376bfe.

**Sync points vs. non-DEBUG builds (`AwaitableEventRaceTests`, `AwaitableEventCancellationTests`) — fixed, and the problem was slightly worse than described.**

The tests are now `[SkippableFact]`/`[SkippableTheory]` and skip when the sync points are unavailable. Notably, the availability check is a **run-time probe**, not `#if DEBUG` in the test assembly, because the two do not agree: the test project defines `DEBUG` in the `LamaDebug` configuration, but `Metalama.Patterns.Caching.Backend` has no `DefineConstants` and so only gets `DEBUG` in `Debug`. A `#if DEBUG` guard would therefore still have failed under `LamaDebug`. The probe creates an event with a recording provider and checks whether a sync point is actually reached.

Verified both directions:
- `Debug`: the 3 tests **run** (0 skipped) — the guard does not over-skip.
- `Release`: the 3 tests **skip** in 57 ms instead of failing/hanging. Full suite: 459 passed / 0 failed (Debug), 387 passed / 0 failed (Release).

**Typo in the sync-point name — fixed.** `"Operation activated, move tot he next one."` → `"move to the next one."`. Agreed this matters more now that these strings are sync-point names that tests arm; confirmed no test armed the old spelling.

**`Debug.Assert` vs `CachingAssertionFailedException` — not changing, deliberately.**

`CachingAssertionFailedException.cs:13-15` is the right default, but `AwaitableEvent.cs` is code ported from `PostSharp.Patterns.Common` that uses `Debug.Assert` throughout (under a file-wide `#pragma warning disable`), and the new assertions sit directly alongside pre-existing ones in the same methods. Converting only the new ones would leave the file internally inconsistent, and converting all of them would change Release behaviour of a delicate lock-free primitive — out of scope for this fix. This exception is now written down in `CLAUDE.md` in this PR: prefer `Invariant.Assert`, or `CachingAssertionFailedException` where the engine isn't referenced, with `Debug` tolerated in already-ported code that uses it throughout.

Worth noting these asserts are load-bearing for correctness review rather than production: with the per-operation event introduced here, a spurious wakeup can no longer occur, so the `Debug.Assert(op.State == SUCCESS)` after a successful wait is now guaranteed to hold.

— Claude for @gfraiteur

PR #1735, opened 2026-07-19
Fix duplicate design-time validators, and clean up the cross-version manifest path (#1710)

### LINKED ISSUE #1710: Design-time InvalidCastException: ContractOptions merged across two TFM-specific compile-time assemblies
## Summary

Design-time analysis crashes with an `InvalidCastException` while merging hierarchical `ContractOptions`. Two different compile-time builds of `Metalama.Patterns.Contracts` -- one targeting **`.netframework4.7.2`** and one targeting **`.netstandard2.0`** -- are loaded into the **same** `MetalamaAssemblyLoadContext`, so their `ContractOptions` types are distinct and not castable to each other.

## Exception

`System.InvalidCastException` (wrapped in `AggregateException`):

> [A]Metalama.Patterns.Contracts.ContractOptions cannot be cast to [B]Metalama.Patterns.Contracts.ContractOptions.
> Type A originates from `ml!Metalama.Patterns.Contracts_6eaaffef90c904df` at `...\CompileTime\Metalama.Patterns.Contracts\.netframework4.7.2\6eaaffef90c904df\...`.
> Type B originates from `ml!Metalama.Patterns.Contracts_c37cfcd97bd8ae3c` at `...\CompileTime\Metalama.Patterns.Contracts\.netstandard2.0\c37cfcd97bd8ae3c\...`.
> Both in the context `"CompileTimeDomain" MetalamaAssemblyLoadContext #41`.

## Where it happens

Triggered while collecting **inherited aspects** across projects and evaluating `ContractBaseAttribute.IsInheritable`:

```
ContractOptions.IIncrementalObject.ApplyChanges            (netstandard2.0 build)
HierarchicalOptionsManager.OptionTypeNode.MergeOptions     (line 247)
HierarchicalOptionsManager.OptionTypeNode.GetOptions
HierarchicalOptionsManager.GetOptions
DeclarationEnhancements`1.GetOptions
ContractExtensions.GetContractOptions
ContractBaseAttribute.IsInheritable
AspectInstance..ctor
TransitivePipelineContributorSource.InheritedAspectSourceImpl -> ProcessAspectInstance
...
DesignTimeAspectPipeline.Execute -> TheDiagnosticAnalyzer.AnalyzeSemanticModel
```

`MergeOptions(baseOptions, overridingOptions, ...)` holds `ContractOptions` instances from two different TFM-specific compile-time builds; `ApplyChanges` then casts the `changes` argument to its own `ContractOptions` type and fails.

## Analysis / likely cause

The loaded-assemblies list confirms multiple compile-time variants of the same package coexist in one load context:
- `ml!Metalama.Patterns.Contracts_c37cfcd97bd8ae3c` (.netstandard2.0)
- `ml!Metalama.Patterns.Contracts_6eaaffef90c904df` (.netframework4.7.2)
- `ml!Metalama.Patterns.Contracts_17e41778d04c760a` (a third hash)

This looks like cross-project aspect **inheritance** merging `HierarchicalOptions` whose runtime `Type` comes from different target-framework builds of the same compile-time assembly. When a project inherits a `Contract*` aspect from a referenced project compiled against a different TFM variant of the Contracts compile-time assembly, the option merge mixes the two `ContractOptions` types.

## Environment

- Metalama.DesignTime `2026.1.19.1029-local-GaelFraiteur-debug`
- Host: JetBrains Rider (`JetBrains.Roslyn.Worker`), .NET 10, Windows `10.0.26200`
- Scenario: design-time (`DesignTimeAspectPipeline` -> `TheDiagnosticAnalyzer.AnalyzeSemanticModel`)

## Repro hypothesis

A solution with at least two projects targeting **different TFMs** (e.g. `net472` and a `netstandard2.0`/modern target), both using `Metalama.Patterns.Contracts`, where contract aspects with hierarchical options are **inherited** across the project boundary.

## Crash report metadata

- InvariantHash: `48fc2037f518175d2c32e7fb2a7e7451`
- ClientId: `b21efd28f7f1043`
- Time: `2026-06-26T10:48:10Z`
- Category: Exception (telemetry crash report)

-- Claude for @gfraiteur

gfraiteur:
Two related changes to the transitive-manifest path, both consequences of #1710: a bug fix with tests, then a cleanup of dead code and misleading names in the same area.

## Part 1 — the fix (`bbcc30d`)

`CreateTransitiveManifest` now takes an `includeValidators` flag, so design-time validators travel in the transitive manifest only for consumers that have no other channel to receive them.

| Manifest | Flag | Consumer |
|---|---|---|
| `SerializedTransitiveAspectManifest` | `false` | same version |
| `LiveTransitiveAspectManifest` (#1710 fast path) | `false` | same version |
| `SerializeTransitiveAspectManifestForOtherVersion` | `true` | different version |

Design-time validators reach a consuming project through two channels, and exactly one of them must carry them for any given consumer:

1. the **design-time channel**, `DesignTimeProjectVersion.ReferencedExtensions`, which reads the producer's live `DesignTimeAspectPipelineResult` and deduplicates diamond-shaped reference graphs; and
2. the **manifest channel**, which walks each direct reference's transitive manifest and is *not* deduplicated.

Before #1710 the same-version path read the live result, whose `ITransitiveAspectsManifest.Extensions` filtered validators out, so only channel 1 carried them. #1710 rerouted that path onto `CreateTransitiveManifest` — first via deserialization in `3b3807d74e`, then via the `LiveTransitiveAspectManifest` fast path in `19124afc98` — which is built for the cross-version consumer and keeps validators. Both channels then fired.

`HasTransitiveAspectManifestContent` deliberately stays computed from the unfiltered collection: were it to go `false` for a validators-only project, the reference would carry no live manifest either, and channel 1 would lose the validators along with it.

### Impact

Caught by the `Metalama.Extensions.Validation` cross-project design-time tests in Metalama.Premium, which reported every cross-project reference diagnostic twice — six times across a diamond, where the undeduplicated manifest channel compounds:

| Test | Expected | Was |
|---|---|---|
| `ValidatorTests.CrossProject` | 2 | 4 |
| `ValidatorTests.CrossProject_Diamond` | 2 | 12 |
| `ValidatorTests.CrossProject_ModificationInAttributeConstruction` | 1 | 2 |

## Part 2 — the cleanup (`1d28501`, `e586f9d`)

**What "cross-version" means:** two projects in the same solution, one referencing the other, each built against a **different version of Metalama**. Both versions are loaded in the **same process** and reach each other through the version-neutral `Metalama.Framework.DesignTime.Contracts` assembly. Their `Metalama.Framework.Engine` types have distinct identities, so no manifest object can be passed across — the manifest is serialized by the producing version and deserialized into the consuming version's object model. Same assembly-identity problem #1710 addressed; not a transport problem.

- **Removes a deserialization whose result has had no reader since #1710.** The cross-version branch of `GetDesignTimeProjectVersionAsync` deserialized the referenced project's manifest onto the `DesignTimeProjectReference`. Its two readers, `ReferencedExtensions` and `TryGetReusableTransitiveAspectsManifest`, both narrow to the concrete `DesignTimeAspectPipelineResult`, which a deserialized `TransitiveAspectsManifest` never is, so both silently ignored it. Its only remaining effect was satisfying the reference's both-or-neither invariant. It was load-bearing before #1710, when `DesignTimeProjectVersion.GetTransitiveAspectsManifest` returned it and `TransitivePipelineContributorSource` consumed it; #1710 routed the consumer through the serialized bytes and left this behind.
- **Types `TransitiveAspectsManifest` as the concrete result**, so "a cross-version reference has no live result" is stated by the type instead of by two readers narrowing and discarding whatever does not match. The invariant relaxes accordingly: a live result implies a serialized manifest, not the converse.
- **Renames `SerializeTransitiveAspectManifestForRpc` → `...ForOtherVersion`.** Nothing on that path is RPC. The genuine RPC — user process ↔ analysis process, `Metalama.Framework.DesignTime.Rpc`, consumed by Metalama.Vsx — is untouched.
- **Rewords the docs across the area**: the distinction is same-version vs different-version, not in-process vs out-of-process. Both are in-process.
- **`e586f9d` drops `DesignTimeAspectPipelineResult : ITransitiveAspectsManifest`**, unused once the property is typed concretely. Also deletes the "still used for PE references" comment, stale since #1710: a PE reference reads its manifest from the assembly metadata resource written by the compile-time pipeline and never touches a `DesignTimeAspectPipelineResult`. Kept as its own commit so it can be reverted alone if the interface is wanted for a reason not visible in the code.

## Behavioral change

The same-version transitive manifest no longer carries design-time validators, restoring pre-#1710 behavior. The cross-version manifest is unchanged.

## Testing

`Metalama.Framework.Tests.UnitTests` design-time suite: 437 passed, 5 skipped, 0 failed. Premium's `Metalama.Extensions.Validation.UnitTests` (13 tests, including the three above) pass against a locally built Metalama carrying this branch.

New `TransitiveManifestValidatorChannelTests` covers both same-version manifests, and was confirmed to fail without Part 1. The concrete `IDesignTimeValidatorExtension` implementations live in Metalama.Premium, so this repo cannot reproduce the failure end to end; the tests drive a fake contributor through the real `Update` → extension-collection routing. `LiveManifest_DropsValidators_ButKeepsOtherExtensions` also asserts the validator is still present on channel 1, so the fix cannot be trivially satisfied by dropping validators altogether.

### Gaps worth knowing

- **The cross-version path has no automated coverage** — nothing in the repo touches `ITransitiveCompilationService`, `GetTransitiveAspectManifestAsync`, or the entry-point consumer, which is why a vestigial deserialization sat there unnoticed. Exercising it needs two Metalama versions in one process, so Part 2's change to that branch is reviewed, not tested. Worth sanity-checking a two-version solution in VS before merging.
- **The `includeValidators: true` direction is untested.** Asserting on it requires a compile-time serializer for the fake extension. Behavior there is unchanged, but note the failure mode is silent: too few validators means diagnostics simply vanish, where too many were loudly duplicated.
- Part 2 leaves `GetOrCreatePipelineAsync`/`GetConfigurationAsync` in the cross-version branch. With the deserialization gone their only role is a gate — the referenced project must be compilable by the current version, and failing there beats a later consumer-side binding error. Documented rather than removed, since I could not verify they are inert without the coverage above.

## Issues Fixed

None — this is a regression from the (closed) #1710 work, referenced above for context.

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipeline.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipeline.cs
@@ -608,13 +608,11 @@ public sealed partial class DesignTimeAspectPipeline : BaseDesignTimeAspectPipel
                     }
                     else
                     {
-                        // The referenced project must be compilable by the *current* Metalama version, because the
-                        // consuming project binds the manifest against its own compile-time closure, which contains
-                        // this reference. Obtaining the configuration is what gates that, so a failure here aborts
-                        // rather than surfacing later as a binding error attributed to the consumer.
-                        //
-                        // Note that we deliberately do not deserialize the manifest here. See the comment on the
-                        // DesignTimeProjectReference constructed below.
+                        // Require the referenced project to have a valid configuration for the current Metalama
+                        // version before accepting its manifest. The consuming project resolves the manifest's
+                        // run-time type names through its own compile-time closure, which contains this reference,
+                        // so a reference the current version cannot configure would fail later as a binding error
+                        // reported against the consumer. Failing here reports it against the reference instead.
 
                         var pipelineResult = await this._pipelineFactory.GetOrCreatePipelineAsync( reference, cancellationToken );
 
@@ -638,12 +636,8 @@ public sealed partial class DesignTimeAspectPipeline : BaseDesignTimeAspectPipel
                                 $"Cannot get configuration: {configuration.DebugReason}" );
                         }
 
-                        // Carry only the serialized manifest. The consumer deserializes it into its own compile-time
-                        // copy (issue #1710), so there is nothing for a live manifest to do here: a live manifest is
-                        // read exclusively by DesignTimeProjectVersion.ReferencedExtensions and
-                        // TryGetReusableTransitiveAspectsManifest, both of which need the producer's concrete
-                        // DesignTimeAspectPipelineResult, and a cross-version producer cannot supply one — its
-                        // result is an object of the *other* version's Metalama.Framework.Engine.
+                        // A cross-version reference carries the serialized manifest alone: the producer's live
+                        // result is an object of the other version's Metalama.Framework.Engine and cannot cross.
                         compilationReferences.Add(
                             new DesignTimeProjectReference(
                                 reference.ProjectKey,
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipeline.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipeline.cs
@@ -608,8 +608,13 @@ public sealed partial class DesignTimeAspectPipeline : BaseDesignTimeAspectPipel
                     }
                     else
                     {
-                        // To deserialize the manifest, we need a service provider with the CompileTimeProject of the referenced project, compiled
-                        // for the current Metalama version.
+                        // The referenced project must be compilable by the *current* Metalama version, because the
+                        // consuming project binds the manifest against its own compile-time closure, which contains
+                        // this reference. Obtaining the configuration is what gates that, so a failure here aborts
+                        // rather than surfacing later as a binding error attributed to the consumer.
+                        //
+                        // Note that we deliberately do not deserialize the manifest here. See the comment on the
+                        // DesignTimeProjectReference constructed below.
 
                         var pipelineResult = await this._pipelineFactory.GetOrCreatePipelineAsync( reference, cancellationToken );
 
@@ -633,18 +638,16 @@ public sealed partial class DesignTimeAspectPipeline : BaseDesignTimeAspectPipel
                                 $"Cannot get configuration: {configuration.DebugReason}" );
                         }
 
-                        var manifest = TransitiveAspectsManifest.Deserialize(
-                            new MemoryStream( result.Manifest! ),
-                            configuration.Value.ServiceProvider,
-                            reference.Compilation.AssemblyName );
-
-                        // Also carry the raw serialized manifest (produced by the referenced project) so the consumer
-                        // can deserialize inherited aspects and options into its own compile-time copy (issue #1710).
+                        // Carry only the serialized manifest. The consumer deserializes it into its own compile-time
+                        // copy (issue #1710), so there is nothing for a live manifest to do here: a live manifest is
+                        // read exclusively by DesignTimeProjectVersion.ReferencedExtensions and
+                        // TryGetReusableTransitiveAspectsManifest, both of which need the producer's concrete
+                        // DesignTimeAspectPipelineResult, and a cross-version producer cannot supply one — its
+                        // result is an object of the *other* version's Metalama.Framework.Engine.
                         compilationReferences.Add(
                             new DesignTimeProjectReference(
                                 reference.ProjectKey,
-                                manifest,
-                                SerializedTransitiveAspectManifest.Create( result.Manifest!.ToImmutableArray() ) ) );
+                                serializedTransitiveAspectManifest: SerializedTransitiveAspectManifest.Create( result.Manifest!.ToImmutableArray() ) ) );
 
                         if ( result.IsPipelinePaused )
                         {
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipeline.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipeline.cs
@@ -554,29 +554,28 @@ public sealed partial class DesignTimeAspectPipeline : BaseDesignTimeAspectPipel
                             $"The pipeline for the reference '{reference.ProjectKey}' failed: {referenceResult.DebugReason}" );
                     }
 
-                    // Serialize the referenced project's transitive manifest. The serialized form is
-                    // compilation-neutral by definition: compile-time types are always written as their run-time
-                    // names. We must use the *referenced* project's service provider because only its closure can
-                    // name (resolve) that project's own compile-time copy of a shared, possibly multi-targeted
-                    // assembly; the consuming project's provider would fail to serialize types bound to a copy that
-                    // is not in its closure. The consumer then deserializes these bytes with its own service
-                    // provider, which binds the run-time names to the consumer's own compile-time copy (issue #1710).
-                    // The successful result always carries the reference pipeline's configuration (even when the
-                    // pipeline is paused), so the serialized manifest is always available alongside the live one.
-                    var serializedManifest = referenceResult.Value.Result.SerializedTransitiveAspectManifest;
-
-                    // When the reference exports nothing to inherit, its serialized manifest is `default`. We then
-                    // carry neither the live nor the serialized manifest, keeping the both-or-neither invariant of
-                    // DesignTimeProjectReference and letting the consumer skip deserialization entirely. Dropping the
-                    // live manifest is safe here: an empty manifest also has an empty Extensions collection, so
-                    // DesignTimeProjectVersion.ReferencedExtensions loses nothing.
+                    // Carry the reference's manifest only when it exports something to inherit, which is what makes
+                    // this the common fast path: a Metalama project exporting no inheritable aspect, option,
+                    // annotation or validator carries neither the live nor the serialized form, so nothing is
+                    // serialized here and nothing is deserialized or merged on the consumer's side. Dropping the live
+                    // manifest along with it is safe, because a project with nothing to inherit also has an empty
+                    // Extensions collection, so DesignTimeProjectVersion.ReferencedExtensions loses nothing.
+                    //
+                    // The serialized form is compilation-neutral by definition: compile-time types are always written
+                    // as their run-time names. It is produced by the *referenced* project's service provider because
+                    // only its closure can name (resolve) that project's own compile-time copy of a shared, possibly
+                    // multi-targeted assembly; the consuming project's provider would fail to serialize types bound
+                    // to a copy that is not in its closure. The consumer then deserializes those bytes with its own
+                    // provider, which binds the run-time names to its own compile-time copy (issue #1710). A
+                    // successful result always carries the reference pipeline's configuration, even when the pipeline
+                    // is paused, so the serialized manifest is always available alongside the live one.
                     compilationReferences.Add(
-                        serializedManifest.IsDefaultOrEmpty
-                            ? new DesignTimeProjectReference( referenceResult.Value.ProjectVersion.ProjectKey )
-                            : new DesignTimeProjectReference(
+                        referenceResult.Value.Result.HasTransitiveAspectManifestContent
+                            ? new DesignTimeProjectReference(
                                 referenceResult.Value.ProjectVersion.ProjectKey,
                                 referenceResult.Value.Result,
-                                serializedManifest ) );
+                                referenceResult.Value.Result.SerializedTransitiveAspectManifestWithoutValidators )
+                            : new DesignTimeProjectReference( referenceResult.Value.ProjectVersion.ProjectKey ) );
 
                     if ( referenceResult.Value.Status == DesignTimeAspectPipelineStatus.Paused )
                     {
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipeline.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipeline.cs
@@ -640,7 +640,7 @@ public sealed partial class DesignTimeAspectPipeline : BaseDesignTimeAspectPipel
                         compilationReferences.Add(
                             new DesignTimeProjectReference(
                                 reference.ProjectKey,
-                                serializedTransitiveAspectManifest: SerializedTransitiveAspectManifest.Create( result.Manifest!.ToImmutableArray() ) ) );
+                                serializedTransitiveAspectManifest: GetCrossVersionManifest( result ) ) );
 
                         if ( result.IsPipelinePaused )
                         {
@@ -660,6 +660,29 @@ public sealed partial class DesignTimeAspectPipeline : BaseDesignTimeAspectPipel
         }
 
         return new DesignTimeProjectVersion( compilationVersion, compilationReferences, pipelineStatus );
+    }
+
+    /// <summary>
+    /// Wraps the manifest received from a cross-version reference, taking the content hash from the producer when it
+    /// reports one.
+    /// </summary>
+    /// <remarks>
+    /// The hash is what <c>TransitiveManifestDeserializationCache</c> keys on downstream, so supplying it here is
+    /// what lets an unchanged reference skip being deserialized again. A producer older than
+    /// <see cref="ITransitiveCompilationResult2"/> reports none, and the bytes are hashed on first use as before.
+    /// That is the only case in which the ordering rule, a referencing project never being older than the project it
+    /// references, puts the older Metalama on the producing side.
+    /// </remarks>
+    private static SerializedTransitiveAspectManifest? GetCrossVersionManifest( ITransitiveCompilationResult result )
+    {
+        if ( result.Manifest is not { Length: > 0 } bytes )
+        {
+            return null;
+        }
+
+        return SerializedTransitiveAspectManifest.Create(
+            bytes.ToImmutableArray(),
+            result is ITransitiveCompilationResult2 { ManifestHash: var hash } ? hash : null );
     }
 
     public ValueTask<FallibleResultWithDiagnostics<DesignTimeAspectPipelineResultAndState>> ExecuteAsync(
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineResult.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineResult.cs
@@ -749,16 +749,18 @@ public sealed partial class DesignTimeAspectPipelineResult
     /// </summary>
     /// <remarks>
     /// <para>
-    /// Compressed, unlike <see cref="SerializedTransitiveAspectManifest"/>, and not because these bytes are stored
-    /// anywhere: they are consumed in this same process and discarded, exactly like the same-version ones. The
-    /// reason is compatibility. These bytes are read by the <em>other</em> version's deserializer, and the
-    /// uncompressed format is recognized by peeking a marker byte that was only introduced in 2026.1 (issue #1710).
-    /// A peer older than that has no such peek: it would treat the marker as the first byte of a DEFLATE stream and
-    /// fail. The compressed format is the one every peer understands, so it is what we emit outwards.
+    /// Compressed, unlike <see cref="SerializedTransitiveAspectManifest"/>, following the rule that the compressed
+    /// form is the <em>interchange</em> format, read by a Metalama build that may not be this one, while the
+    /// uncompressed form is the <em>private</em> format, read by this same build in this same process and then
+    /// discarded. The other user of the interchange format is <c>TransitiveAspectsManifest.ToResource</c>, embedded
+    /// in the PE binary and read later by whichever version compiles against that assembly.
     /// </para>
     /// <para>
-    /// This is therefore the one place that compresses without storing. Everything stored compresses too
-    /// (<c>TransitiveAspectsManifest.ToResource</c>, embedded in the PE binary), but the converse does not hold.
+    /// Concretely, these bytes are read by the other version's deserializer, and the uncompressed form is recognized
+    /// by peeking a marker byte introduced in 2026.1 (issue #1710). A peer older than that has no such peek: it
+    /// would treat the marker as the first byte of a DEFLATE stream and fail. Once pre-2026.1 interop is dropped
+    /// this can stop compressing, tracked by issue #1737. The path is cold (mixed-version solutions only, once per
+    /// cross-version reference per pipeline run), so there is nothing to gain by removing it sooner.
     /// </para>
     /// </remarks>
     internal byte[] SerializeTransitiveAspectManifestForOtherVersion()
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineResult.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineResult.cs
@@ -687,19 +687,21 @@ public sealed partial class DesignTimeAspectPipelineResult
     /// </item>
     /// <item>
     /// <c>true</c> for a consumer built against a different version of Metalama. Such a reference carries no live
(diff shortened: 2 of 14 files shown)


Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineResult.cs
```diff
@@ -718,23 +722,34 @@ private TransitiveAspectsManifest CreateTransitiveManifest()
...
     /// </summary>
-    internal byte[] SerializeTransitiveAspectManifestForRpc()
-        => this.CreateTransitiveManifest().ToBytes( this.Configuration.AssertNotNull().ServiceProvider, compress: true );
+    /// <remarks>
+    /// Compressed, unlike <see cref="SerializedTransitiveAspectManifest"/>: the consumer may be an older version
+    /// that only understands the legacy compressed format.
+    /// </remarks>
+    internal byte[] SerializeTransitiveAspectManifestForOtherVersion()
+        => this.CreateTransitiveManifest( includeValidators: true )
+            .ToBytes( this.Configuration.AssertNotNull().ServiceProvider, compress: true );
```
gfraiteur: We should not compress here (we only compress when the manifest is stored in the assembly as a resource).

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipeline.cs
```diff
@@ -608,8 +608,13 @@ internal async ValueTask<FallibleResultWithDiagnostics<DesignTimeProjectVersion>
                     }
                     else
                     {
-                        // To deserialize the manifest, we need a service provider with the CompileTimeProject of the referenced project, compiled
-                        // for the current Metalama version.
+                        // The referenced project must be compilable by the *current* Metalama version, because the
+                        // consuming project binds the manifest against its own compile-time closure, which contains
+                        // this reference. Obtaining the configuration is what gates that, so a failure here aborts
```
gfraiteur: I don't understand this comment. Restate.

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipeline.cs
```diff
@@ -633,18 +638,16 @@ internal async ValueTask<FallibleResultWithDiagnostics<DesignTimeProjectVersion>
...
-                            new MemoryStream( result.Manifest! ),
-                            configuration.Value.ServiceProvider,
-                            reference.Compilation.AssemblyName );
-
-                        // Also carry the raw serialized manifest (produced by the referenced project) so the consumer
-                        // can deserialize inherited aspects and options into its own compile-time copy (issue #1710).
+                        // Carry only the serialized manifest. The consumer deserializes it into its own compile-time
+                        // copy (issue #1710), so there is nothing for a live manifest to do here: a live manifest is
+                        // read exclusively by DesignTimeProjectVersion.ReferencedExtensions and
+                        // TryGetReusableTransitiveAspectsManifest, both of which need the producer's concrete
```
gfraiteur: Remove the comment about the live manifest which does not make sense to anyone reading the target state of the code instead of the diff of the pr.

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeProjectVersion.cs
```diff
@@ -21,8 +21,13 @@ internal sealed class DesignTimeProjectVersion : ITransitiveAspectManifestProvid
...
     public IProjectVersion ProjectVersion { get; }
 
+    /// <summary>
+    /// Gets the design-time extension collections of the referenced projects. Cross-version references contribute
+    /// nothing here, because they carry no live result; their contribution travels in the serialized manifest
+    /// instead, which the engine deserializes into this project's own compile-time copy.
+    /// </summary>
     public IEnumerable<DesignTimeAspectPipelineResultExtensionCollection> ReferencedExtensions
-        => this._references.Values.Select( r => (r.TransitiveAspectsManifest as DesignTimeAspectPipelineResult)?.Extensions ).WhereNotNull();
+        => this._references.Values.Select( r => r.TransitiveAspectsManifest?.Extensions ).WhereNotNull();
```
gfraiteur, line 44: This seems like a smell to me, I wonder if we have a cross-version or cross-tfm test for this scenario. It seems we should sometimes rely on the serialized manifest.

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipeline.cs
gfraiteur: Restated in `1405c2f`. The comment now says what the configuration is required *for* rather than narrating what the code no longer does:

> Require the referenced project to have a valid configuration for the current Metalama version before accepting its manifest. The consuming project resolves the manifest's run-time type names through its own compile-time closure, which contains this reference, so a reference the current version cannot configure would fail later as a binding error reported against the consumer. Failing here reports it against the reference instead.

Worth flagging that this is my reading of why the calls are there, not something I could verify: with the deserialization gone, `GetOrCreatePipelineAsync` / `GetConfigurationAsync` have no other use here. If they are in fact inert, they can go too and the comment with them.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipeline.cs
gfraiteur: Removed in `1405c2f`. You were right on the principle, not just this instance: the comment explained why the code *doesn't* do something, which only parses if you are reading the diff. What replaces it states what the reference carries:

```csharp
// A cross-version reference carries the serialized manifest alone: the producer's live
// result is an object of the other version's Metalama.Framework.Engine and cannot cross.
```

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeProjectVersion.cs
gfraiteur: Documented rather than restructured, in `1405c2f`. Taking the two halves separately.

**Is it a hole?** No, but nothing at the call site said so, which is what made it read as one. A cross-version reference contributes nothing here because it carries no live result, and its contribution travels through the serialized manifest instead, which `TransitivePipelineContributorSource` deserializes into this project's compile-time copy. That is exactly why `CreateTransitiveManifest` takes `includeValidators`: whichever channel is unavailable to a consumer, precisely one carries the validators. The remarks now say this and cross-reference both directions.

The split is also not merely an optimization, which is the part worth pinning: this channel deduplicates diamond-shaped reference graphs and the manifest channel does not, so routing a same-version reference through both delivers its validators more than once. That is the bug this PR fixes.

**Do we have a test?** Cross-TFM yes, cross-version no.

- `CrossTfmInheritedOptionsTests` covers same version, differing compile-time copies. That is the case where the live manifest cannot be reused and the serialized one is taken instead, so the fallback is exercised.
- Cross-version has no coverage anywhere in the repo. Nothing touches `ITransitiveCompilationService`, `GetTransitiveAspectManifestAsync`, or the entry-point consumer. Exercising it needs two Metalama versions in one process.

That gap is why the vestigial deserialization removed in `1d28501` survived unnoticed since #1710, and it means the cross-version branch of this PR is reviewed rather than tested. If you want that covered I can look at what a harness would take, but it did not look like something to fold into this PR.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineResult.cs
gfraiteur: I did not make this change, because I believe it would break pre-2026.1 peers. Filed as #1737 to do once they are out of support, and reframed the `<remarks>` in `8efd47a`. Reasoning, since it hinges on a direction that is easy to get backwards.

**The sniff is one-directional.** `Deserialize` picks a format by peeking a marker byte (`0xFF`, which cannot start a DEFLATE stream). A new reader therefore handles old bytes. An old reader has no peek at all: on `develop/2025.1`, `Deserialize` is unconditionally

```csharp
using var deflate = new DeflateStream( stream, CompressionMode.Decompress );
```

`git branch --contains 4c62714248` puts the marker on 2026.1 branches only.

**We are the producer here, not the consumer.** `SerializeTransitiveAspectManifestForOtherVersion` runs inside our `ITransitiveCompilationService`, which is invoked when a project of *another* version references one of ours. That peer deserializes with its own version's code. So in a solution where a 2025.1 project references a 2026.1 project, our bytes are read by 2025.1, and emitting uncompressed would feed `0xFF...` to `DeflateStream` and throw. The reverse direction is safe, because 2025.1 always compresses and our reader sniffs.

**Contract negotiation does not keep the pairing apart.** `DesignTimeEntryPointManager.Consumer.ValidateContractVersions` requires an exact revision match on contract `"1.0"`, but `ContractVersion_1_0` has been `3` since `8b2ad17848`, well before this release line, so both versions advertise 3 and connect.

**On the rule itself**, I think the principle behind your comment is slightly broader than storage, and stated that way it covers all three call sites without exception:

> Compressed is the interchange format, read by a Metalama build that may not be this one. Uncompressed is the private format, read by this same build, in this same process, immediately.

`ToResource` (PE resource) and this method are interchange; `SerializedTransitiveAspectManifest` is private and genuinely never crosses a version boundary. Storage is then a consequence of the rule rather than the rule itself. The `<remarks>` now say this.

The cost of keeping it is small: mixed-version solutions only, once per cross-version reference per pipeline run. It is not the hot same-version path that #1710 rightly made uncompressed.

Happy to flip it to `compress: false` in this PR instead if you consider pre-2026.1 cross-version interop out of scope. It is a one-word change.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineResult.cs
```diff
@@ -701,15 +702,18 @@ private TransitiveAspectsManifest CreateTransitiveManifest()
             // `false` would risk missing required WithInitialize wrapping.
```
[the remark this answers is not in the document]
gfraiteur: Correct, and self-inflicted: `e586f9d` deleted the `ITransitiveAspectsManifest.ContainsInitializableTypes` member that this "as above" pointed at, so the rationale for the flag was lost along with it. Fixed in `1405c2f` by restoring the reasoning inline rather than leaving a dangling reference.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeProjectReference.cs
```diff
@@ -13,18 +13,31 @@ namespace Metalama.Framework.DesignTime.Pipeline;
 /// </summary>
```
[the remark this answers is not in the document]
gfraiteur: Correct, fixed in `1405c2f`. The summary now describes the type as it now is: a `ProjectKey`, the serialized manifest when the reference exports anything to inherit, and, for a same-version reference only, that project's live pipeline result as well.

— Claude for @gfraiteur

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/Pipeline/TransitiveManifestValidatorChannelTests.cs
```diff
@@ -0,0 +1,200 @@
...
+/// </list>
+/// A same-version, in-process consumer always has channel 1 (its reference carries the live result; see the
+/// both-or-neither invariant on <c>DesignTimeProjectReference</c>), so the in-process manifests must not carry
+/// validators. A cross-version (RPC) consumer has only channel 2, so its manifest must.
+/// </summary>
+/// <remarks>
+/// This regressed once already (issue #1710): the design-time reference path was rerouted from the live result,
+/// whose <c>ITransitiveAspectsManifest.Extensions</c> filters validators out, onto the manifest built for RPC,
+/// which keeps them. Both channels then fired, and a downstream consumer reported every cross-project reference
+/// diagnostic twice — six times across a diamond, where the undeduplicated manifest channel compounds.
```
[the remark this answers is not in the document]
gfraiteur: Correct, and it is a documented rule in this repo's `CLAUDE.md` that I had not read. Fixed in `1405c2f`, along with every other em dash this branch introduced.

— Claude for @gfraiteur

Metalama.Framework/src/tests/Metalama.Framework.Tests.UnitTests/DesignTime/Pipeline/TransitiveManifestValidatorChannelTests.cs
```diff
@@ -0,0 +1,200 @@
...
+        Assert.Empty( deserialized.Extensions );
+    }
+
+    /// <summary>
+    /// Stands in for a design-time validator, which in production comes from Metalama.Extensions.Validation and so
+    /// is not available here. One class covers both roles, with <c>isValidator</c> the only difference: routing
+    /// keys off <see cref="ContributorKind.IsDesignTimeValidator"/> (in conjunction with the interface, which both
+    /// roles implement), so varying just the flag isolates the behaviour under test.
+    /// </summary>
+    private sealed class FakeContributor : ITransitivePipelineContributor, IDesignTimeValidatorExtension
```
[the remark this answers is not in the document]
gfraiteur: The hazard is real and is fixed in `1405c2f`, but by a different route than suggested.

`FakeContributor` now takes a real `SymbolDictionaryKey`, built with `CreatePersistentKey` from a type in the test compilation. The fixture never looks a validator up by symbol, but the key indexes the validator dictionary that `Update` populates, so an identity-less one is a hazard the moment anything compares keys.

I did not take the suggestion to implement only `IDesignTimePipelineResultExtension`. The routing in `DesignTimeAspectPipelineResultExtensionCollection.Builder.Add` keys off `ContributorKind.IsDesignTimeValidator` **and** `extension is IDesignTimeValidatorExtension`:

```csharp
if ( extension.ContributorKind.IsDesignTimeValidator && extension is IDesignTimeValidatorExtension validator )
```

Dropping the interface would stop the fake being routed as a validator at all, so the test would no longer exercise the path it exists to cover. A real validator implements both, and the fake should too; only the flag varies between the two roles, which is what isolates the behaviour under test.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineResult.cs
```diff
@@ -692,24 +673,49 @@ internal bool HasTransitiveAspectManifestContent
            || !this.Annotations.IsEmpty
            || !this.Extensions.IsEmpty;
 
-    private TransitiveAspectsManifest CreateTransitiveManifest()
+    /// <summary>
+    /// Creates the transitive manifest. <paramref name="includeValidators"/> selects whether the design-time
+    /// validators travel in the manifest, and must match whether the consumer has the other channel that carries
```
gfraiteur:
```suggestion
    /// validators are included in the manifest, and must match whether the consumer has the other channel that carries
```

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineResult.cs
```diff
@@ -692,24 +673,49 @@ internal bool HasTransitiveAspectManifestContent
...
                 this.InheritableOptions,
                 this.Annotations,
                 containsInitializableTypes: true );
 
     /// <summary>
-    /// Gets the transitive manifest serialized for in-process consumption by the current-version design-time
-    /// pipeline. It is serialized uncompressed: the bytes are produced and consumed in the same process and
-    /// discarded, so compression would be pure overhead. Returns <c>default</c> when there is nothing to inherit
+    /// Gets the transitive manifest serialized for a consumer built against the <em>same</em> version of Metalama.
+    /// It exists because the consumer must bind the manifest to its own compile-time copy of each type, which a
```
gfraiteur: Here I don't understand the comment. Serialization are cross-version, cross-tfm, and cross-process. So why do you write it must be same version? It's confusing. Is it because of the backward-compatibility tning where we cannot compress the stream?

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineResult.cs
gfraiteur: Applied in `b44991b`.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineResult.cs
gfraiteur: You are right, and the wording was doing real damage: "for a consumer built against the same version" reads as a claim about serialization in general, when serialization is precisely what bridges versions elsewhere. Rewritten in `b44991b`.

**To your direct question:** partly, but the compression is the smaller half. Two separate facts were run together in one sentence.

*Why this member exists* has nothing to do with versions. The consuming project must bind the manifest to **its own compile-time copy** of each type, and the round-trip through run-time type names is what rebinds them (#1710). Two projects on the *same* Metalama version can hold different compile-time copies, typically when a shared assembly is multi-targeted. So what this crosses is a compile-time copy: it is the cross-TFM case, not a cross-version one. When the copies do match, the consumer skips it and reuses `LiveTransitiveAspectManifest`.

*Why it may stay uncompressed* is the back-compat point you guessed. Its only reader is this build, which understands the marker. That is also the only reason the cross-version form still compresses, so the two are not inconsistent, which was the thing the old text failed to make clear.

**On the framing**, a manifest is serialized in exactly three places, and they differ only in which boundary the bytes cross:

| Member | Crosses |
| --- | --- |
| `SerializeTransitiveAspectManifestForOtherVersion` | a Metalama version |
| `TransitiveAspectsManifest.ToResource` | a process, and arbitrary time (embedded in the PE binary, read by whichever version later compiles against it) |
| `SerializedTransitiveAspectManifest` | neither: same build, same process, discarded immediately |

So cross-version and cross-process are both real, as you say, just not on this member. The remarks now lead with that table's content rather than asserting "same version" without saying what the alternative would be.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineResult.cs
gfraiteur: Your invariant settles it. Done in `77a989d0`: `compress: false`, and #1737 closed as fixed here rather than deferred.

The reasoning I was missing is that the ordering makes the dangerous direction unreachable. We are the *referenced* project when this runs, so the reader is the *referencing* one, and if its version is never older than ours then every possible reader already has the marker peek. The pre-2026.1 reader I was worried about cannot be on this side of the call.

Worth recording that this bounds what we **write**, not what we **read**. The converse case is ordinary: we may reference an older project, whose manifest arrives in the legacy compressed form. `Deserialize` therefore has to keep accepting both, even though `ToResource` is now the only writer that compresses (for the unrelated reason of keeping the PE binary small). The remarks on the method say this, so nobody later removes the sniff as dead code.

One caveat, since it is the sole thing holding this up: I could not find the invariant enforced anywhere. `SerializationProtocol.CurrentVersion` and `LastSupportedVersion` are both `2` and have been across 2025.1 and 2026.1, so the protocol check would accept a 2026.1-written stream in a 2025.1 reader rather than rejecting it, and I found no diagnostic on a reference to a newer Metalama. So if a solution ever does violate the ordering, the symptom moves from "works" to an `InvalidDataException` out of `DeflateStream`, with nothing pointing at the version skew. Taking your word that the configuration is not supported; flagging it in case it is worth a diagnostic separately.

Test coverage adjusted rather than dropped: `CompressedAndUncompressedFormats_RoundTripToSameContent` now asserts both design-time manifests carry the marker, and exercises the legacy compressed shape explicitly via `ToBytes( compress: true )` instead of relying on this member to produce it.

— Claude for @gfraiteur

PR #1736, opened 2026-07-19
Clean up the cross-version transitive manifest path (#1710)

gfraiteur:
> **Stacked on #1735.** Base is that PR's branch, so the diff here is only the cleanup. Retarget to `develop/2026.1` once #1735 merges.

Docs-and-dead-code follow-up to #1735. No behavioral change intended beyond removing work whose result was discarded.

## What "cross-version" actually means

Two projects in the same solution, one referencing the other, each built against a **different version of Metalama**. Both versions are loaded in the **same process** and reach each other through the version-neutral `Metalama.Framework.DesignTime.Contracts` assembly (`[ComImport]`/`[Guid]` for type equivalence, an `AppDomain` data slot as the rendezvous). Their `Metalama.Framework.Engine` types have distinct identities, so no manifest object can be passed across — the manifest is serialized by the producing version and deserialized into the consuming version's object model. That is the same assembly-identity problem #1710 addressed, not a transport problem.

## Summary

- **Removes a deserialization whose result has had no reader since #1710.** The cross-version branch of `GetDesignTimeProjectVersionAsync` deserialized the referenced project's manifest and stored it on the `DesignTimeProjectReference`. Its two readers — `ReferencedExtensions` and `TryGetReusableTransitiveAspectsManifest` — both narrow to the concrete `DesignTimeAspectPipelineResult`, which a deserialized `TransitiveAspectsManifest` never is, so both silently ignored it. Its only remaining effect was satisfying the reference's both-or-neither invariant.
- **Types `TransitiveAspectsManifest` as the concrete result**, so "a cross-version reference has no live result" is stated by the type instead of by two readers narrowing and discarding whatever does not match. The invariant relaxes accordingly: a live result implies a serialized manifest, not the converse.
- **Renames `SerializeTransitiveAspectManifestForRpc` → `...ForOtherVersion`.** Nothing on that path is RPC. The genuine RPC in this codebase — user process ↔ analysis process, `Metalama.Framework.DesignTime.Rpc`, consumed by Metalama.Vsx — is untouched.
- **Rewords the docs across the area**: the distinction is same-version vs different-version, not in-process vs out-of-process. Both are in-process.

### Why it was left behind

It was load-bearing before #1710. `DesignTimeProjectVersion` then exposed `GetTransitiveAspectsManifest`, which returned this object, and that is what `TransitivePipelineContributorSource` consumed for a `CompilationReference`. #1710 replaced it with `GetSerializedTransitiveAspectsManifest` and routed the consumer through the serialized bytes — where the deserialization *is* cached, by `TransitiveManifestDeserializationCache`. The eager deserialize simply stayed.

This is why caching it would have been the wrong fix: the cache would have been keyed against the *referenced* project's compile-time copy, not the consuming project's that the cache documents, to memoize a value nobody reads.

## Second commit, separable

`e586f9d` drops `DesignTimeAspectPipelineResult : ITransitiveAspectsManifest`, which becomes unused once the property is typed concretely — no remaining cast to that interface, no interface-typed slot it flows into. It also deletes the "still used for PE references" comment, stale since #1710: a PE reference reads its manifest from the assembly metadata resource produced by the compile-time pipeline and never touches a `DesignTimeAspectPipelineResult`. Kept as its own commit so it can be reverted alone if the interface is wanted for a reason not visible in the code.

## Testing

`Metalama.Framework.Tests.UnitTests` design-time suite: 437 passed, 5 skipped, 0 failed. Premium's `Metalama.Extensions.Validation.UnitTests` (13 tests, including the three `CrossProject` failures #1735 fixes) pass against a locally built Metalama carrying both branches.

**The cross-version path itself has no automated coverage** — no test in the repo touches `ITransitiveCompilationService`, `GetTransitiveAspectManifestAsync`, or the entry-point consumer, which is why a vestigial deserialization sat there unnoticed. Exercising it needs two Metalama versions in one process, so the change to that branch is reviewed, not tested. Reviewers may want to sanity-check a two-version solution in VS before merging.

One judgement call worth a look: the branch still calls `GetOrCreatePipelineAsync` / `GetConfigurationAsync` for the reference. Those are now only a gate — the referenced project must be compilable by the current version, and a failure aborts here rather than surfacing later as a consumer-side binding error. I kept them and documented that. If they are genuinely unnecessary they could go too, but I could not verify that without the coverage above.

## Issues Fixed

None — follow-up cleanup to the (closed) #1710 work.

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipeline.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipeline.cs
@@ -608,8 +608,13 @@ public sealed partial class DesignTimeAspectPipeline : BaseDesignTimeAspectPipel
                     }
                     else
                     {
-                        // To deserialize the manifest, we need a service provider with the CompileTimeProject of the referenced project, compiled
-                        // for the current Metalama version.
+                        // The referenced project must be compilable by the *current* Metalama version, because the
+                        // consuming project binds the manifest against its own compile-time closure, which contains
+                        // this reference. Obtaining the configuration is what gates that, so a failure here aborts
+                        // rather than surfacing later as a binding error attributed to the consumer.
+                        //
+                        // Note that we deliberately do not deserialize the manifest here. See the comment on the
+                        // DesignTimeProjectReference constructed below.
 
                         var pipelineResult = await this._pipelineFactory.GetOrCreatePipelineAsync( reference, cancellationToken );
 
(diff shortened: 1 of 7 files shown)


gfraiteur:
Folding this into #1735 at the author's request — a single PR is preferred over the stack. The three commits are unchanged and now all live on `topic/2026.1/1710-duplicate-design-time-validators`; #1735's description has been updated to cover the cleanup as well. Closing this and deleting the branch (no commits are lost — they are all on #1735's branch).

Issue #1758, opened 2026-07-29
Surface an unconsumable reference in the IDE through an info bar


gfraiteur:
## Context

When a project references an assembly compiled by a Metalama of an older design-time contracts generation
(below 2026.1, see #1605), the design-time pipeline cannot obtain that reference's transitive aspects. #1757 makes
this degrade gracefully: the reference contributes no aspect, the rest of the project keeps working, and the user is
warned at **compile time** by `LAMA0077`.

Compile time is currently the only reliable channel. A diagnostic produced on the design-time failure path never
reaches the user, because design-time diagnostics surface through the analyzer, which reads a *successful* pipeline
result.

That leaves a gap: a user who is editing but has not built since the reference changed sees aspects quietly missing
from the reference, with nothing in the IDE explaining why.

## Proposal

Surface the condition in the IDE through an info bar, using the existing cross-version notification channel.

`IDesignTimeNotificationService` (`Metalama.Framework.DesignTime.Contracts.Notifications`) already exists for exactly
this kind of message, and its documentation states that it is the cross-version-safe route by which the Visual Studio
extension consumes design-time events:

> This interface replaces direct consumption of `Metalama.Framework.DesignTime.Rpc` by cross-version consumers
> (e.g. the Visual Studio extension). The Metalama-side implementation translates internal RPC events into
> `IDesignTimeNotificationEvent` instances and delivers them to the observer in-process; no RPC pipe is opened across
> a Metalama-version boundary.

The message flows *from the current Metalama to the extension*, so it never needs to reach the older Metalama and is
unaffected by the contracts-generation break that causes the condition in the first place.

## Work involved

1. A new event type in `DesignTimeNotificationEventTypes`, whose string values are, by the type's own remarks,
   "part of the cross-version contract and frozen forever", plus a new `IDesignTimeNotificationEvent`-derived
   interface with its own frozen `[Guid]`.
2. Raising it from the design-time pipeline where the reference is skipped
   (`DesignTimeAspectPipeline.SkipReferenceUnsupportedAtDesignTime`).
3. Rendering the info bar in `Metalama.Vsx`, which is a separate repository.

## Caveats

- It only reaches users who have the Visual Studio extension installed, so it complements the compile-time warning
  rather than replacing it.
- It should not be noisy: one bar per solution session naming the offending references, not one per project or per
  request.

Split from #1757, which fixes the hang and the degradation but deliberately leaves this channel out of scope.

— Claude for @gfraiteur

Issue #1750, opened 2026-07-29
Design-time crash when a project has duplicate or empty project references


PostSharpAgent:
## Symptom

The design-time incremental diff aborts while building the map of project references between two
compilations:

```
System.ArgumentException: An element with the same key but a different value already exists. Key: '<assembly>'
System.ArgumentException: An element with the same key but a different value already exists. Key: ', 0'
```

The second key, `', 0'`, is worth noticing: it is an empty assembly name followed by a zero. Whatever the
key is built from, at least one reference is producing an empty identity, and two empty identities collide
with each other.

## Where it happens

```
System.Collections.Immutable.ImmutableDictionary`2.Add(TKey key, TValue value, KeyCollisionBehavior behavior, MutationInput origin)
Metalama.Framework.DesignTime.Pipeline.Diff.ProjectVersionProvider.Implementation.GetProjectReferencesAsync(Compilation oldCompilation, Compilation newCompilation, TestableCancellationToken cancellationToken)
Metalama.Framework.DesignTime.Pipeline.Diff.ProjectVersionProvider.Implementation.GetReferencesAsync(Compilation oldCompilation, Compilation newCompilation, TestableCancellationToken cancellationToken)
```

`GetProjectReferencesAsync` adds one entry per project reference, keyed by an identity derived from the
reference, using `ImmutableDictionary.Add`, which throws on a duplicate key rather than replacing or
skipping.

Two references can legitimately produce the same key: a solution that references the same project through
two paths, or a compilation reference whose identity is not yet resolved and comes back empty, which the
`', 0'` case demonstrates has actually happened in the field.

## Scale

Reported by **11 distinct installations** across **17 crash signatures**, producing **34 reports**, over the
retained window (three months).

| Lineage | First version | Last version | Signatures |
|---|---|---|---|
| 2026.1 | 2026.1.18 | 2026.1.18 | 2 |
| 2026.0 | 2026.0.7-rc | 2026.0.23 | 17 |
| 2025.1 | 2025.1.8 | 2025.1.17 | 12 |

**Not a regression**, present across three lineages, and observed on 2026.1.18 as recently as 2026-07-16.

This is a small cluster and is filed as its own issue rather than folded into the other duplicate-key
crashes because the site and the fix are different: this one is about project references, not aspect classes
or syntax trees. It shares only the shape.

## Suggested action

- **Find out what produces an empty identity.** The `', 0'` key is the most actionable thing here, and it is
[...]

Issue #1743, opened 2026-07-28
Duplicate aspect weaver or assembly identity aborts pipeline initialization


PostSharpAgent:
## Symptom

Pipeline initialisation aborts because two things claim the same identity. Aspect code generation and
diagnostics stop for the project.

```
System.ArgumentException: An element with the same key but a different value already exists.
Key: 'Metalama.Community.AutoCancellationToken.AutoCancellationTokenWeaver'

System.ArgumentException: An element with the same key but a different value already exists.
Key: 'Metalama.Community.Virtuosity.VirtuosityWeaver'
```

Note "**but a different value**": the two entries are not equal. Two distinct objects are being registered
under one aspect-driver name, which is what happens when the same aspect assembly reaches the compilation
twice through different routes and is loaded twice.

## Where it happens

Two registration sites, both keyed by a name that the reference graph does not guarantee is unique:

```
System.Collections.Immutable.ImmutableDictionary.ToImmutableDictionary[TSource,TKey](IEnumerable`1 source, Func`2 keySelector)
Metalama.Framework.Engine.Aspects.AspectDriverFactory..ctor(CompilationModel compilation, ImmutableArray`1 plugins, ProjectServiceProvider& serviceProvider)
Metalama.Framework.Engine.Pipeline.AspectPipeline.TryInitialize(IDiagnosticAdder diagnosticAdder, Compilation compilation, IReadOnlyList`1 compileTimeTreesHint, CancellationToken cancellationToken, AspectPipelineConfiguration& configuration)
```

```
Metalama.Framework.Engine.Aspects.TransitivePipelineContributorSource..ctor(CompilationContext compilationContext, AspectClassCollection aspectClasses, ProjectServiceProvider serviceProvider)
Metalama.Framework.Engine.Pipeline.AspectPipeline.CreatePipelineContributorSources(AspectPipelineConfiguration configuration, CompilationContext compilationContext, CancellationToken cancellationToken)
```

`AspectDriverFactory` keys on the weaver type name; `TransitivePipelineContributorSource` keys on assembly
identity (`<name>, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null`). Both use a plain
`ToImmutableDictionary` / `Add`, which throws rather than deduplicating.

The named keys are **Metalama Community packages**, which is a strong hint about the route: a project that
picks up `Metalama.Community.*` both as a package and through a project reference, or through two
transitive paths resolving to different builds, produces two loaded copies of the same weaver.

## Scale

[...]

Issue #1605, opened 2026-04-28
Add version-invariant notification contract in Metalama.Framework.DesignTime.Contracts


gfraiteur:
## Background

A recent diagnosis (see PR #1603 thread and the new design doc `Metalama.Framework/docs/cross-process-communication.md`) showed that the VSX-side CodeLens notification path violates Rule 3 of our cross-process / cross-version design: it consumes `Metalama.Framework.DesignTime.Rpc` directly across both a process boundary and a Metalama-version boundary.

`Metalama.Framework.DesignTime.Rpc` is by design a **same-version, cross-process** layer. Its wire format may evolve freely (e.g., #1299 switched from JSON to MessagePack), and its contract types intentionally lack `[Guid]` markers. Cross-version traffic must instead go through `Metalama.Framework.DesignTime.Contracts`, which uses CLR COM type-equivalence (every interface and DTO carries a frozen `[Guid("…")]`).

Symptom of the violation in the wild: `StreamJsonRpc.ConnectionLostException` thrown from `PostSharp.VisualStudio.CodeLens.NotificationListener.GetNotificationListenerAsync` when VSX 2026.0.x talks to a Metalama 2026.1.x deployment. Stack ends in `Metalama.Framework.DesignTime.Rpc.Notifications.CodeLensProcessRpcClient.OnRpcConnectedAsync`.

## Scope

Add a **version-invariant notification subscription contract** to `Metalama.Framework.DesignTime.Contracts` so VSX (and any other cross-version consumer) can subscribe to design-time notifications without touching the internal RPC layer.

### Surface to design

A new interface (proposed name: `IDesignTimeNotificationService`) in `Metalama.Framework.DesignTime.Contracts/Notifications/` with:

- `[Guid("…")]` (freshly minted, frozen forever).
- The minimal subset of `IEventHubRpcApi` semantics that VSX actually needs (subscribe to event-type names; receive raised events). Mirror the existing `RaiseEventAsync` / `SubscribeAsync` shape but with `[Guid]`-marked DTOs in `Contracts`, not in `Rpc`.
- Companion `[Guid]`-marked DTOs (e.g., `ICompilationResultChangedEvent`, `IEndpointChangedEvent`) covering at minimum the events the existing CodeLens path subscribes to: `CompilationResultChangedEventData`, `EndpointChangedEventData`.
- Per the design doc checklist: never mutate published members; evolve via new `[Guid]` interfaces (`IDesignTimeNotificationService2`, etc.).

[...]

PR #1760, opened 2026-07-30
Make ProjectKey and the compile-time assembly name identify what they name (#1749)

### LINKED ISSUE #1749: Two compile-time projects with the same assembly name crash CompileTimeProjectRepository and the serialization binder
## Symptom

Two sites that key compile-time projects by assembly name throw when the same name appears twice:

```
System.ArgumentException: An item with the same key has already been added.
Key: <assembly>, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null

System.ArgumentException: An item with the same key has already been added. Key: Metalama.Framework
```

## Where it happens

Building the compile-time project repository from the compilation's references:

```
System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
Metalama.Framework.Engine.CompileTime.CompileTimeProjectRepository.Builder.TryGetCompileTimeProjectFromCompilation(ClassifyingCompilationContext compilationContext, IReadOnlyList`1 compileTimeTreesHint, IDiagnosticAdder diagnosticSink, Boolean cacheOnly, CancellationToken cancellationToken, CompileTimeProject& compileTimeProject)
Metalama.Framework.Engine.CompileTime.CompileTimeProjectRepository.Builder.TryBuild(Compilation compilation, IReadOnlyList`1 compileTimeTreesHint, IDiagnosticAdder diagnosticSink, Boolean cacheOnly, CancellationToken cancellationToken, CompileTimeProjectRepository& loader)
```

and building the closure index during compile-time serialization:

```
System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
Metalama.Framework.Engine.CompileTime.CompileTimeProject.get_ClosureProjectsByCompileTimeAssemblyName()
Metalama.Framework.Engine.CompileTime.Serialization.CompileTimeSerializationBinder.BindToName(Type type, String& typeName, String& assemblyName)
Metalama.Framework.Engine.CompileTime.Serialization.SerializationWriter.WriteTypeName(Type type)
```

The second is the more serious of the two, for two reasons. It fails during **serialization**, so the
failure surfaces while writing a type name rather than while loading anything, which makes the cause
invisible from the message. And `ClosureProjectsByCompileTimeAssemblyName` is a property, so the same
`ToDictionary` runs on every access.

`Key: Metalama.Framework` is a notable instance: the closure contains two compile-time projects both
claiming to be `Metalama.Framework`, which is what happens when a reference graph pulls in two Metalama
versions.

## Scale

Reported by **16 distinct installations** across **40 crash signatures**, producing **63 reports**, over the
retained window (three months).

| Lineage | First version | Last version | Signatures |
|---|---|---|---|
| 2026.1 | 2026.1.16 | 2026.1.18 | 15 |
| 2026.0 | 2026.0.5-preview | 2026.0.23 | 28 |
| 2025.2 | 2025.2.5-rc | 2025.2.5-rc | 4 |
| 2025.1 | 2025.1.9 | 2025.1.17 | 20 |

**Not a regression**, present across four lineages and still arriving on both supported ones.

## Suggested action

- **A compile-time closure containing two projects with the same assembly name is a diagnosable condition**,
  and naming both of them would let a user fix their reference graph. Throwing `ArgumentException` from
  `ToDictionary` tells them nothing.
- **`ClosureProjectsByCompileTimeAssemblyName` should be computed once**, not on every property access, so
  that whatever policy is chosen is applied in one place and the cost is paid once.
- The `Metalama.Framework` case specifically suggests checking what happens when a project transitively
  references two Metalama versions, which the other clusters in this corpus show is common in the field.

---

*Filed from CEIP crash reports. 40 invariant problems; ids are recorded against this candidate in the triage
database. Figures cover the retained window (three months) and count distinct reporting installations.*

### LINKED ISSUE #1757: Design-time pipeline hangs on a reference built with a pre-2026.1 Metalama
## Symptom

At design time, a project that references an assembly built with a **pre-2026.1** Metalama hangs its pipeline
permanently. `DesignTimeAspectPipeline.GetDesignTimeProjectVersionAsync` acquires the lock on the project and then
waits forever inside `IDesignTimeEntryPointConsumer.GetServiceProviderAsync`:

```
Thread 17, AspectPipeline: Lock on 'Consumer, f835c184f5ec0369' acquired.
Thread 17, DesignTimeEntryPointManager: GetProviderForVersionAsync(2025.1.18.0)
Thread 17, DesignTimeEntryPointManager: GetProviderForVersionAsync(2025.1.18.0): waiting for a new registration.
Thread 22, AspectPipeline: WARNING: The following call stack has been holding the lock for 'Consumer' for a long time
```

The blocked thread is inside Roslyn's generator driver, so the whole design-time experience for that project stops:
no generated sources, no diagnostics, no code lens.

## This is not the contracts break itself

The incompatibility is **by design**. #1605 rotated every `[Guid]` and switched the `DesignTimeEntryPointManager`
`AppDomain` data slot:

| Contracts generation | Metalama versions | `AppDomain` data slot |
|---|---|---|
| v1 | up to 2026.0.x | `Metalama.Framework.DesignTime.Contracts.DesignTimeEntryPointManager` |
| v2 | 2026.1 and later | `Metalama.Framework.DesignTime.Contracts.**v2**.DesignTimeEntryPointManager` |

A v2 process therefore *can never* observe a v1 registration, and that is intended. Making the two generations
interoperate is not what this issue asks for.

## The defect is the failure mode

`DesignTimeEntryPointManager.Consumer.GetProviderForVersionAsync` loops on `await _registrationTask.Task` with no
exit condition. For a pre-v2 producer the awaited registration is definitionally impossible, so the loop never
terminates, and it does so while the caller holds the project's pipeline lock.

The correct behaviour is to recognise the situation immediately and report it. `LAMA0061` already says exactly the
right thing and is already used for the equivalent compile-time case in
`CompileTimeProjectRepository.Builder.TryDeserializeCompileTimeProject`:

> The referenced assembly '{0}' has been compiled with Metalama {1}. It must be recompiled with the current version
> because backward compatibility of compiled assemblies has been broken.

## Two adjacent gaps, reachable without the above

1. **`ContractVersionMismatchDetected` has no subscribers.** `IDesignTimeEntryPointConsumer.ObserveOnContractVersionMismatchDetected`
   is declared and raised, but nothing in the product observes it, so the designed mismatch signal goes nowhere.
2. **`InvalidCompilerServiceProvider` is not handled by its caller.** Its `GetService` returns `null` for every
   service, and `DesignTimeAspectPipeline` does
   `serviceProvider.GetService( typeof(ITransitiveCompilationService) ).AssertNotNull()`. So the designed
   *same-generation* contract mismatch path throws an assertion failure instead of reporting a diagnostic.

## A timeout alone is not a fix

Bounding the wait with a linked `CancellationTokenSource` converts the deadlock into a livelock. `GetServiceProviderAsync`
caches the pending task per version (`_getProviderTasks.GetOrAdd( version, GetProviderForVersionAsync )`), so cancelling
one await abandons that caller but leaves the shared task pending; every subsequent design-time request pays the
timeout again. Measured with a 30 second timeout: 9 consecutive failures in 300 seconds, each re-acquiring and holding
the project lock.

## Proposed fix

1. In `DesignTimeAspectPipeline`, compare the reference's Metalama version against the minimum version that uses the
   current contracts generation (2026.1) **before** waiting. Below it, report `LAMA0061` against the reference and
   fail the reference immediately.
2. Handle a returned `InvalidCompilerServiceProvider` with the same diagnostic instead of `AssertNotNull`.
3. For same-generation versions that may genuinely still be initializing, keep a bounded wait, but make the negative
   result sticky so it is not re-paid on every request.

## Reproduction

`Metalama.Framework/src/tests/DesignTimeStandalone/Issue1749.FrameworkVersions`, run with
`Metalama.DesignTime.HostSimulator`. `OldAspects` is compiled against the publicly released Metalama 2025.1.18 and is
consumed as a file reference; `Consumer` is compiled by the version built from this repository. The scenario is
exercised by the build through `ManyDesignTimeSolutions`.

Found while investigating #1749; the two are unrelated beyond having been surfaced by the same reproduction.

— Claude for @gfraiteur

gfraiteur:
## Summary

Closes #1749. The family of defects behind it came from **two under-specified identifiers used as if they were identities**, plus a check-then-act gap in how compile-time assemblies are isolated. Also includes a design-time host simulator, the test infrastructure to run scenarios through it, and the fixes for #1757.

`Build.ps1 test` is green: 30m07s, zero failures, all five design-time simulator scenarios succeeded.

## The conceptual defect

**1. `ProjectKey` did not identify a project.** It is `(assembly name, hash of the compilation symbols)`, and it must be serializable and computable from a bare `Compilation`, because it identifies a project across processes and a `Compilation` does not know its project. The symbols stood in for target framework and configuration, which assumes two projects sharing an assembly name are one logical project in two flavours. Two unrelated projects producing one assembly name at one target framework break that, and nothing downstream can detect it: a consumer is handed another project's pipeline, and the failure surfaces far away while serializing a transitive manifest.

Fixed at the input rather than the structure. `Metalama.Framework.targets` now defines a `METALAMA_PROJECT_<hash>` symbol from the project path, target framework, configuration and platform. The path is **relative to `MetalamaRootDirectory`**, which defaults to `$(SolutionDir)`, so the symbol does not depend on the directory into which the repository is cloned; see *Build determinism* below. Compilation symbols are the only project-identifying data that travels *inside* a `Compilation`, so the existing key becomes unique with **no change to `ProjectKey`, no serialization change and no RPC contract change**. A compiler-visible property could not serve: those reach an analyzer through `AnalyzerConfigOptions`, which `ProjectKeyFactory` cannot see.

The symbol is defined **even when `MetalamaEnabled` is `False`**. It has no effect on the compilation of the project itself, and a project that Metalama does not process is still *referenced* by projects that it does, which identify it by the same key. Defining it only for enabled projects left two Metalama-disabled projects of one assembly name sharing a key, which two design-time scenarios reproduce.

`ProjectKey` now hashes **only** the `METALAMA_PROJECT_*` symbols when present, so adding an unrelated `#define` no longer changes a project's key and churns the design-time pipeline cache. Every symbol is still hashed when no discriminator exists.

**The guarantee has one documented limit.** A project that does not reference Metalama at all never receives the symbol, so two such projects producing one assembly name still share a key. That is legitimate, cannot be repaired from our side, and must not cost the consuming project its design-time support, so it is the one case that is resolved (first reference wins) rather than reported. `ProjectKey.IsMetalamaEnabled` is what separates the two cases.

**2. The `ml!<name>_<hash>` compile-time name did not identify its content.** `ComputeProjectHash` covered the assembly name and version but neither the culture nor the public key — exactly the two components the C# compiler lets differ so two assemblies of one simple name can be referenced side by side. Two distinct assemblies therefore collapsed onto one compile-time name.

Fixed by hashing every component of the assembly identity, appended individually rather than through `GetDisplayName`, whose format is not part of the contract of `AssemblyIdentity` and must not feed a hash. This makes the collision impossible rather than reported. That matters: several projections of one run-time assembly in a closure are legitimate, since the per-TFM copies of a multi-targeted library are exactly that.

**3. Compile-time assembly isolation had a check-then-act gap.** Two projects referencing two versions of one *untransformed* compile-time assembly failed under the shared compiler with `LAMA0001 … Assembly with same name is already loaded`. Both compilations are valid and neither closure contains a conflict, so `LAMA0079` neither did nor should catch it; the versions met in the `CompileTimeDomain` shared by the compiler process.

The isolation mechanism already existed, and was given the wrong input then asked the wrong question. `AspectPipeline` chose a domain from the extension assemblies alone, and an untransformed compile-time assembly is an ordinary reference; and compatibility was decided from what a domain had *already loaded*, while a project loads long after being handed its domain. `CompileTimeDomain.TryReserveAssemblies` now checks and records the intended identities as one step under the factory lock, comparing the full identity so that culture and content type count as conflicts too.

A request whose *own* assemblies conflict is not answered with "incompatible". No domain can satisfy such a request, so that answer would mint a fresh domain per request and never say why. The conflict belongs to the input and is diagnosed as `LAMA0079`, so the domain keeps the first identity of each simple name and logs a warning.

This is a gap, not a race: two strictly sequential requests reproduce it, which is why disabling parallelism never helped.

## Reporting a duplicate `ProjectKey`

A collision is now reported wherever the invariant is supposed to hold, instead of being worked around:

- `DesignTimeAspectPipelineFactory.TryGetPipeline` compares the path, the target framework and the configuration of the cached pipeline's options against the caller's, and throws naming the property that differs. Serving a foreign pipeline is what turned one project's compilation into another's, far from where it failed.
- `ProjectVersionProvider` groups the references of both the old and the new compilation and throws naming every colliding assembly identity and the key. Both reference sets route through one helper; the previous code guarded only one of them, and did so by discarding the duplicate silently.
- `WorkspaceProvider.GetProjectAsync` reports the collision instead of returning an arbitrary project.

The first two are limited to keys of projects that reference Metalama, for the reason given above; `WorkspaceProvider` degrades to "project not found", which is a state it already supports, rather than throwing into `GetCompilationAsync`. `DesignTimeAspectPipelineFactory` needs no such distinction, because a pipeline exists only for a project that Metalama processes, so every key reaching it carries the symbol.

## Build determinism

The symbol is part of the compilation, so its value reaches the output: with a deterministic build, changing it changes both the assembly and the PDB, which was verified by building one project twice with two different values. Hashing the **full** project path would therefore have made the compiled assembly depend on the clone directory, breaking reproducible builds, and `/pathmap` could not normalize it because the hash is computed in MSBuild and never seen by the compiler as a path.

The identifier is therefore the path relative to `MetalamaRootDirectory`, which defaults to `$(SolutionDir)`. Two clones at different roots now produce the same symbol; verified against the real targets file, both yielding `N747154918` for one project. Set the property explicitly when one project is built from several solutions and must produce the same symbol in each. When neither the property nor `$(SolutionDir)` is defined, which is the case for `dotnet build` on a bare project file, only the file name of the project is used.

Two further normalizations: the directory separator, so a project yields one symbol on every platform, and the spaces of `$(Platform)`, because a solution configuration names it `Any CPU` whereas a project names it `AnyCPU`.

`[MSBuild]::StableStringHash` is 32 bits, so two projects of one solution can collide, with the probability of the birthday problem. The wider overload is not available in every supported MSBuild version, and a collision is not silent: the path, target framework and configuration are compared before a cached pipeline is served, so it surfaces as the exception above rather than as a wrong pipeline.

`MetalamaDefineProjectDiscriminator` has been **removed**. It suppressed the symbol at the cost of the design-time experience, and the reproducible build that motivated it is now supported directly.

## Diagnostics

| Id | Meaning |
|---|---|
| `LAMA0078` | a project reference below 2026.1, whose aspects the IDE cannot show |
| `LAMA0079` | two references provide one compile-time assembly (untransformed only) |
| `LAMA0080` | *(warning)* the `METALAMA_PROJECT_*` symbol is missing, so design-time is silently broken |
| `LAMA0081` | a project reference built by a different version of Metalama |

`LAMA0080` is reported at compile time although the consequence is a design-time one, because a diagnostic on the design-time path does not reach the user (#1758). It is a **warning** and the pipeline continues after it: the build that reports it is correct, and failing would deny the user a working build over a condition that affects only the editor. It is gated on the new `IProjectOptions.RequiresCompilationConstant`, which only `MSBuildProjectOptions` sets, since only an MSBuild-driven host runs the targets. Test harnesses and embedding hosts are exempt without faking anything.

`LAMA0081` was `LAMA0077` until #1743 took that id while this branch was open. The merge was clean and left **two** `LAMA0077` descriptors in the file, which is worth a duplicate-id guard in its own right.

## Design-time host simulator

`Metalama.Framework/src/tests/Metalama.DesignTime.HostSimulator`, run by `ManyDesignTimeSolutions` over `src/tests/DesignTimeStandalone`.

- Reproduces design-time defects outside an IDE, with **no reference to Metalama**: it hosts Metalama through the analyzer references of the simulated solution, so it exercises whichever version each project references.
- One `AssemblyLoadContext` per analyzer directory, mirroring Roslyn's `AnalyzerAssemblyLoader.DirectoryLoadContext`; source generators through `CSharpGeneratorDriver`; then the per-document analyzer requests an editor issues.
- `--traversal Solution|Graph|Reverse`, `--order`, `--permutations`, because a design-time defect can depend on which project is analyzed first. Bounded by `--timeout` with a hard watchdog, since a design-time defect can be a deadlock.
- Diagnostics are written in canonical MSBuild format, so a design-time scenario asserts with the same `test.json` syntax as a compile-time one.

Requires PostSharp.Engineering >= 2023.2.398; `Directory.Packages.props` moves to 2023.2.408, which also carries the two harness fixes below.

## Scenarios

| Scenario | Kind | Covers |
|---|---|---|
| `Standalone/Issue1749` | build | `LAMA0079`, **and** the shared-compiler collision |
| `Standalone/Issue1749.PublicKeyVariants` | build | the closure collision, now simply supported |
| `Standalone/Issue1749.SameAssemblyIdentity` | build | `TransitivePipelineContributorSource` |
| `Standalone/Issue1757` | build | `LAMA0078` and `LAMA0081` |
| `DesignTimeStandalone/Issue1749` | simulator | the same reference graph as `Standalone/Issue1749`, through the simulator |
| `DesignTimeStandalone/Issue1749.SameAssemblyIdentity` | simulator | `TransitiveAspectsManifest`, a different site for the same configuration |
| `DesignTimeStandalone/Issue1749.SameProjectKey` | simulator | one project's pipeline given another's compilation |
| `DesignTimeStandalone/Issue1749.UpstreamReuse` | simulator | the upstream-reuse cache keyed wrongly |
| `DesignTimeStandalone/Issue1749.FrameworkVersions` | simulator | two versions of `Metalama.Framework` in one reference graph (#1757) |

`Standalone/Issue1749` deliberately does **not** pass `UseSharedCompilation=false`. The default shared compiler is what exercises the third defect, so one ordinary CI build covers both collisions with no flags, forbidding `Assembly with same name is already loaded` alongside its expected `LAMA0079`. The forbidden pattern matches the message text because that failure had no diagnostic id of its own.

`DesignTimeStandalone/Issue1749.SameProjectKey` is what caught the `MetalamaEnabled=false` gap described above: its two `Contract` projects disable Metalama, so they received no discriminator while the symbol was defined only for enabled projects.

## Fixes for #1757

- `DesignTimeAspectPipeline` no longer waits for an entry point that can never register. #1605 rotated the GUIDs and the `AppDomain` slot of the design-time contracts, so v1 (≤ 2026.0) and v2 (2026.1+) cannot observe each other. That is by design; the unbounded wait under the project's pipeline lock was not.
- A reference now **degrades** to contributing no aspect instead of failing the whole call, so the consuming project keeps design-time support for its own aspects.
- Same-generation versions wait on the **caller's** cancellation token, with no timeout of our own.
- `EmptyWorkspaceProvider` (renamed from `FakeWorkspaceProvider`) returns `null` instead of throwing `NotSupportedException`. That provider is what a plain analyzer gets, that is, an IDE without the Metalama extension, so having no workspace is a normal state of a supported host; the exception propagated out of the source generator and disabled every design-time feature of the project.

## Other fixes from the same investigation

- `AssemblyLocator.TryFindAssembly` and `CompileTimeProject.TryCreateUntransformed` prefer an exact identity match. `AssemblyName.ReferenceMatchesDefinition` matches on the simple name when the reference specifies no version, culture or public key, so both bound to the first loose match. `TryCreateUntransformed` now falls back to the highest version rather than to load order; that tie-break is defensive and has no scenario of its own.
- `TransitivePipelineContributorSource` guards its manifest `Add` (shared with #1743, whose version is kept), and `TransitiveAspectsManifest.Create` groups by `AspectClass.FullName` rather than by the `AspectClass` object.
- `CompileTimeProjectManifest.MetalamaVersion` round-trips. Pre-existing: the property was get-only and its constructor overwrote it with the *reading* engine's version, so `LAMA0061` named the wrong version.
- `ProjectKeyFactory` sorts symbols before hashing, so one project cannot yield two keys, and computes the hash without allocating: the symbols are read as a span, counted in one pass, and the usual case of a single discriminator needs neither a buffer nor a sort.
- Domain selection moves to `CompileTimeDomainFactoryBase`. It was duplicated between the production and test factories and the copies had drifted, so the first version of the isolation fix was invisible to every unit test.
- `Standalone/Issue1757` had no `.sln`, so `ManySolutions` discovered its projects individually and its assertions had never run. Note `dotnet new sln` now emits `.slnx`, which discovery does not glob.
- `CompileTimeDomain.IsCompatibleWithAssemblies` is **removed**: no production code called it, so its two tests would have kept passing whatever `TryReserveAssemblies` did. They now target `TryReserveAssemblies`.
- The system-assembly filter is shared through `CompileTimeConstants.IsSystemAssemblyFileName`. Two copies existed and had already diverged, one testing the assembly name and the other the file name.
- The components of the assembly identity are separated in the compile-time project hash. The hasher appends a string with neither a length nor a terminator, so two identities could otherwise produce one byte sequence.
- The design-time host simulator supplies `SolutionDir` and the other solution properties. `MSBuildWorkspace` evaluates each project on its own and defines none of them, so every simulated project would have taken the no-directory fallback and the scenarios would have stopped exercising the discriminator.
- An interruption by the user is no longer reported as a simulation timeout.

## Tests

Three regression tests in `CompileTimeDomainTests`, each verified failing before the fix and passing on net8.0 and net48:

- `GetOrCreateDomain_VersionConflictBeforeLoading_CreatesNewDomain` — deterministic and single-threaded, because the gap needs no concurrency.
- `GetOrCreateDomain_ConcurrentSelectionBeforeLoading_CreatesNewDomain` — forces the production interleaving (select, select, load, load) with `SemaphoreSlim` sync points rather than leaving it to the scheduler.
- `GetOrCreateDomain_SelfConflictingRequest_ReusesOneDomain` — a request whose own assemblies conflict must not consume a fresh domain each time it is made. Without the fix, two identical requests received domains 1 and 2.

`TryReserveAssemblies_Compatible_ReturnsTrue` and `TryReserveAssemblies_VersionConflict_ReturnsFalse` replace the two tests that exercised the removed `IsCompatibleWithAssemblies`, so they now cover the method production actually calls.

## Breaking Changes

- `IProjectOptions` gains `ProjectReferenceNames` and `RequiresCompilationConstant`. Not `[InternalImplement]`, so a direct external implementer breaks; `DefaultProjectOptions` and `ProjectOptionsWrapper` provide virtual defaults.
- `CompileTimeProjectManifest`'s constructor gains a trailing optional `metalamaVersion` parameter. Internal type.
- Every `ml!` compile-time assembly name changes, because the hash now covers the full assembly identity. Caches are invalidated once, as with any hash-input change.
- Every project that references Metalama 2026.1+ carries a new `METALAMA_PROJECT_*` compilation symbol, **including projects with `MetalamaEnabled=false`**. It cannot be suppressed; `MetalamaRootDirectory` controls its value instead.
- `MetalamaRootDirectory` is a new public MSBuild property, documented in metalama/Metalama.Documentation#285.
- A duplicate `ProjectKey` between two Metalama-enabled projects now throws at design time instead of being silently resolved.

## Issues Fixed

- #1749 - Duplicate compile-time assembly name
- #1757 - Design-time pipeline hangs on a reference built with a pre-2026.1 Metalama

## Related

- **#1750 is not fixed here** and #1755 should stay open: its root cause is that a null or empty `Compilation.AssemblyName` yields an identity-less `ProjectKey`, because `Invariant.AssertNotNull` is a no-op outside `DEBUG`. This branch now rewrites `GetProjectReferencesAsync` to route both reference sets through one `GetSingleCompilation` helper, which **will conflict textually with #1755's rewrite of the same method**. #1755 additionally rejects the identity-less key, so **#1755's version should win**, with the `IsMetalamaEnabled` distinction from this branch carried over.
- **#1758** remains open: a design-time-only condition is still not surfaced in the IDE, though `LAMA0080` now covers the case that mattered here by reporting at compile time.
- Two harness defects found here are fixed in PostSharp.Engineering: `Build.ps1` discarded the tool's exit code, so every failure reported success (2023.2.408), and `ManySolutions` did not glob `*.slnx`, so a solution created by `dotnet new sln` was discovered as loose projects and its `test.json` never applied (2023.2.407). The first is why `Standalone/Issue1757` looked green while its assertions had never run.
- metalama/Metalama.Documentation#285 documents `MetalamaRootDirectory` and the project-identity mechanism, and should merge together with this PR.
- PR #1753 was closed in favour of this work.

— Claude for @gfraiteur


Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeProjectRepository.Builder.cs
```diff
@@ -407,6 +412,67 @@ private bool TryGetCompileTimeProjectFromPath(
...
+        /// </para>
+        /// </remarks>
+        private void ReportMixedVersionWarnings(
+            AssemblyIdentity runTimeAssemblyIdentity,
+            CompileTimeProjectManifest manifest,
+            IDiagnosticAdder diagnosticAdder )
+        {
+            var projectReferenceNames = this._projectOptions?.ProjectReferenceNames ?? ImmutableArray<string>.Empty;
+
+            if ( projectReferenceNames.IsDefaultOrEmpty
```
gfraiteur, line 535: Add short comment for this gate.

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeProjectRepository.Builder.cs
```diff
@@ -407,6 +412,67 @@ private bool TryGetCompileTimeProjectFromPath(
...
+                 || !projectReferenceNames.Any( n => string.Equals( n, runTimeAssemblyIdentity.Name, StringComparison.OrdinalIgnoreCase ) ) )
+            {
+                return;
+            }
+
+            var currentVersion = AssemblyMetadataReader.GetInstance( typeof(CompileTimeProjectManifest).Assembly ).PackageVersion;
```
gfraiteur: can be cached as static field

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Diff/ProjectVersionProvider.Implementation.cs
```diff
@@ -129,8 +132,16 @@ private async ValueTask<ReferenceChanges> GetReferencesAsync(
...
+            // produce one assembly name yield one ProjectKey, and a plain ToDictionary throws on the duplicate. The
+            // first reference wins here too, so both sides of the diff resolve a colliding key the same way (#1749).
+            //
+            // PR #1755 rewrites this whole method for #1750, routing both reference sets through one helper and also
+            // rejecting the identity-less ProjectKey that an empty Compilation.AssemblyName produces. Its version is
+            // better and should win the merge; this exists only so that this branch is not red in the meantime.
             var oldProjectReferences = oldCompilation?.ExternalReferences.OfType<CompilationReference>()
-                .ToDictionary( x => x.Compilation.GetProjectKey(), x => x.Compilation );
+                .GroupBy( x => x.Compilation.GetProjectKey() )
+                .ToDictionary( g => g.Key, g => g.First().Compilation );
```
gfraiteur: ProjectKey is no longer supposed to have duplicates, so we can fail here. However, we need to fail with a meaningful exeption message including the duplicate SyumbolKey. So, use three steps: compute the group, throw if count > 1, then to ToDicitonary.

ALTERNATIVELY we might judge that this assertion is already sufficiently asserted elsewhere in this PR, and allow the exception to be weakly documented -- but then there is no need for group and then first

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineFactory.cs
```diff
@@ -177,6 +182,26 @@ private bool TryGetPipeline(
...
+            if ( projectOptions.ProjectPath != null
+                 && pipelineOptions.ProjectPath != null
+                 && !string.Equals( projectOptions.ProjectPath, pipelineOptions.ProjectPath, StringComparison.OrdinalIgnoreCase ) )
+            {
+                throw new InvalidOperationException(
+                    $"Two different projects have the same Metalama project key '{projectKey}': '{pipelineOptions.ProjectPath}' and "
+                    + $"'{projectOptions.ProjectPath}'. A project key is an assembly name and a hash of the compilation symbols, and it must "
+                    + "identify a project uniquely. Check that the METALAMA_PROJECT_* compilation symbol defined by Metalama.Framework.targets "
+                    + "is present in both projects and has not been removed by a DefineConstants assignment." );
+            }
```
gfraiteur, line 231: This check is not sufficient. Also check the TFM and build configuration.

Metalama.Framework/src/Metalama.Framework.DesignTime/Services/FakeWorkspaceProvider.cs
```diff
@@ -7,9 +7,17 @@
...
 
+/// <summary>
+/// The <see cref="WorkspaceProvider"/> of a host that has no workspace, i.e. Metalama running as a plain analyzer.
+/// </summary>
+/// <remarks>
+/// That is the normal situation in an IDE without the Metalama extension, so it returns <c>null</c> rather than
+/// throwing. It used to throw <see cref="NotSupportedException"/>, which travelled out of the source generator and cost
+/// the project every design-time feature instead of the one feature that needs a workspace. See #1749.
+/// </remarks>
 internal sealed class FakeWorkspaceProvider : WorkspaceProvider
```
gfraiteur: Should be renamed to EmptyWorkspaceProvider.

Metalama.Framework/src/Metalama.Framework.DesignTime/Services/LocalWorkspaceProvider.cs
```diff
@@ -27,14 +27,14 @@ public LocalWorkspaceProvider( GlobalServiceProvider serviceProvider ) : base( s
         this._designTimeExtensionManager = serviceProvider.GetService<DesignTimeExtensionManager>();
     }
 
-    protected override Task<Workspace> GetWorkspaceAsync( CancellationToken cancellationToken = default )
+    protected override async Task<Workspace?> GetWorkspaceAsync( CancellationToken cancellationToken = default )
```
gfraiteur: IMO adding async does not do anything, and would be reverted in the next code review. If it does something, add a comment.

Metalama.Framework/src/Metalama.Framework.DesignTime/Services/RemoteWorkspaceProvider.cs
```diff
@@ -105,5 +105,5 @@ private RemoteWorkspaceProvider( GlobalServiceProvider serviceProvider, Workspac
         this._workspace = Task.FromResult( workspace );
     }
 
-    protected override Task<Workspace> GetWorkspaceAsync( CancellationToken cancellationToken = default ) => this._workspace;
+    protected override async Task<Workspace?> GetWorkspaceAsync( CancellationToken cancellationToken = default ) => await this._workspace;
```
gfraiteur: IMO adding async does not do anything, and would be reverted in the next code review. If it does something, add a comment.

Metalama.Framework/src/Metalama.Framework.DesignTime/ProjectKeyFactory.cs
```diff
@@ -45,44 +45,56 @@ internal static ProjectKey Create( string assemblyName, ParseOptions? parseOptio
...
+        // Without this, the same project could yield two different keys, which is as harmful as two projects yielding
+        // one key. CompileTimeCompilationBuilder.ComputeSourceHash already sorts them for the same reason.
+        var allSymbolNames = parseOptions.PreprocessorSymbolNames
+            .OrderBy( x => x, StringComparer.Ordinal )
+            .ToReadOnlyList();
 
-        if ( preprocessorSymbolNames is ImmutableArray<string> immutableArray )
-        {
-            // The Roslyn implementation of PreprocessorSymbolNames is an ImmutableArray, so we allocate less memory.
+        var discriminatorSymbolNames = allSymbolNames
```
gfraiteur: This implementation should be written so that it does not allocate memory, as it is a performance-sensitive method.

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/AssemblyLocator.cs
```diff
@@ -79,7 +80,14 @@ public bool TryFindAssembly( AssemblyIdentity assemblyIdentity, [NotNullWhen( tr
             }
             else
             {
-                reference = candidates[0].MetadataReference;
+                // Prefer an exact identity match when there is one. AssemblyName.ReferenceMatchesDefinition compares
+                // little more than the simple name, so two references of one name that differ in version, culture or
```
gfraiteur: This comment is AI slop. Be accurate, software engineering.

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeCompilationBuilder.cs
```diff
@@ -180,9 +180,13 @@ private ulong ComputeProjectHash(
...
 
-        h.Append( assemblyIdentity.Name );
-        h.Append( assemblyIdentity.Version.ToString() );
-        this._logger.Trace?.Log( $"ProjectHash: AssemblyIdentity='{assemblyIdentity.Name}, {assemblyIdentity.Version}'" );
+        // The FULL identity, not just the name and version. Culture and the public key are exactly the two components
+        // that the C# compiler lets differ so that two assemblies of one simple name can be referenced side by side,
+        // so hashing only the name and the version let two genuinely distinct assemblies collapse onto one
+        // 'ml!<name>_<hash>' compile-time name. The compile-time name is supposed to identify the content it projects,
+        // and the identity of the assembly is part of that content. See #1749.
+        h.Append( assemblyIdentity.GetDisplayName() );
```
gfraiteur: DisplayName is a human-readable string that might change, and should not affect the hash. Use all components explicitly.

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
```diff
@@ -325,20 +356,99 @@ public bool IsCompatibleWithAssemblies( IEnumerable<string> assemblyPaths )
...
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Determines whether two identities of one simple name cannot both be loaded into a single domain.
+        /// </summary>
+        private static bool Conflicts( AssemblyIdentity x, AssemblyIdentity y )
+            => x.Version != y.Version || !x.PublicKeyToken.SequenceEqual( y.PublicKeyToken );
```
gfraiteur: What about culture here?

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeProject.cs
```diff
@@ -355,9 +355,15 @@ public static bool TryCreateUntransformed(
...
-            .FirstOrDefault();
+        var loadedCandidates = AppDomainUtility
+            .GetLoadedAssemblies( a => !domain.IsCollectible( a ) && AssemblyName.ReferenceMatchesDefinition( assemblyName, a.GetName() ) );
+
+        // Prefer an exact identity match. AssemblyName.ReferenceMatchesDefinition compares little more than the simple
+        // name, so taking the first match could bind this project to an already-loaded assembly of a different version,
+        // culture or public key. Two versions of one untransformed assembly cannot share a closure (LAMA0079), but they
+        // can be loaded in one process by two different projects of a solution. See #1749.
+        var assembly = loadedCandidates.FirstOrDefault( a => a.GetName().ToAssemblyIdentity().Equals( assemblyIdentity ) )
+                       ?? loadedCandidates.FirstOrDefault();
```
gfraiteur: FirstOfDefault should not be used on an unsorted collection. Sort it first - I assume by descending version? Can we have a standalone test that exercices this path, or is it pure speculation?

Metalama.Framework/src/Metalama.Framework.Engine/Options/IProjectOptions.cs
```diff
@@ -136,6 +136,36 @@ public interface IProjectOptions : IProjectService, IEquatable<IProjectOptions>
...
     ImmutableArray<TargetedAssemblyReference> CompileTimeAssemblies { get; }
 
+    /// <summary>
+    /// Gets the assembly names of the references that come from a <c>ProjectReference</c>, i.e. of the other projects
+    /// of the same solution.
+    /// </summary>
+    /// <remarks>
+    /// A <c>ProjectReference</c> is a design-time concept. By the time the compiler runs, every reference is a
+    /// <c>PortableExecutableReference</c> and the distinction is lost, so it is supplied by MSBuild, which has already
+    /// computed it. Names and not paths, because the consumer compares them against the assembly identity it reads
```
gfraiteur: Use full sentences, not just stems like "Names and not path". Update CLAUDE.md for acceptable natural language style: strictly no contractions, professional, documentation-grade language, rather on the formal side. Strict grammar. No unusual metaphores or abbrevitations.

Metalama.Framework/src/Metalama.Framework.Engine/Options/MSBuildPropertyNames.cs
```diff
@@ -33,6 +33,8 @@ public static class MSBuildPropertyNames
     public const string MetalamaConcurrentBuildEnabled = nameof(MetalamaConcurrentBuildEnabled);
     public const string MetalamaCompileTimePackages = nameof(MetalamaCompileTimePackages);
     public const string MetalamaCompileTimeAssemblies = nameof(MetalamaCompileTimeAssemblies);
+
+    public const string MetalamaProjectReferenceNames = nameof(MetalamaProjectReferenceNames);
```
gfraiteur, line 36: Should also be in the All collection by convention.

Metalama.Framework/src/Metalama.Framework.Package/build/Metalama.Framework.targets
```diff
@@ -4,6 +4,38 @@
         <!-- Define a METALAMA compilation symbol (aka constant). -->
         <DefineConstants>$(DefineConstants);METALAMA</DefineConstants>
 
+        <!--
+            Define a compilation symbol that identifies this project uniquely within the solution.
+
+            A ProjectKey is an assembly name plus a hash of the compilation symbols, and it must be serializable and
```
gfraiteur: Comment should be more succinct (public facing)

Metalama.Framework/src/tests/DesignTimeStandalone/Issue1749.SameAssemblyIdentity/README.md
```diff
@@ -0,0 +1,107 @@
+# Issue 1749 at design time — two aspect classes with one full name
+
+**Still red, and now for a third reason.** Two failures have been fixed underneath it and a third has surfaced. See
```
gfraiteur: Is it really still read? Don't forget it.

Metalama.Framework/src/tests/DesignTimeStandalone/Issue1749.SameProjectKey/README.md
```diff
@@ -0,0 +1,103 @@
+# Issue 1749 — one project's pipeline given another project's compilation
+
+**Red. The fix for the site it names lives in another pull request**, not here: PR #1755 rewrites
```
gfraiteur: Should no longer be red. Two projects can no longer have the same ProjectKey.

gfraiteur: Correct, and its own `test.json` already said so: it *forbids* the diagnostic, so "Red" contradicted the assertion the scenario ships with. Rewritten as green.

It did fail in the verification run, but on the exception I had just added rather than on the original defect, and that is what exposed the `MetalamaEnabled=false` gap described on the `ProjectVersionProvider` thread. The README now records why that property on the two `Contract` projects is essential to the scenario.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeProjectRepository.Builder.cs
gfraiteur: Added: the gate is now explained as "the reference must be one of the project references of the current project, and an empty list means the host does not supply the property".

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeProjectRepository.Builder.cs
gfraiteur: Cached as `_currentMetalamaVersion`, a `private static readonly` field on `Builder`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.Package/build/Metalama.Framework.targets
```diff
@@ -142,7 +174,30 @@
...
+         compiler runs every reference is a PortableExecutableReference and the distinction is lost;
+         ResolveAssemblyReferences has already computed it as %(ReferenceSourceTarget).
+
+         Assembly names and not paths: the consumer compares them against the assembly identity it reads from the
+         reference, which is robust, whereas a path has to match exactly across two subsystems. That comparison used
+         to fail silently, because when the referenced project produces a reference assembly (the default) the
+         compiler is given the reference assembly under 'obj\<config>\<tfm>
+ef' while @(ReferencePath) still holds
+         'bin\<config>\<tfm>\X.dll'. Names also cannot contain the ',' that separates the list items, so they need
+         no escaping. -->
```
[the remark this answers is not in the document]
gfraiteur: Fixed. That was a literal carriage return I wrote into the file by accident, not a wrap: the `\r` of `\ref` was interpreted as an escape. The path now reads `obj\<config>\<tfm>\ref\X.dll`, and the file has no stray CR left.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/tests/Metalama.DesignTime.HostSimulator/MSBuildEnvironment.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        var output = new StringBuilder();
+        process.OutputDataReceived += ( _, e ) => output.AppendLine( e.Data );
+        process.BeginOutputReadLine();
+        process.WaitForExit();
+
+        if ( process.ExitCode != 0 )
+        {
+            throw new InvalidOperationException( $"'dotnet --list-sdks' failed with exit code {process.ExitCode}." );
+        }
```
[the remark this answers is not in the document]
gfraiteur: Fixed: `ErrorDataReceived` + `BeginErrorReadLine` now drain stderr, and the exception quotes it alongside the exit code.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Directory.Packages.props
```diff
@@ -7,7 +7,9 @@
     <RestorePrerelease>true</RestorePrerelease>
   </PropertyGroup>
   <PropertyGroup>
-    <PostSharpEngineeringVersion Condition="'$(PostSharpEngineeringVersion)'==''">2023.2.396</PostSharpEngineeringVersion>
+    <!-- 2023.2.398 is the first version that exposes ManySolutions and TestableSolution, which the design-time
+         scenarios under Metalama.Framework/src/tests/DesignTimeStandalone derive from. -->
+    <PostSharpEngineeringVersion Condition="'$(PostSharpEngineeringVersion)'==''">2023.2.408</PostSharpEngineeringVersion>
```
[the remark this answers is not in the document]
gfraiteur: 2023.2.408 is correct, so the description was the stale side and is now updated. 408 also carries the two harness fixes this branch found: the discarded `Build.ps1` exit code, and `*.slnx` discovery in 407.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
```diff
@@ -325,20 +346,84 @@ public bool IsCompatibleWithAssemblies( IEnumerable<string> assemblyPaths )
...
-                        this._logger.Trace?.Log(
-                            $"Domain {this._domainId} is incompatible with assembly '{assemblyName}' at '{path}': " +
-                            $"already-loaded assembly has identity '{existingEntry.Identity}'." );
+                var identity = assemblyName.ToAssemblyIdentity();
 
-                        return false;
-                    }
+                if ( !this.IsCompatibleWithIdentity( identity, path ) )
+                {
+                    return false;
```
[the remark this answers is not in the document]
gfraiteur: Real gap, thank you. Returning `false` would have been the wrong repair though: it means "try another domain", and no domain can satisfy a self-conflicting set, so the factory would have minted a fresh domain per request and never said why. The conflict belongs to the input and is diagnosed as LAMA0079, so the domain now keeps the first identity of each simple name (the one the load context resolves to) and logs a warning. Pinned by `GetOrCreateDomain_SelfConflictingRequest_ReusesOneDomain`, which I verified fails without the change: two identical requests got domains 1 and 2.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Diff/ProjectVersionProvider.Implementation.cs
gfraiteur: Took the first option. `GetSingleCompilation` groups, throws when the group holds more than one, and names every colliding assembly identity along with the key. Both reference sets route through it now, because the `ContainsKey` guard 66 lines below was the same decision made the other way and leaving it would have been inconsistent.

One correction to the premise, found by the test suite rather than by reading: a duplicate key is still legitimate. `ProjectKey` uniqueness comes from `METALAMA_PROJECT_*`, which `Metalama.Framework.targets` defined inside `<PropertyGroup Condition="'$(MetalamaEnabled)'!='False'">`, so a project with Metalama disabled never received it. `DesignTimeStandalone/Issue1749.SameProjectKey` has exactly that: two `Contract` projects, one assembly name, `MetalamaEnabled=false`. An unconditional throw broke it and `DesignTimeStandalone/Issue1749`.

So: the discriminator is now defined unconditionally (it is a pure identifier with no effect on the project's own compilation, and a project Metalama does not process is still referenced by projects it does), and the throw is guarded on `ProjectKey.IsMetalamaEnabled`. A project that does not reference Metalama at all can still collide, cannot be fixed from our side, and must not kill the consumer's design-time support, so that case keeps first-wins with a warning.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineFactory.cs
gfraiteur: Fixed. `GetProjectIdentityMismatch` compares the path, the target framework and the configuration, each only when both sides define it, and the message names which one differs and with what values.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.DesignTime/Services/FakeWorkspaceProvider.cs
gfraiteur: Renamed to `EmptyWorkspaceProvider`, with the sole reference in `DesignTimeAnalysisProcessServiceProviderFactory` updated.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.DesignTime/Services/LocalWorkspaceProvider.cs
gfraiteur: You are right that it bought nothing worth keeping. It existed only to bridge `Task<Workspace>` to the nullable `Task<Workspace?>` return type, which generic invariance forbids. Changing the field to `TaskCompletionSource<Workspace?>` removes the need, so the method is a plain non-async one again.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.DesignTime/Services/RemoteWorkspaceProvider.cs
gfraiteur: Same as the sibling: the `async` only bridged `Task<Workspace>` to `Task<Workspace?>`. The field is now `Task<Workspace?>` and the method is `=> this._workspace` with no state machine.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.DesignTime/ProjectKeyFactory.cs
gfraiteur: Rewritten to allocate nothing on the common path. It spans the `ImmutableArray` instead of enumerating it as `IEnumerable`, counts the discriminators in one pass, and hashes the single-discriminator case (the usual one) directly, with no buffer and no sort. Only the multi-symbol fallback rents a buffer, from `ArrayPool<string>.Shared`, returned with `clearArray: true`. The LINQ chain and both intermediate lists are gone.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/AssemblyLocator.cs
gfraiteur: Rewritten. It now states the mechanism and nothing else: `ReferenceMatchesDefinition` matches on the simple name when the reference specifies no version, culture or public key, so the candidate list can hold several distinct identities; the ordering is by version only; an exact identity match is preferred when one exists.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeCompilationBuilder.cs
gfraiteur: Agreed, the display name is not part of the contract. The components are now appended individually: name, version, culture name, public key token, `IsRetargetable` and `ContentType`. The comment records why `GetDisplayName` must not feed a hash. It remains only in the trace message, where a human reads it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeDomain.cs
gfraiteur: It was missing, yes. `Conflicts` is now `!x.Equals( y )`, which covers every component of the identity except the simple name that the caller has already matched, so culture, retargetability and content type are included. Being stricter can only over-isolate, which costs an extra domain and never a wrong load.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.Engine/CompileTime/CompileTimeProject.cs
gfraiteur: Fixed: the fallback is now `MaxByOrNull( a => a.GetName().Version )` rather than `FirstOrDefault` on load order.

On whether it is speculation: partly, and I should say so plainly. The exact-match branch is exercised, since `Standalone/Issue1749.PublicKeyVariants` puts two identities of one simple name in one process. The tie-break among several *non-exact* matches is not covered by any test I can point to, so the ordering is defensive rather than demonstrated.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.Engine/Options/IProjectOptions.cs
gfraiteur: Rewritten in full sentences, split into three `<para>` elements, with the fragments removed. I added the rule to `CLAUDE.md` under Code Style: documentation-grade language, complete sentences, no contractions, no figurative language, no unusual abbreviations. I also swept the comments this PR adds for the same faults and fixed the clearest ones ("still travels through", "rendezvouses", "for good").

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.Engine/Options/MSBuildPropertyNames.cs
gfraiteur: Added to `All`, and moved up next to the other `MetalamaCompileTime*` constants so the declaration order matches.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/Metalama.Framework.Package/build/Metalama.Framework.targets
gfraiteur: Cut from 24 lines to 6: what the symbol is, why the key needs it, how to suppress it and at what cost, and a link to the issue. The design rationale that was there belongs in the code and is now carried by `IProjectOptions.RequiresCompilationConstant` and `ProjectKeyFactory`. The MSB4186 and minus-sign notes stay, next to the two lines they explain.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Metalama.Framework/src/tests/DesignTimeStandalone/Issue1749.SameAssemblyIdentity/README.md
gfraiteur: You were right to doubt it. It is green: the suite run shows `VariantA`, `VariantB` and `Consumer` all `ok`, with code generated. The headline was written mid-investigation and never updated, and it also quoted a `ProjectVersionProvider` warning that no longer exists. Rewritten to describe the fixed state, including how the `AssertionFailedException` that sat behind the `FullName` grouping is resolved by the full-identity `ml!` hash and the `ProjectKey` fix.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #1794, opened 2026-08-03
Do not assume a syntax tree path is unique in a compilation (#1742)

### LINKED ISSUE #1742: Metalama assumes a syntax tree's file path is unique in a compilation; linked files and generated hint names break it
## Symptom

Metalama indexes a compilation's syntax trees by file path, and throws when one path appears twice. The
pipeline stops for the project.

```
System.ArgumentException: An element with the same key but a different value already exists. Key: '<path>'
Metalama.Framework.Engine.AssertionFailedException: <project>( id = 8979 ): Attempting to add duplicate syntax tree '<path>'.
System.InvalidOperationException: There cannot be two files in the compilation with the same filename '<name>'
```

Three messages, one assumption: **that a file path uniquely identifies a syntax tree in a compilation.**
Roslyn does not guarantee that. A compilation can hold two trees with the same path, and routinely does:

- linked files (`<Compile Include="..\Shared\X.cs" Link="X.cs" />`) included twice by overlapping globs,
- a source generator emitting a hint name that collides with a real file,
- multi-targeted or shared projects where the same file participates twice.

None of those are user errors, and none of them are things Metalama can prevent.

## Where it happens

Three distinct sites make the same assumption:

```
Metalama.Framework.Engine.Utilities.Roslyn.CompilationExtensions.GetIndexedSyntaxTreesCore(Compilation compilation)
Metalama.Framework.Engine.Utilities.Caching.WeakCache`2.GetOrAdd(TKey key, Func`2 func)
```

```
Metalama.Framework.DesignTime.Pipeline.DesignTimeAspectPipelineResult.<Update>g__IndexNewTree|36_1(String filePath, SyntaxTreePipelineResult newSyntaxTreeResult, <>c__DisplayClass36_0&)
Metalama.Framework.DesignTime.Pipeline.DesignTimeAspectPipelineResult.Update(PartialCompilation compilation, DesignTimeProjectVersion projectVersion, DesignTimePipelineExecutionResult pipelineResults, AspectPipelineConfiguration configuration)
```

and the explicit check that produces `There cannot be two files in the compilation with the same filename`.

The first is the worst placed of the three: it sits behind a `WeakCache.GetOrAdd`, so the throw happens
inside a cache factory and every later request for the same compilation repeats it.

## Scale

Reported by **52 distinct installations** across **48 crash signatures**, producing **143 reports**, over the
retained window (three months).

| Lineage | First version | Last version | Signatures |
|---|---|---|---|
| 2026.1 | 2026.1.1-preview | 2026.1.18 | 28 |
| 2026.0 | 2026.0.5-preview | 2026.0.23 | 45 |
| 2025.2 | 2025.2.1-preview | 2025.2.5-rc | 9 |
| 2025.1 | 2025.1.4-rc | 2025.1.17 | 28 |

**Not a regression.** Present across four lineages since 2025.1 and still arriving on both supported
lineages. Do not bisect for it.

## Suggested action

The three sites should agree on one answer, and the assertion text suggests they currently do not: one
throws `ArgumentException` from a dictionary, one an `AssertionFailedException`, one a deliberate
`InvalidOperationException`.

- **Key by `SyntaxTree` identity rather than by path** where the index only needs to distinguish trees.
- **Where a path genuinely is the key** (design-time results are addressed by document path), duplicates
  need a defined policy: last one wins, or a diagnostic naming the colliding paths. An assertion is the one
  option that is certainly wrong, because the condition is reachable from ordinary project layouts.
- **Report it as a Metalama diagnostic if it must fail.** "There cannot be two files in the compilation with
  the same filename" is at least an accurate sentence; the other two tell the user nothing they can act on.

---

*Filed from CEIP crash reports. 48 invariant problems; ids are recorded against this candidate in the triage
database. Figures cover the retained window (three months) and count distinct reporting installations.*

gfraiteur:
Fixes #1742.

Metalama indexed the syntax trees of a compilation by `SyntaxTree.FilePath` and assumed the path identified the tree. Roslyn makes no such guarantee. The command-line compiler deduplicates source files itself (`CS2002`), but Roslyn Workspaces has no equivalent step, so overlapping globs, an explicit `Compile` item repeating a glob, `Link` metadata, or a shared project all produce two trees with one path out of an otherwise valid project.

**Seven sites** made the assumption and disagreed about it: two tolerated a duplicate silently, one tolerated it with a log warning, and four threw. The design-time pipeline therefore died on its first execution for the project and kept dying, because the throw happened inside a `WeakCache` factory.

## The model

A duplicated path is **one document included twice**, which is the conclusion the command-line compiler reaches. The alternative cannot be represented: every design-time consumer addresses a document by path, from the Roslyn analyzer callback that carries only a `SemanticModel` to the cross-process preview contract, so a second document at the same path could never be asked for.

## What changed

**Resolved once, at the boundary.** `RemoveDuplicatePathSyntaxTrees` drops every tree whose path an earlier tree already holds, applied in the diff layer and in `PartialCompilation.CreateComplete`/`CreatePartial`. It returns its argument without allocating when nothing is duplicated, which is every compilation the command-line compiler produces.

This matters more than making the index tolerant. `LinkerInjectionStep` rewrites the trees the index exposes; a tree left in the compilation but absent from the index keeps its declarations in the code model and is never rewritten, so aspects would apply to some declarations of the project and silently not to others. `EverySyntaxTreeIsRewritten` pins that.

**Reported as `LAMA0307`**, once, on the file that keeps the path. The condition costs the user a declaration, so it is not merely resolved.

**Two keys for two lifetimes.** Within a compilation, membership and enumeration are keyed by syntax tree identity. Across compilations, every design-time cache is keyed by the new `DocumentKey`: ordinal comparison baked in, hash cached on construction, and distinct from the introduced-tree and hint-name key spaces, which is where the third crash came from.

**An ordering bug the tests uncovered.** `DesignTimeAspectPipelineResult.Update` interleaved un-indexing an old result with indexing a new one, per tree. Introduced trees are keyed by a name rendered from the target type while the results carrying them are keyed by a source path, so when a partial type's primary declaration moves between files the old owner's un-index deleted the entry the new owner had just written. The pass is now two-phase.

**Incidental fixes.** `GetClosure.AddTree` was quadratic in the tree count. `PartialCompilation.Update` resolved the replaced tree by path while `Compilation.ReplaceSyntaxTree` resolves it by identity. `TransformedPathGenerator` threw on a collision of `(file name, low 32 bits of the content hash)` with a message asserting the two files held identical code, which is wrong in the case that reaches production; it now resolves the collision with a deterministic ordinal.

## API

`IPartialCompilation` gains `[InternalImplement]`, which it should always have had: only `PartialCompilation` implements it and weavers only consume it.

| Member | State |
|---|---|
| `SyntaxTrees` | `[Obsolete]`, unchanged in type and behaviour |
| `SyntaxTreeCollection` | new, `IReadOnlyCollection<SyntaxTree>` |
| `TryGetSyntaxTree( DocumentKey, out SyntaxTree? )` | new |

Every call site in the repository is migrated. `Metalama.Framework.DesignTime.Contracts` is untouched.

## Tests

18 new tests across `DuplicateSyntaxTreePathTests`, `IntroducedSyntaxTreeOwnershipTests` and `TransformedPathGeneratorTests`, written red first: all three exception texts quoted in the issue were reproduced verbatim before the fix.

| Suite (net8.0) | Result |
|---|---|
| `Tests.UnitTests` | 2243 passed, 0 failed |
| `Tests.AspectTests` | 2271 passed, 0 failed |
| `Tests.TemplateTests` | 317 passed, 0 failed |
| `Tests.LinkerTests` | 189 passed, 0 failed |

Whole solution builds warning-free on all target frameworks.

— Claude for @gfraiteur

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Dependencies/DependencyCollectorByDependentSyntaxTreeAndMasterProject.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Dependencies/DependencyCollectorByDependentSyntaxTreeAndMasterProject.cs
@@ -2,8 +2,8 @@
 // SharpCrafters s.r.o. licenses this file to you under either the MIT license or a proprietary license, depending on the repository from which it was obtained.
 // Refer to LICENSE.md in the repository root for complete details.
 
-using Metalama.Framework.Engine.CodeModel;
 using Metalama.Framework.Engine;
+using Metalama.Framework.Engine.CodeModel;
 
 namespace Metalama.Framework.DesignTime.Pipeline.Dependencies;
 
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Dependencies/DependencyCollectorByDependentSyntaxTreeAndMasterProject.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Dependencies/DependencyCollectorByDependentSyntaxTreeAndMasterProject.cs
@@ -2,6 +2,7 @@
 // SharpCrafters s.r.o. licenses this file to you under either the MIT license or a proprietary license, depending on the repository from which it was obtained.
 // Refer to LICENSE.md in the repository root for complete details.
 
+using Metalama.Framework.Engine.CodeModel;
 using Metalama.Framework.Engine;
 
 namespace Metalama.Framework.DesignTime.Pipeline.Dependencies;
@@ -11,24 +12,24 @@ namespace Metalama.Framework.DesignTime.Pipeline.Dependencies;
 /// </summary>
 internal sealed class DependencyCollectorByDependentSyntaxTreeAndMasterProject
 {
-    private readonly Dictionary<string, ulong> _masterFilePathsAndHashes = new( StringComparer.Ordinal );
+    private readonly Dictionary<DocumentKey, ulong> _masterFilePathsAndHashes = new();
     private readonly HashSet<TypeDependencyKey> _masterPartialTypes = new();
     private int _hashCode;
 
-    public string DependentFilePath { get; }
+    public DocumentKey DependentFilePath { get; }
 
-    public IReadOnlyDictionary<string, ulong> MasterFilePathsAndHashes => this._masterFilePathsAndHashes;
+    public IReadOnlyDictionary<DocumentKey, ulong> MasterFilePathsAndHashes => this._masterFilePathsAndHashes;
 
     public IReadOnlyCollection<TypeDependencyKey> MasterPartialTypes => this._masterPartialTypes;
 
     public bool Contains( TypeDependencyKey type ) => this._masterPartialTypes.Contains( type );
 
-    public DependencyCollectorByDependentSyntaxTreeAndMasterProject( string dependentFilePath )
+    public DependencyCollectorByDependentSyntaxTreeAndMasterProject( DocumentKey dependentFilePath )
     {
         this.DependentFilePath = dependentFilePath;
     }
 
-    public void AddSyntaxTreeDependency( string masterFilePath, ulong masterHash )
+    public void AddSyntaxTreeDependency( DocumentKey masterFilePath, ulong masterHash )
     {
 #if DEBUG
         if ( this._isReadOnly )
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Dependencies/DependencyCollectorByDependentSyntaxTreeAndMasterProject.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Dependencies/DependencyCollectorByDependentSyntaxTreeAndMasterProject.cs
@@ -12,24 +12,24 @@ namespace Metalama.Framework.DesignTime.Pipeline.Dependencies;
 /// </summary>
 internal sealed class DependencyCollectorByDependentSyntaxTreeAndMasterProject
 {
-    private readonly Dictionary<DocumentKey, ulong> _masterFilePathsAndHashes = new();
+    private readonly Dictionary<DocumentKey, ulong> _masterDocumentKeysAndHashes = new();
     private readonly HashSet<TypeDependencyKey> _masterPartialTypes = new();
     private int _hashCode;
 
-    public DocumentKey DependentFilePath { get; }
+    public DocumentKey DependentDocumentKey { get; }
 
-    public IReadOnlyDictionary<DocumentKey, ulong> MasterFilePathsAndHashes => this._masterFilePathsAndHashes;
+    public IReadOnlyDictionary<DocumentKey, ulong> MasterDocumentKeysAndHashes => this._masterDocumentKeysAndHashes;
 
     public IReadOnlyCollection<TypeDependencyKey> MasterPartialTypes => this._masterPartialTypes;
 
     public bool Contains( TypeDependencyKey type ) => this._masterPartialTypes.Contains( type );
 
-    public DependencyCollectorByDependentSyntaxTreeAndMasterProject( DocumentKey dependentFilePath )
+    public DependencyCollectorByDependentSyntaxTreeAndMasterProject( DocumentKey dependentDocumentKey )
     {
-        this.DependentFilePath = dependentFilePath;
+        this.DependentDocumentKey = dependentDocumentKey;
     }
 
-    public void AddSyntaxTreeDependency( DocumentKey masterFilePath, ulong masterHash )
+    public void AddSyntaxTreeDependency( DocumentKey masterDocumentKey, ulong masterHash )
     {
 #if DEBUG
         if ( this._isReadOnly )
@@ -37,16 +37,16 @@ internal sealed class DependencyCollectorByDependentSyntaxTreeAndMasterProject
             throw new InvalidOperationException();
         }
 #endif
-        lock ( this._masterFilePathsAndHashes )
+        lock ( this._masterDocumentKeysAndHashes )
         {
-            if ( !this._masterFilePathsAndHashes.TryGetValue( masterFilePath, out var existingHash ) )
+            if ( !this._masterDocumentKeysAndHashes.TryGetValue( masterDocumentKey, out var existingHash ) )
             {
-                this._masterFilePathsAndHashes.Add( masterFilePath, masterHash );
-                this._hashCode ^= HashCode.Combine( masterFilePath, masterHash );
+                this._masterDocumentKeysAndHashes.Add( masterDocumentKey, masterHash );
+                this._hashCode ^= HashCode.Combine( masterDocumentKey, masterHash );
             }
             else if ( existingHash != masterHash )
             {
-                throw new AssertionFailedException( $"Hashes '{existingHash}' and '{masterHash}' do not match for '{masterFilePath}'." );
+                throw new AssertionFailedException( $"Hashes '{existingHash}' and '{masterHash}' do not match for '{masterDocumentKey}'." );
             }
         }
     }
@@ -86,15 +86,15 @@ internal sealed class DependencyCollectorByDependentSyntaxTreeAndMasterProject
         }
 
         if ( this._hashCode != other._hashCode
-             || this._masterFilePathsAndHashes.Count != other._masterFilePathsAndHashes.Count
+             || this._masterDocumentKeysAndHashes.Count != other._masterDocumentKeysAndHashes.Count
              || this._masterPartialTypes.Count != other._masterPartialTypes.Count )
         {
             return false;
         }
 
-        foreach ( var dependency in this._masterFilePathsAndHashes )
+        foreach ( var dependency in this._masterDocumentKeysAndHashes )
         {
-            if ( !other._masterFilePathsAndHashes.TryGetValue( dependency.Key, out var otherHash ) || otherHash != dependency.Value )
+            if ( !other._masterDocumentKeysAndHashes.TryGetValue( dependency.Key, out var otherHash ) || otherHash != dependency.Value )
             {
                 return false;
             }
--- a/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Dependencies/DependencyGraphByDependentProject.cs
+++ b/Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Dependencies/DependencyGraphByDependentProject.cs
@@ -13,10 +13,10 @@ namespace Metalama.Framework.DesignTime.Pipeline.Dependencies;
 /// </summary>
 internal readonly struct DependencyGraphByDependentProject
 {
-    private static readonly ImmutableDictionary<DocumentKey, DependencyGraphByMasterSyntaxTree> _emptyDependenciesByMasterFilePath =
+    private static readonly ImmutableDictionary<DocumentKey, DependencyGraphByMasterSyntaxTree> _emptyDependenciesByMasterDocumentKey =
         ImmutableDictionary<DocumentKey, DependencyGraphByMasterSyntaxTree>.Empty;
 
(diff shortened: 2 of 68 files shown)


Metalama.Framework/src/Metalama.Framework.Sdk/CodeModel/IPartialCompilation.cs
```diff
@@ -49,8 +53,30 @@ public interface IPartialCompilation
...
+    /// <summary>
+    /// Gets the syntax tree of the current subset that represents a given document, and returns
+    /// <see langword="false"/> if there is none.
+    /// </summary>
+    /// <remarks>
+    /// At most one syntax tree of an <see cref="IPartialCompilation"/> represents a document. Roslyn does not guarantee
+    /// that of a <see cref="Microsoft.CodeAnalysis.Compilation"/>, so Metalama establishes it: a tree whose path an
+    /// earlier tree already holds is removed from the compilation before the subset is built. See issue #1742.
+    /// </remarks>
+    bool TryGetSyntaxTree( DocumentKey documentKey, [NotNullWhen( true )] out SyntaxTree? syntaxTree );
```
gfraiteur, line 78: `TryGetValue( DocumentKey.FromPath( ...` is being used too often. Add extension method TryGetValueByPath

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Dependencies/DependencyCollectorByDependentSyntaxTreeAndMasterProject.cs
```diff
@@ -11,24 +12,24 @@ namespace Metalama.Framework.DesignTime.Pipeline.Dependencies;
 /// </summary>
 internal sealed class DependencyCollectorByDependentSyntaxTreeAndMasterProject
 {
-    private readonly Dictionary<string, ulong> _masterFilePathsAndHashes = new( StringComparer.Ordinal );
+    private readonly Dictionary<DocumentKey, ulong> _masterFilePathsAndHashes = new();
     private readonly HashSet<TypeDependencyKey> _masterPartialTypes = new();
     private int _hashCode;
 
-    public string DependentFilePath { get; }
+    public DocumentKey DependentFilePath { get; }
```
gfraiteur: Rename property to DependentDocumentKey

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Dependencies/DependencyGraphByDependentProject.cs
```diff
@@ -12,24 +13,24 @@ namespace Metalama.Framework.DesignTime.Pipeline.Dependencies;
 /// </summary>
 internal readonly struct DependencyGraphByDependentProject
 {
-    private static readonly ImmutableDictionary<string, DependencyGraphByMasterSyntaxTree> _emptyDependenciesByMasterFilePath =
-        ImmutableDictionary<string, DependencyGraphByMasterSyntaxTree>.Empty.WithComparers( StringComparer.Ordinal );
+    private static readonly ImmutableDictionary<DocumentKey, DependencyGraphByMasterSyntaxTree> _emptyDependenciesByMasterFilePath =
+        ImmutableDictionary<DocumentKey, DependencyGraphByMasterSyntaxTree>.Empty;
```
gfraiteur, line 17: Rename to _emptyDependenciesByMasterDocumentKey

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Dependencies/DependencyGraphByDependentProject.cs
```diff
@@ -12,24 +13,24 @@ namespace Metalama.Framework.DesignTime.Pipeline.Dependencies;
...
 
-    private static readonly ImmutableDictionary<string, DependencyCollectorByDependentSyntaxTreeAndMasterProject> _emptyDependenciesByDependentFilePath =
-        ImmutableDictionary<string, DependencyCollectorByDependentSyntaxTreeAndMasterProject>.Empty.WithComparers( StringComparer.Ordinal );
+    private static readonly ImmutableDictionary<DocumentKey, DependencyCollectorByDependentSyntaxTreeAndMasterProject> _emptyDependenciesByDependentFilePath =
```
gfraiteur: Also rename

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/PartialCompilation.CompleteImpl.cs
```diff
@@ -30,8 +31,21 @@ private CompleteImpl(
...
             [Memo]
-            public override ImmutableDictionary<string, SyntaxTree> SyntaxTrees => this.Compilation.GetIndexedSyntaxTrees();
+            [Obsolete( "Use SyntaxTreeCollection to enumerate the syntax trees, or TryGetSyntaxTree to find one by its DocumentKey." )]
+            public override ImmutableDictionary<string, SyntaxTree> SyntaxTrees
+                => this.Compilation.GetIndexedSyntaxTrees().ToImmutableDictionary( x => x.Key, x => x.Value, StringComparer.Ordinal );
+
+            public override IReadOnlyCollection<SyntaxTree> SyntaxTreeCollection => this.Compilation.GetSyntaxTreeIndex().SyntaxTrees;
+
+            public override bool TryGetSyntaxTree( DocumentKey documentKey, [NotNullWhen( true )] out SyntaxTree? syntaxTree )
+                => this.Compilation.GetIndexedSyntaxTrees().TryGetValue( documentKey.Path, out syntaxTree );
```
gfraiteur: GetIndexedSyntaxTrees should return the dictionary keyed by DocumentKey for optimal performance.

Metalama.Framework/src/Metalama.Framework.Engine/Utilities/Roslyn/SyntaxTreeIndex.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// <summary>
+    /// Gets the syntax trees of the compilation indexed by path, holding the first tree of each path.
+    /// </summary>
+    /// <remarks>
+    /// A <see cref="Dictionary{TKey,TValue}"/> and not an <see cref="ImmutableDictionary{TKey,TValue}"/>. The index is
+    /// built once per compilation and never updated, so the structural sharing an immutable dictionary provides buys
+    /// nothing, while its cost is paid on every build and every lookup: an ordered tree walk instead of a hash lookup,
+    /// and two allocations per entry instead of one array.
+    /// </remarks>
+    public IReadOnlyDictionary<string, SyntaxTree> SyntaxTreesByPath => this._syntaxTreesByPath;
```
gfraiteur: should be indexed by DocumentKey

Metalama.Framework/src/Metalama.Framework.Sdk/CodeModel/IPartialCompilation.cs
gfraiteur: Done in 47ded3e, as `TryGetSyntaxTreeByPath` per your correction: an extension on `IPartialCompilation`, not a generic dictionary one.

It now covers the four sites where a path genuinely arrives as a string, all of them boundaries: `HtmlCodeWriter` pairing the input and output compilations, and the preview RPC. The two `DocumentKey.FromPath` calls left in production are dictionary lookups from Roslyn analyzer callbacks, which is exactly the boundary the type exists to mark.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Dependencies/DependencyCollectorByDependentSyntaxTreeAndMasterProject.cs
gfraiteur: Done in 47ded3e: `DependentDocumentKey`.

Applied to the rest of the family too, since they are one key space: `MasterFilePathsAndHashes` to `MasterDocumentKeysAndHashes`, `DependentFilePaths` to `DependentDocumentKeys`, the `dependentFilePath` and `masterFilePath` parameters, and the components of `SyntaxTreeDependency` and `PartialTypeDependency`.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Dependencies/DependencyGraphByDependentProject.cs
gfraiteur: Done in 47ded3e: `_emptyDependenciesByMasterDocumentKey`, along with the `DependenciesByMasterFilePath` property it backs.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.DesignTime/Pipeline/Dependencies/DependencyGraphByDependentProject.cs
gfraiteur: Done in 47ded3e: `_emptyDependenciesByDependentDocumentKey`, and `DependenciesByDependentFilePath` with it.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/PartialCompilation.CompleteImpl.cs
gfraiteur: Done in 47ded3e. `GetIndexedSyntaxTrees` returns `IReadOnlyDictionary<DocumentKey, SyntaxTree>`, so `TryGetSyntaxTree` is a direct lookup with no `.Path` unwrapping.

`PartialImpl` stores its subset the same way, which leaves the obsolete `SyntaxTrees` property as the only place that materializes a string-keyed dictionary, and nothing reads it.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/Utilities/Roslyn/SyntaxTreeIndex.cs
gfraiteur: Done in 47ded3e: `SyntaxTreesByDocumentKey`, a `Dictionary<DocumentKey, SyntaxTree>` built with `syntaxTree.GetDocumentKey()`.

Beyond consistency this is the faster arrangement, which is the point you make on `CompleteImpl`: a lookup made from a key reuses the hash the key computed on construction, instead of rehashing a long path string on every call.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Sdk/CodeModel/DocumentKey.cs
```diff
@@ -0,0 +1,162 @@
...
+        if ( path == null )
+        {
+            documentKey = default;
+
+            return false;
+        }
+
+        documentKey = DocumentKey.FromPath( path );
+
+        return true;
```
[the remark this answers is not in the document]
gfraiteur: Valid, and the method also had no caller, so it is removed in c08c7ff rather than corrected.

On the detail: Roslyn documents `FileLinePositionSpan.Path` as null for an invalid span and as `string.Empty` when the path is not available, so both are reachable and only the first was tested. `Location.None` yields the null case, but an external location or a syntax tree with an empty `FilePath` yields the empty one, and that path fell through to `DocumentKey.Compilation` with a `true` result. The XML doc promised the opposite.

It was added speculatively, on the assumption that mapping a diagnostic location to a document key would be wanted. It was not: `SplitResultsByTree` reads `Location.GetLineSpan().Path` into a local string-keyed dictionary that lives for the duration of one call, which is not a key space `DocumentKey` is meant to police. An unused public API with a contract that contradicts its behaviour is worse than no API, so it goes.

— Claude for @gfraiteur

Issue #1797, opened 2026-08-03
Design-time: a transitive aspect instance pins the compilation it was produced in


gfraiteur:
> [!NOTE]
> **Rewritten and closed on 2026-08-04.** This issue was rewritten twice as the investigation corrected itself. Both
> earlier diagnoses were wrong, and both are named below rather than deleted, because each was asserted from reading
> the code and each cost time. The comments preserve the original text.

## What was wrong

A `TransitiveAspectInstance` produced by `PullStrategy.IntroduceParameterAndPull` is stored in the design-time pipeline
results and carried forward across every edit, so the live references it held pinned the Roslyn `Compilation` it was
produced in for the whole editing session. One compilation per project, for as long as the project stayed open.

## Fixed

| Change | Commit |
|---|---|
| `TransitiveAspectInstance.TargetDeclaration` made durable | `34fad4d285` |
| `PullConstructorParameterTransitiveAspect._parameter` and the `IType` references of the pull strategies made durable | `7b93261c3c` |

`TransitiveContributorMemoryLeakTests` asserted the retention and instructed that the assertion be inverted once it was
closed. It is now a `MemoryLeakAssert.Collected` and passes.

## Two defects this exposed, both fixed

**A durable reference dropped the type arguments of a constructed generic type.** `DurableRefFactory.FromDeclarationOrType`
matched `IDeclaration` before `IType`, and an `INamedType` is both, so `ICovariant<Derived>` was encoded as a
declaration identifier, which names the generic definition, and resolved back as `ICovariant<T>`. `IRef.ToDurable` had
the same defect by a different route. Both now use a `SerializableTypeId` for a constructed generic type, and a
canonical generic instance keeps the declaration identifier so the two routes still agree. `844e16933b`, `4d107d00b6`.

The widening was silent and produced a usable type rather than an error, which is why it surfaced only as a spurious
`T` in generated code. It broke the constructor parameter pull, which compares the type of the parameter it introduced
against the type it is asked to introduce. `RefTests.DurableRefToConstructedGenericTypeLosesTheTypeArguments` and its
`ToDurableRef` counterpart recorded the widening as intended behaviour; both are corrected to assert that the type
arguments are kept.

**A constructor identifier naming an already-introduced parameter matched nothing.** The identifier describes the
signature as it stood when it was written, which may already include a parameter an earlier aspect introduced.
`DocumentationIdHelper` compared it against the source parameters only and required the counts to be equal, so a
[...]

PR #1798, opened 2026-08-03
Design-time: stop accumulating contributors, and stop pinning compilations (#1796, #1799, part of #1797)

### LINKED ISSUE #1796: Design-time: cross-project validators accumulate in DesignTimeAspectPipelineResult.Extensions on every pipeline run
Split off from #1793, which fixed the three retention paths that hold a Roslyn `Compilation`. This one is a different
kind of accumulation: unbounded in count, but each entry holds a declaration identifier rather than a compilation.

`DesignTimeAspectPipelineResult.Update` appends the `externalExtensions` bucket to the accumulated `Extensions`
collection on every run of the pipeline:

`Metalama.Framework.DesignTime/Pipeline/DesignTimeAspectPipelineResult.cs:303-311`

```csharp
if ( externalExtensions != null )
{
    extensionsBuilder ??= this.Extensions.ToBuilder();

    foreach ( var externalExtension in externalExtensions )
    {
        extensionsBuilder.Add( externalExtension );
    }
}
```

A contributor reaches that bucket when its syntax tree lies outside the partial compilation that the pipeline just
analysed, which the code at lines 498-508 notes happens with cross-project validators:

```csharp
else
{
    // This happens with cross-project validators
    externalValidators ??= new List<IDesignTimePipelineResultExtension>();
    externalValidators.Add( designTimeExtension );
}
```

No removal path can reach those entries. `UnindexOldTree` at lines 184-195 removes only the contributors that were
filed under a `SyntaxTreePipelineResult`, and an external contributor was never filed under any. The `Invalidator`
passes `Extensions` through unchanged, and the only reset of the accumulated result requires a compile-time change,
so a session of run-time-only edits never clears it.

Two consequences:

- The collection grows by one entry per pipeline run, for the lifetime of the process. Each entry is a
  `DesignTimeReferenceValidatorInstance` holding a `SerializableDeclarationId`, a cached driver, the user validator
  object and an `IAspectState`, so the retained memory is a few hundred bytes per run rather than the tens of
  megabytes that a retained compilation costs.
- `DesignTimeAspectPipelineResultExtensionCollection.Builder.Add` also duplicates the key into `_validatorsBuilder`
  (`DesignTimeAspectPipelineResultExtensionCollection.cs:120-128`), so `GetValidatorsForSymbol` returns one copy per
  run. The validation work, and presumably the diagnostics it produces, therefore grow with the length of the editing
  session. This is likely to be more noticeable than the memory.

Only solutions that carry cross-project validators are affected.

## Why this was not covered by the tests added in #1793

Reproducing it requires a project whose reference carries a transitive validator. The unit test project has no
fixture for that, so the suite added in #1793 does not exercise this path. A test should assert that, after N runs of
the pipeline over a project with such a reference and with only run-time code changing between runs, both
`pipeline.AspectPipelineResult.Extensions.Extensions.Length` and the number of validators returned by
`GetValidatorsForSymbol` stay bounded by the number of distinct validators rather than growing with N.

— Claude for @gfraiteur

### LINKED ISSUE #1799: Design-time: four retention defects on the extension and fabric surface, with red reproduction tests
Following [#1793](https://github.com/metalama/Metalama/issues/1793), I audited the surface that the extension packages
(`Metalama.Extensions.Validation`, `Metalama.Extensions.Architecture`, `Metalama.Extensions.CodeFixes`) actually use at
design time. Four defects survived adversarial verification, and each now has a reproduction test in
`Metalama.Framework.Tests.UnitTests/DesignTime/`. **All four tests are red.**

Every one of them is a case of the same rule, documented in `Metalama.Framework/docs/design-time-memory.md`: an object
that outlives a single request must not reach a `Compilation`. What is new here is *which* objects outlive a request.
Three collections do so without looking as though they did, and the document has been extended to say so.

---

## 1. The pipeline configuration retains the version it was built from, for the whole session

`AspectPipelineConfiguration` is built once and reused for every subsequent version of the project; it is discarded
only when the compile-time code changes. In a session in which the user edits run-time code alone it is therefore the
configuration built from the **first** version that stays in use, and the version it pins is the oldest of the session,
held in addition to the current one.

`FabricsContributors` reaches the amender of every static fabric, and the amender holds two independent references to
the compilation it was built from:

- `FabricDriver.BaseAmender<T>.UserCodeExecutionContext` (`FabricDriver.BaseAmender.cs:71`), which reaches
  `UserCodeExecutionContext.Compilation : CompilationModel`. The very next line, `:72`, is
  `this.TargetDeclaration = targetDeclaration.ToDurable();`, so the discipline is applied one line away and not here.
- `FabricInstance.TargetDeclaration` (`FabricInstance.cs:32`), which is `targetDeclaration.ToRef()` rather than
  `ToDurable()`, giving a `SymbolRef` whose `RefFactory` reaches `CompilationContext.Compilation`.

**These must be fixed together.** Fixing either alone releases nothing, because the other still pins the same version.
A third, smaller edge survives both: `FabricInstance.Driver` reaches `FabricDriver.DiagnosticLocation`, a Roslyn
`Location`, which pins the fabric file's `SyntaxTree` (though not its compilation).

Measured retention path:

```
DesignTimeAspectPipeline._currentState.PipelineResult
  -> Configuration : AspectPipelineConfiguration
    -> FabricsContributors -> _contributors[0] : AspectQuerySource<IDeclaration>
      -> _query : ChildQuery<IMethod,Int32> -> Owner : Amender
        -> UserCodeExecutionContext -> CompilationContext -> Compilation
```

Scope: static fabrics (`ProjectFabric`, `NamespaceFabric`) that register at least one query-backed contributor.
`TypeFabricDriver` builds its amender per execution and is not affected.

Test: `MemoryLeaks/FabricMemoryLeakTests`. `WithFabric_InitialCompilationIsCollected` is red;
`WithoutFabric_InitialCompilationIsCollected` is the control over the identical project minus the fabric file and
passes, and the two `SurvivingCompilationCountDoesNotGrow` cases pass, which establishes that the cost is one version
rather than growth. The memory suite had no fabric coverage at all before this, which is why this was invisible.

## 2. Nothing states that what an extension stores must be durable

`SyntaxTreePipelineResult` describes itself as "compilation-independent and cacheable", and the pipeline relies on it:
`Update` carries forward the entry of every file the run did not analyse, and `SplitResultsByTree` deliberately
discards a freshly produced contributor whose file is not dirty so that the earlier instance survives (#1768). A
contributor filed under a file the user never edits therefore keeps the version in which that file was last analysed.

The engine's own half is careful: `SymbolDictionaryKey.CreatePersistentKey` stores only a `SerializableDeclarationId`,
and `SymbolDictionaryKey.ToRef()` routes through `DurableRefFactory`. But the payload is opaque to the engine, and
neither `IDesignTimePipelineResultExtension` nor `ITransitivePipelineContributor.ToDesignTime()` says the payload must
be durable. An extension author has no way to learn the requirement except by measuring the analysis process.

The requirement is currently violated in Premium: `TypeEqualityPredicate._typeRefs`
(`Metalama.Extensions.Architecture/Predicates/TypeEqualityPredicate.cs:15`, filled at `:37` and `:56` with
`Definition.ToRef()`) holds symbol-backed references, and it is reachable from a stored design-time validator through
`DesignTimeReferenceValidatorInstance.Implementation` then `ValidatorImplementation.Implementation` then
`ReferencePredicateValidator._allowedScope`. It is the only shipped Architecture predicate that holds an `IRef`; the
name-based and access-based predicates hold strings and enums.

Proposed framework-side fix: state the requirement on `IDesignTimePipelineResultExtension` and on
`ITransitivePipelineContributor.ToDesignTime()`, and add a DEBUG invariant in `DesignTimeAspectPipelineResult.Update`
asserting that no stored extension reaches a non-durable `IFullRef`.

Test: `MemoryLeaks/ExtensionContributorMemoryLeakTests`, a matched pair over one test-only `PipelineExtension`
registered through `ITestExtensionCollector`, differing only in whether the contributor's reference is made durable.
The durable case passes and the non-durable case is red, so the pair discriminates. No Premium dependency.

## 3. A diagnostic whose argument is a declaration pins the compilation

This one was not in my original audit; I found it while building the test for defect 2, because it made the positive
control fail.

`DiagnosticDefinition` formats lazily, so the arguments given to `WithArguments` are stored in the
`DiagnosticDescriptor` of the resulting diagnostic and held for as long as the diagnostic is. The diagnostics of a file
live in that file's `SyntaxTreePipelineResult` and are carried forward exactly as in defect 2. An argument that is an
`IDeclaration` therefore reaches `CompilationModel`, then `CompilationContext`, then `Compilation`:

```
DesignTimeAspectPipelineResult.SyntaxTreeResults[...] : SyntaxTreePipelineResult
  -> Diagnostics[0] : SimpleDiagnostic
    -> _descriptor.Title : NonLocalizedString._arguments[0] : SourceMethod
      -> Compilation : CompilationModel -> CompilationContext -> Compilation
```

I think this is the most consequential of the four, because passing the declaration is the obvious way to write the
message, it is what our own documentation and samples do, and it is what reference validation does on every reference
it rejects.

Measured cost: **one** retained version per project, not growth.
`DiagnosticWithDeclarationArgument_RetentionDoesNotGrowWithTheNumberOfEditedFiles` passes at 4 and at 10 distinct
edited files. Editing different files in turn was the arrangement most likely to accumulate versions, and it does not.

Test: `MemoryLeaks/DiagnosticArgumentMemoryLeakTests`, a matched pair differing only in whether the argument is `m` or
`m.Name`. The string case passes; the declaration case is red.

## 4. `RpcService.WaitUntilInitializedAsync` does not observe its cancellation token

```csharp
protected Task WaitUntilInitializedAsync( CancellationToken cancellationToken = default )
    => this._initialized.Task.WarnIfLongAsync( this.Logger, nameof(this.WaitUntilInitializedAsync), cancellationToken );
```

`_initialized` is completed only by `OnRpcConnected`, when a client attaches. Neither path of `WarnIfLongAsync` honours
the token: with warnings off it returns the task unchanged (`LongTaskHelper.cs:120-126`), and with warnings on
`WarnIfTaskLongCoreAsync` awaits `task` bare at `:165`, the token reaching only the `Task.Delay`.

The asymmetry shows it is an oversight: the client-side counterpart `RpcClient.WaitUntilInitializedAsync`
(`RpcClient.cs:62`) composes with `WithCancellation`, and is covered by
`RpcClientTests.WaitUntilInitializedAsync_Cancelled_ThrowsOperationCancelledException`, which passes.

Every server-side service method awaits this first, including `RpcService{TApi}.RaiseEventAsync`, which awaits it
*before* checking whether there is any client to raise the event to. A parked frame retains everything it captured,
and on this surface those captures are Roslyn objects: the source generator parks while holding the `Compilation` it
is publishing generated sources for, and the user-process CodeLens and Aspect Explorer services park while holding a
`Compilation` and an `ISymbol`. The configuration in which this bites is an analysis process no client attaches to,
which is what happens when the Visual Studio extension is not installed.

Measured retention path:

```
TestService._initialized : TaskCompletionSource<Boolean>
  -> _task.m_continuationObject[0] : AsyncStateMachineBox<WarnIfTaskLongCoreAsync>
    -> ... -> the awaiting frame -> everything it captured
```

Suggested fix: `this._initialized.Task.WithCancellation( ct ).WarnIfLongAsync( ... )`, the composition that
`VsAnalysisProcessProjectSourceGenerator.cs:64` already uses.

Extensions reach this surface: an extension may register RPC services during initialization, as
`DesignTimeExtensionRpcTests` demonstrates, and those services inherit the same wait.

Test: `Rpc/RpcServiceCancellationTests`, two red tests: one that the wait ends when its token is cancelled, one that
cancelling releases what the parked frame held. Both are bounded by a shortened test-context timeout so that the
failure is reported rather than hanging.

---

## Checked and clean

So that the audit is falsifiable, these were examined and found correct: the ephemeron-based caches
(`DesignTimeAspectPipeline._compilationResultCache`, `CompilationContextFactory._instances`,
`SemanticModelProvider._instances`, `SymbolId._cache`, `QueryExecutionContext._staticCache`); the path-keyed and
value-only structures (`SymbolDictionaryKey.CreatePersistentKey`, `ReferenceIndexerOptions`,
`ReferenceIndexerRequirements`); the one-generation bounds (`_foreignExtensions`, whose `Update` replaces rather than
appends, `_invalidSyntaxTreeResults`, and `_extensionsFromProjectReferences`, rebuilt on every `Update`); the
`PipelineExtension` implementations themselves, none of which has a single instance field, so no extension stores the
`CompilationModel` it is handed; the event bookkeeping in `AnalysisProcessEventHub`, which uses `WeakEvent` for its
high-traffic events and where every subscriber unsubscribes; and the serialization path, which rebuilds every
reference through `DurableRefFactory`.

Premium's `TransitiveValidatorInstance` does use a non-durable `ToRef()`, but its only caller is
`ValidationPipelineExtension.GetTransitiveManifestExtensions`, reached only from `CompileTimeAspectPipeline`, which is
compile time only, so it is not a design-time leak.

## Verification

Full unit suite on `net8.0`: 2269 passed, 5 failed, and the 5 are exactly the new red tests. CI-mode build
(`-p:ContinuousIntegrationBuild=true`) reports 0 errors.

— Claude for @gfraiteur

gfraiteur:
Design-time retention work across three issues. The branch began as the fix for #1796 and grew as the audit behind it
found more; the commits are separated by defect so they can be read independently.

## #1796 — cross-project contributors accumulated on every run

`SplitResultsByTree` files each transitive pipeline contributor under the document of its syntax tree. At design time
the pipeline runs on the dirty trees only, so a contributor belonging to a tree that is not dirty had no result to be
filed under, and went to a bucket that `Update` appended to the accumulated `Extensions` collection on every run.
Nothing could remove those entries, so the collection gained one per run for the lifetime of the process, and because
each entry also indexes itself by validated declaration, a lookup returned one copy per run and the validation work
grew with the length of the editing session.

Two cases had to be told apart, which the first attempt did not do:

- **an own tree that is merely not dirty** is skipped, because the tree keeps the result of the run that did include
  the contributor and that result already carries it (#1768);
- **a foreign tree**, which is what a cross-project validator reports, is kept, in a set that `Update` replaces on
  every run rather than appending to. Replacing is lossless because a run re-produces the complete set.

Skipping both, which the first attempt did, silently dropped cross-project validators. `TransitiveValidationSource` in
Metalama.Premium is the producer, and `SideBySideVersionTests.TransitiveValidator` there is what caught it.

**Replacing that set unconditionally was itself a defect**, found in review. A run can produce no foreign contributor
for reasons that say nothing about whether the rules still apply: `DesignTimePipelineStage` invokes the extensions only
when the run produced at least one contributor of extension kind, and `PipelineState` reads `TransitiveContributors`
from a result that is null when the pipeline failed. The set is now replaced only when the run produced some. The
residual is documented where the decision is made: a rule genuinely removed survives until the next run producing any
foreign contributor, and removing a rule is a compile-time edit, which discards the configuration and this result with
it.

## #1799 — retention defects on the extension, fabric and query surface

`AspectPipelineConfiguration` is long-lived by design, reused across keystrokes because rebuilding it per keystroke
would mean recompiling the compile-time assemblies each time. That reuse stays. What follows from it is a constraint:
it must not pin a compilation, because in a session of run-time edits alone it would retain the *first* version of the
session, held in addition to the current one.

| Commit | Defect |
|---|---|
| `030e70f` | A diagnostic reported on a declaration pinned the compilation: `DiagnosticDefinition` formats lazily, so the argument is held by the stored diagnostic. |
| `24e9061`, `737bce8` | Three waits on a `TaskCompletionSource` did not observe their cancellation token, so a caller parked while holding a `Compilation` was never released. |
| `f74f2e1` | The pipeline configuration pinned a compilation by three routes. |
| `8a14020` | Nothing stated that what an extension stores in a design-time contributor must be durable. |
| `ba5ec81` | The durable form chosen for a query base type lost the type arguments of a constructed generic type, and could not denote an introduced type at all. |
| `f72ef24` | A context built on demand inherited from whichever ambient context happened to be current. |

Points a reviewer should look at rather than take on trust.

**The diagnostic fix materializes selectively.** `MetalamaStringFormatter` passes the composite format specifier to an
`IFormattable` argument, so pre-formatting every argument would change `{0:N2}` and make `{0:x}` throw. Only the
compilation-bound arguments are materialized, and every formatter branch that handles one ignores the specifier, so no
message can change. `DiagnosticArgumentMaterializationTests` asserts that per category of argument, and the two
specifiers above are the cases that discriminate. The cost was measured, because it is paid on the compile-time path
too: about 3.2 µs to create a diagnostic whose argument is a declaration, of which about 2.3 are the display string,
against about 0.25 with a string argument. Ten thousand never-displayed diagnostics would cost about 20 ms, which does
not justify gating on the execution scenario, which this static extension has no way of reading.

**`IQueryOwner` gained a factory in place of a property.** A fabric amender no longer stores a
`UserCodeExecutionContext` at all; it keeps the compilation-neutral `UserCodeDescription` and builds a context per run.
This is deliberately structural: clearing the compilation from a stored context would leave a field that a later change
could repopulate. The two owners whose lifetime is a single run, the aspect builder and a type fabric's amender, return
the context they hold, rebound.

That change moved *when* the inheritance from the ambient context is sampled, which is a behavioural change and not
only a memory one. It used to happen once, while the fabric was running; it now happens on demand. `ToCollection` is
public, so a fabric query can be executed from inside an aspect and would acquire that aspect's target declaration and
meta API, the latter being what `WithCompilationAndDiagnosticAdder` asserts against carrying across compilations.
`UserCodeExecutionContext.CreateWithoutInheritance` is the shape for an owner that is not itself running user code.

**`SelectTypesDerivedFrom( INamedType )` needed a type identifier, not a declaration identifier.** The overload had no
test at all before this branch, here or in Metalama.Premium, where both fabric tests use the `System.Type` overload.
Making its base type durable with `ToDurable()` resolved `Base<int>` back to `Base<T>`, silently, because a
`SerializableDeclarationId` names a declaration: the query then matched the types derived from every construction of
the generic type. The durable form is a `SerializableTypeId`.

A type identifier resolves through the symbol table, so it cannot denote a type an aspect introduced, and converting
one unconditionally replaced a query that works with one that throws `SymbolNotFoundException` while it runs, far from
the call site. The conversion is therefore verified rather than assumed, and the type is kept as it is when no
identifier denotes it. Keeping it is safe here for a reason worth stating, because it is what bounds the retention:
only an aspect can hold an introduced type, the queries of an aspect do not outlive the run, and a fabric, whose
queries do, runs before any aspect and never sees one.

Both cases have an aspect test that fails on the previous conversion, and `RefTests` records the difference between the
two identifiers directly.

**One defect is documented rather than fixed.** What a contributor carries is opaque to the pipeline, so the durability
requirement can be stated but not enforced. Its test is a matched pair differing by one `ToDurable` call: the durable
case asserts release, the other asserts retention. The violation known at the time,
[`TypeEqualityPredicate`](https://github.com/metalama/Metalama.Premium/pull/74) in Metalama.Premium, is fixed there.

## #1797 — transitive aspect instances, half fixed

A `TransitiveAspectInstance` reached its compilation by two routes. The target declarations, its own and the one the
pull aspect records, are now durable, and `TransitiveAspectInstance.TargetDeclaration` is typed `IDurableRef<IDeclaration>`
so that the requirement is the compiler's business rather than the caller's discipline.

The second route stays open, and the reason is not the one previously given in that issue. An `IntroducedRef` **does**
have a serializable identifier: it overrides `ToSerializableId`, `FullRef.ToDurable` is built on it, and the identifier
of a pulled parameter names the constructor signature and the ordinal, which the resolution path handles explicitly and
`ConstructorParameterIdResolutionTests` covers. `ToDurable` therefore compiles on the parameter reference and closes
the retention, which was verified.

It also breaks the cross-project case: `PullParameterTests.CrossProjectIntegration` then fails, the consuming project
resolving nothing and the transitive aspect silently not applying. The identity of an introduced declaration is not
settled at the moment the advice runs, so making it durable has to happen later than that. That is a change of design
and is not attempted here.

## Testing

`MemoryLeakAssert.RetainedThrough` records a retention that is known and tracked but not yet fixed. Asserting liveness
alone would hold whether the documented route retains the object or something else does, and would go on holding after
that route was removed, so the assertion now requires the retention chain, which `RetentionPathFinder` already computes
for failures, to pass through a named type. The two controls on this branch, for the transitive aspect instance above
and for the non-durable extension payload, therefore fail in both directions that matter: when the defect is fixed,
which is the signal to replace the call with `Collected`, and when the object is retained for a different reason than
the one written down.

## Residuals, and three claims of mine that were wrong

`Metalama.Framework/docs/design-time-memory.md` is where the state of this work lives, because a pull request
description is not where the next person to work on it will look. It records what is open, what was positively
established as clean, and what was never examined.

Three conclusions I had recorded were subsequently disproved, and all three are kept, rewritten as the measurement, so
that the same wrong conclusion is not reached again from the same reading of the same code:

- **`WithCancellation` does not accumulate.** `Task.WhenAny` removes its continuation from the task that did not win.
  Measured on a task that is never completed, the count is the same small constant after 10, 100 and 1000 cancelled
  waits, for both implementations in use.
- **`TransitiveManifestDeserializationCache` is invalidated.** The wholesale replacement in `EnsureBoundTo` is not a
  substitute for missing eviction, it *is* the invalidation, keyed on the consuming project's `CompileTimeProject`.
- **`ToDurable()` cannot apply to an `IntroducedRef`** was false, as set out under #1797 above.

## Verification

Unit tests 2320 on `net8.0` and 2264 on `net48`, aspect tests 2273, all with 0 failures.
`PullParameterTests.CrossProjectIntegration`, which runs on `net48` only, is included in the net48 figure.

Each project this branch modifies builds with 0 errors under `-p:ContinuousIntegrationBuild=true`. Building the whole
`.slnf` that way reports 381 style errors, all of them in the hand-written **test input** files of the aspect, linker
and template suites and none in any file this branch touches; that is a pre-existing condition of the repository.

`develop/2026.1` is merged in, which brings the change that keys the syntax tree index by `DocumentKey` (#1742). It
conflicted in `Update` and additionally broke the skip trace of `SplitResultsByTree` and the accumulation tests, one of
those only under `ContinuousIntegrationBuild`, where the obsolete `PartialCompilation.SyntaxTrees` becomes an error.

Fixes #1796
Fixes #1799

### DIFF
--- a/Metalama.Framework/docs/design-time-memory.md
+++ b/Metalama.Framework/docs/design-time-memory.md
@@ -86,6 +86,18 @@ When adding a member to a type that is stored in a `SyntaxTreePipelineResult`, t
 convenient to keep" but "does this reach a `Compilation`". Note that a `Microsoft.CodeAnalysis.Diagnostic` does: it
 holds a `Location`, which holds its source tree.
 
+**A type is not a declaration, and `ToDurable()` treats it as one.** `ToDurable()` is backed by a
+`SerializableDeclarationId`, which names a *declaration*, so converting `Base<int>` yields a reference that resolves
+to `Base<T>`: the type arguments are lost, silently, and the result is a usable type rather than an error. Where the
+value is an `IType` rather than a declaration, and above all where it comes from user code and its shape is therefore
+not known, use `DurableRefFactory.FromTypeId( type.GetSerializableTypeId() )` instead.
+`Query.CreateBaseTypeResolver` does, and `RefTests.DurableRefToConstructedGenericTypeLosesTheTypeArguments` records
+the difference.
+
+Neither identifier can denote a type an aspect introduced, because resolving one goes through the symbol table. Where
+such a type can reach the conversion, verify it rather than assume it: converting unconditionally replaces a query
+that works with one that throws `SymbolNotFoundException` when it runs, far from the call site.
+
 ### Understand `ConditionalWeakTable` before relying on it
 
 `WeakCache<TKey, TValue>` wraps a `ConditionalWeakTable`, and the diff subsystem is built on it. The semantics are
@@ -170,6 +182,13 @@ aspect builder, may return the context it holds, rebound through
 `UserCodeExecutionContext.WithCompilationAndDiagnosticAdder`. And a query builder that is handed a declaration converts
 it to a durable reference before capturing it in the adder closure, because the closure lives as long as the query.
 
+Building a context on demand has a consequence that is easy to miss, because it is not about memory. The ordinary
+constructor fills in the target declaration, the meta API and the syntax builder from the context that happens to be
+current, which is right for a context created while its own user code is running and wrong for one created on demand:
+what it would inherit is then whatever the calling thread was doing. `IQuery.ToCollection` is public, so a fabric query
+can be executed from inside an aspect. Use `UserCodeExecutionContext.CreateWithoutInheritance` in that shape;
+`UserCodeExecutionContextTests` holds the distinction.
+
 **The per-file results.** `SyntaxTreePipelineResult` describes itself as compilation-independent and cacheable, and
 the pipeline relies on that: `DesignTimeAspectPipelineResult.Update` carries forward the entry of every file the run
 did not analyse, and `SplitResultsByTree` goes further, discarding a freshly produced item whose file is not dirty so
@@ -216,6 +235,12 @@ Three parts of the harness matter more than the tests themselves.
   field. A suite of liveness tests that all pass is indistinguishable from a suite whose assertions never fire, so
   this positive control is what gives the rest of the suite its value.
 
+**Recording a retention that is known but not yet fixed** is done with `MemoryLeakAssert.RetainedThrough`, which names
+the route as well as asserting that the object is alive. Asserting liveness alone would hold whether the documented
+route retains the object or something else does, and would go on holding after the documented route was removed. Naming
+the route makes the record fail in both of the directions that matter: when the defect is fixed, which is the signal to
+replace the call with `Collected`, and when the object is retained for a different reason than the one written down.
+
 Two practical points when writing such a test.
 
 **Never let a compilation reach a local variable of the test method.** A debug build keeps every local alive until the
@@ -237,8 +262,8 @@ superseded work, the test must cancel too.
 
 Recorded here rather than in a pull request, because this is the document the next person to work on design-time memory
 reads. Strike an entry when it is closed, and add one rather than leaving a repair half-applied. State for each whether
-it was **measured** or **reasoned**, and phrase a reasoned one as a question: an entry that has not been measured is
-about as often wrong as right.
+it was **measured** or **reasoned**, and phrase a reasoned one as a question, so that a reader can tell an observation
+from a hypothesis without re-deriving it.
 
 ### A transitive aspect instance retains one compilation
 
--- a/Metalama.Framework/docs/design-time-memory.md
+++ b/Metalama.Framework/docs/design-time-memory.md
@@ -245,14 +245,25 @@ This is worth re-checking after any change on that surface, because the mistake 
 grep -rn "Task\.WarnIfLongAsync" --include=*.cs . | grep -v WithCancellation
 ```
 
-### `WithCancellation` leaves a small object behind per cancelled wait
+### `WithCancellation` does not leave anything behind: measured, contrary to a claim made here earlier
 
-The fix above releases the awaiting frame, which is what matters, because that frame is what holds the Roslyn objects.
-It does not detach the `Task.WhenAny` continuation from the source that is never completed, so a cancelled wait leaves
-a promise of about a hundred bytes attached to it. That is unbounded over a session in which no client ever connects,
-though it holds nothing bound to a compilation. The shape that avoids it is a collection of waiters an entry can be
-removed from, as in the `DesignTimeAspectPipelineFactory` fix for
-[#1793](https://github.com/metalama/Metalama/issues/1793). This was reasoned about rather than measured.
+An earlier revision of this section stated that a cancelled `WithCancellation` leaves a `Task.WhenAny` continuation on
+the task it waited for, and that this accumulates on a task that never completes. **That is false, and it was asserted
+from reading the code rather than from measuring it.**
+
+`Task.WhenAny` removes its continuation from the task that did not win. Measured on a task that is never completed, the
+number of continuations left after 10, 100 and 1000 cancelled waits is the same small constant, for both
+implementations in use: `TaskExtensions.WithCancellation` of this codebase, and
+`Microsoft.VisualStudio.Threading.ThreadingTools.WithCancellation`, which the two waits in
+`Metalama.Framework.DesignTime.Rpc` resolve to because that project does not reference the engine.
+
+`WithCancellationMemoryLeakTests` holds the property, since it belongs to the runtime and to a third-party library
+rather than to this codebase, and nothing else here would notice its loss. It counts continuations by reflection over
+a private field of `Task`, which is the only way to observe something that no public API exposes and that no weak
+reference can track.
+
+The lesson is the one this section exists to enforce: an entry here has to say whether it was measured or reasoned
+about, and a reasoned one is a question rather than a finding.
 
 ### `TransitiveAspectInstance` retains the compilation it was produced in
 
--- a/Metalama.Framework/docs/design-time-memory.md
+++ b/Metalama.Framework/docs/design-time-memory.md
@@ -229,24 +229,17 @@ What follows is deliberately not fixed, or not fixed everywhere. It is recorded 
 request description, because the next person to work on design-time memory will read this document and not that.
 Please keep the list honest: strike an entry when it is closed, and add one rather than leaving a repair half-applied.
 
-### Two remaining uncancellable waits on a task completion source
+### Uncancellable waits on a task completion source
 
-`RpcService.WaitUntilInitializedAsync` was fixed for [#1799](https://github.com/metalama/Metalama/issues/1799), but the
-same pattern survives in two places:
+Closed. All three call sites now compose the token with `WithCancellation` before `WarnIfLongAsync`, which is the
+defect: that method uses the token for its own delay and returns the original task untouched when warning logging is
+disabled, so passing it there alone leaves the wait uncancellable by either path. `RpcServiceProviderServerEndpoint`
+also completes `_hubRegistrationTask` when the service hub endpoint cannot be obtained, because registration is
+attempted once and nothing else would ever complete it. The awaiters of `ClientEndpoint` are deliberately left waiting
+in that situation, because a later connection attempt can legitimately complete them; making the wait cancellable is
+the whole of what was needed there.
 
-- `Metalama.Framework.DesignTime/VisualStudio/ServiceProvider/RpcServiceProviderServerEndpoint.cs:133`, on
-  `_hubRegistrationTask`, which `OnServerPipeCreatedAsync` leaves uncompleted when the Visual Studio extension is
-  absent.
-- `Metalama.Framework.DesignTime.Rpc/ClientEndpoint.cs:140`, on an entry of `_clientAwaiters`, which is not completed
-  when a connection attempt fails: the catch path calls `RpcClient.SetFailure`, which completes the client's own
-  source and not the awaiter.
-
-Both pass the cancellation token to `WarnIfLongAsync` alone, which is the defect: that method uses the token for its
-own delay and returns the original task untouched when warning logging is disabled. The fix is the one already applied
-to `RpcService`, namely to compose with `WithCancellation` first. Tracked by
-[#1800](https://github.com/metalama/Metalama/issues/1800).
-
-To find any recurrence:
+This is worth re-checking after any change on that surface, because the mistake is easy to repeat and invisible:
 
 ```bash
 grep -rn "Task\.WarnIfLongAsync" --include=*.cs . | grep -v WithCancellation
--- a/Metalama.Framework/docs/design-time-memory.md
+++ b/Metalama.Framework/docs/design-time-memory.md
@@ -41,6 +41,9 @@ It is useful to classify every field by the lifetime of the object that declares
 | A single request (locals, parameters, a `PartialCompilation` being processed) | Yes, freely. |
 | The current version of a project (`PipelineState.ProjectVersion`) | Yes, exactly one version. |
 | A project (`DesignTimeAspectPipeline`, a `ProjectSourceGenerator`) | No. |
+| The pipeline configuration (`AspectPipelineConfiguration` and everything it reaches) | No: it is long-lived by design, reused across keystrokes. |
+| An ambient `UserCodeExecutionContext` | Yes: it is scoped to one execution of user code. |
+| A per-file result (`SyntaxTreePipelineResult` and everything it holds) | No: it survives every run in which its file is not dirty. |
 | The process (a static field, an `IGlobalService`, an analyzer or generator instance) | No. |
 
 The last row deserves emphasis: Roslyn keeps **one instance** of a `DiagnosticAnalyzer` or an `IIncrementalGenerator`
@@ -130,6 +133,57 @@ An event handler holds its target. `AnalysisProcessEventHub` exposes its high-tr
 `WeakEvent<T>` and `AsyncWeakEvent<T>`, which hold subscribers weakly and compact their list. Where a plain event is
 used instead, the subscriber must unsubscribe in `Dispose`, and the event must carry only values, such as a
 `ProjectKey`, rather than compilation-bound objects.
+
+## What the pipeline stores for longer than one request
+
+Three collections outlive a request without being obviously long-lived, because they belong to objects a reader
+naturally takes to be per-run. Each of them is a place where the rule has been broken.
+
+**The pipeline configuration.** `AspectPipelineConfiguration` is built once, from whichever version of the project was
+current at the time, and `PipelineState` then reuses it for every subsequent version, discarding it only when the
+compile-time code changes. **That long life is deliberate and is not the defect.** Rebuilding the configuration on
+every keystroke would mean recompiling the compile-time assemblies each time, which is far too slow to be done between
+one keystroke and the next; the reuse is what makes the design-time pipeline viable and it must stay.
+
+What follows from it is a constraint, not a licence: because the configuration is long-lived, **it must not pin a
+compilation**. For a session in which the user edits run-time code alone, anything it holds that is bound to a
+compilation retains the *first* version of the session, the oldest rather than the most recent, and holds it in
+addition to the current one. The exception in the rule above does not cover that. Everything reachable from the
+configuration is therefore subject to the rule, including `FabricsContributors`, which reaches the amender of every
+static fabric and, through the amender, the queries it owns. `FabricMemoryLeakTests` guards this.
+
+The corollary is worth stating separately, because it is where the rule is easiest to break by accident.
+`UserCodeExecutionContext` is ambient and scoped to one execution of user code, so it is entitled to hold a
+compilation, a declaration and a syntax builder: that is its purpose. **The objects created from a context and outliving
+it are the ones that must not.** An amender stored a whole context, and a query built by `SelectTypesDerivedFrom`
+captured the `INamedType` it was given. Both are durable, both are reachable from the configuration, and both were
+fixed by keeping a durable reference and resolving it against the compilation of each run, which
+`UserCodeExecutionContext.WithCompilationAndDiagnosticAdder` was already doing for everything else.
+
+**The per-file results.** `SyntaxTreePipelineResult` describes itself as compilation-independent and cacheable, and
+the pipeline relies on that: `DesignTimeAspectPipelineResult.Update` carries forward the entry of every file the run
+did not analyse, and `SplitResultsByTree` goes further, discarding a freshly produced item whose file is not dirty so
+that the item produced earlier survives. A file the user never edits therefore keeps, for the whole session, whatever
+its first analysis produced. Two kinds of member of that result reach the code model and must be checked:
+
+- the extensions, which are opaque to the engine: an extension chooses what its contributor carries, and nothing in
+  `IDesignTimePipelineResultExtension` or `ITransitivePipelineContributor.ToDesignTime` says that it must be durable.
+  It must. `ExtensionContributorMemoryLeakTests` states this as a matched pair, one contributor holding a durable
+  reference and one holding the reference the code model returns.
+- the diagnostics. A `DiagnosticDefinition` formats lazily, so the arguments passed to `WithArguments` are stored in
+  the descriptor of the diagnostic and are held for as long as the diagnostic is. An argument that is an
+  `IDeclaration` reaches its `CompilationModel`. Passing the declaration is the natural way to write the message, and
+  it is what a validator does on every reference it rejects; passing `declaration.Name`, or another value, is what
+  keeps the diagnostic durable. `DiagnosticArgumentMemoryLeakTests` states this, also as a matched pair.
+
+**A `TaskCompletionSource` that no path completes.** Awaiting one retains the awaiting frame, and with it every
+argument and local the frame captured. On the design-time RPC surface those captures are Roslyn objects: a service is
+initialized only when a client attaches to its endpoint, and every method of every server-side service awaits that
+initialization before doing anything, including before checking whether there is any client to serve. When no client
+ever attaches, which is the ordinary state of the analysis process when the Visual Studio extension is absent, each
+such call parks forever. Passing a cancellation token is not by itself enough: it has to be *observed*, which means
+composing it with `WithCancellation` rather than handing it to a helper that only uses it for a timeout.
+`RpcServiceCancellationTests` states the property for the wait itself, and separately for what the parked frame holds.
 
 ## Testing it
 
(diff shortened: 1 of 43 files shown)


gfraiteur:
Converting to draft. **Do not merge**: the fix in this branch is wrong, and I found it by checking `Metalama.Premium`, as asked.

## What is wrong

`SplitResultsByTree` decides where a contributor goes from `extension.SyntaxTree`. For a reference validator that is
`ReferenceValidatorInstance.SyntaxTree => this.ValidatedDeclaration.GetPrimarySyntaxTree()`, that is, the tree of the
**validated declaration**, not of the code that declared the validator.

So there are two distinct situations behind the single "path is not a key of the result builders" condition:

1. **The tree belongs to this project but is not dirty.** The tree keeps the result of the run that did include it and
   that result already carries the contributor. Skipping is correct, and re-adding is what accumulates. This is the
   case I reasoned about, by analogy with the inheritable-aspect fix of #1768.
2. **The tree belongs to another project.** A fabric in this project can validate references to declarations in a
   referenced project, which is a core `Metalama.Extensions.Architecture` scenario. At design time a project reference
   inside the solution is a `CompilationReference`, so the symbol does have a syntax tree, and that tree belongs to the
   other project`s compilation. This project has no `SyntaxTreePipelineResult` keyed by that path and never will.
   **Skipping drops the validator.**

The comment I deleted, "This happens with cross-project validators", was naming case 2. I dismissed it because I could
not find a producer in the open-source repository. The producer is in Metalama.Premium.

## What the correct fix looks like

The two cases have to be told apart, which `compilation.Compilation.ContainsSyntaxTree( syntaxTree )` does:

- own tree, not dirty: skip, as this branch does;
- foreign tree or no tree: keep, but in a bucket that is **replaced** on every run rather than appended.

Replacing is sound for case 2, and this is the part Premium settles:
`ValidationPipelineExtension.ExecuteDesignTimePipelineContributorsAsync` calls
`GetReferenceValidatorsAsync( initialCompilation, ... )`, which runs every validator source over the whole
compilation. The complete set of reference validators is therefore re-produced on every design-time run, so replacing
the previous set loses nothing, while appending is exactly what grows.

A blanket "replace everything external" would not do either: it would duplicate the own-tree contributors that their
own tree`s result already carries. The distinction is load-bearing.

## Next

I am building this branch so that Premium can consume it through a local dependency and its validation tests can
confirm both halves before I push a corrected fix. The tests already on this branch stay valid: they cover case 1, and
a case 2 test needs to be added.

— Claude for @gfraiteur

gfraiteur:
Confirmed empirically against `Metalama.Premium`. The fix in this branch **does** break cross-project validation.

## The experiment

`Metalama.Premium` was pointed at this worktree with
`Build.ps1 dependencies set local Metalama --path <worktree>`, so that its tests compile against this branch.

`Metalama.Extensions.Validation.UnitTests.SideBySideVersionTests.TransitiveValidator` is exactly the case 2 scenario:
a project fabric in the master project calls `ValidateInboundReferences` on `p.Types`, and a dependent project
references one of those types. The test asserts that the analyzer of the dependent project reports one diagnostic.

| Metalama build | Result |
|---|---|
| this branch | **Failed** - `Assert.Single() Failure: The collection was empty` |
| same branch with only `DesignTimeAspectPipelineResult.cs` reverted | **Passed** |

Nothing else differed between the two runs, so the attribution is unambiguous: skipping the contributor drops the
validator and the diagnostic disappears.

That is the failure mode a user would see as "my architecture rules silently stopped being enforced in the IDE", which
is worse than the memory growth this change was meant to fix.

## What this settles for the corrected fix

The two cases must be told apart by whether the tree belongs to this compilation, as set out in the previous comment.
Premium also settles the question I could not answer from the framework alone:
`ValidationPipelineExtension.ExecuteDesignTimePipelineContributorsAsync` calls
`GetReferenceValidatorsAsync( initialCompilation, ... )`, which runs every validator source over the whole
compilation, so the complete set of reference validators is re-produced on every design-time run. Replacing the
foreign-tree set each run is therefore lossless, whereas appending it is what grows.

`TransitiveValidator` becomes the regression test for the foreign-tree half, and the tests already on this branch
cover the own-tree half.

This branch stays a draft until the corrected fix passes both.

— Claude for @gfraiteur

gfraiteur:
Addressed, with a test per remark. Two of the findings were defects I had introduced, and both are now
demonstrated by a test that fails on the previous code.

## 1. `_foreignExtensions` replaced unconditionally — fixed, and worse than reported

`0614deb27f`. The set is also empty when the **pipeline fails**, not only when the extension gate does not fire:
`PipelineState` reads `pipelineResultValue?.TransitiveContributors ?? Empty`.
`RunWithoutContributors_KeepsTheContributorsOfAnotherCompilation` reproduces it and failed before the fix.

The set is now replaced only when the run produced some. The residual is documented in the code: a rule genuinely
removed survives until the next run producing any foreign contributor, and removing a rule is a compile-time edit,
which discards the configuration and this result with it.

## 3. `CreateBaseTypeResolver` — you were right, twice

`ba5ec81093`. Taken first because it was the sharpest, and it turned out to hold two separate defects.

**A constructed generic base type was silently widened.** `ToDurable()` is backed by a `SerializableDeclarationId`,
which names a *declaration*, so `Base<int>` resolved back to `Base<T>`: not an error, a usable type, and the query
matched the types derived from every construction. `Tests/Fabrics/SelectTypesDerivedFromNamedType.cs` is the aspect
test; on the previous conversion it overrides two methods it must not. The durable form is now a
`SerializableTypeId`, and `RefTests.DurableRefToConstructedGenericTypeLosesTheTypeArguments` records the difference
between the two identifiers directly.

**An introduced base type made the query throw.** A type identifier resolves through the symbol table, so it cannot
denote a type an aspect introduced. Converting unconditionally replaced a query that works with one that throws
`SymbolNotFoundException` while it runs, exactly as you predicted, far from the call site and inside user code
execution. `Tests/Fabrics/SelectTypesDerivedFromIntroducedType.cs` shows it end to end. The conversion is now
verified rather than assumed, and the type is kept as it is when no identifier denotes it. Keeping it is safe here,
and the reasoning is worth stating because it is what bounds the retention: only an aspect can hold an introduced
type, the queries of an aspect do not outlive the run, and a fabric, whose queries do, runs before any aspect and
never sees one.

Worth recording: this overload had **no test at all** before this, here or in Metalama.Premium, where both fabric
tests use the `System.Type` overload.

One thing the tests document rather than fix: `SelectTypesDerivedFrom( Base<int> )` matches nothing, because the
derived-type index is keyed on the definition. That is the behaviour before this branch as well, so it is not a
regression, but it may be a gap worth an issue.

## 2. Ambient inheritance in `GetUserCodeExecutionContext` — fixed

`f72ef24c78`. Correct, and I had not remarked on it. The inheritance used to be sampled once, while the fabric was
running; it is now sampled on demand. `ToCollection` is public, so a fabric query can be executed from inside an
aspect and would acquire that aspect's target declaration and meta API. The meta API is the harmful one, since
`WithCompilationAndDiagnosticAdder` asserts against carrying one across compilations.

`UserCodeExecutionContext.CreateWithoutInheritance` is the shape for an owner that is not itself running user code.
`UserCodeExecutionContextTests` holds the distinction; its first test is the control establishing that the ordinary
factory does inherit.

## 4. Eager materialization on the compile-time path — measured, not gated

`b038ad58ce`. Measured rather than argued: creating a diagnostic whose argument is a declaration takes about 3.2 µs,
of which about 2.3 are the display string; the same diagnostic with a string argument takes about 0.25. A build
creating ten thousand such diagnostics and displaying none would spend about 20 ms more. That does not justify making
the behaviour conditional on the execution scenario, which this static extension has no way of reading. The figures
are in the `<remarks>`.

`bool` and `char` are now excluded, and the array case no longer allocates a delegate and an iterator. In the
interest of not overclaiming: no test can distinguish the `bool`/`char` decision, because materializing them yields
the same message; that is stated where the decision is made.

`DiagnosticArgumentMaterializationTests` covers what the mechanism actually risks, per category of argument: that
early formatting differs from late formatting. The two discriminating cases are `{0:x}`, which would throw, and
`{0:N2}`, which would silently lose its grouping.

## 5. `syntaxTree!` — fixed

`0614deb27f`. Explicit `syntaxTree != null &&`, and the comment below no longer reads as an exhaustive claim.

## 6. `IQueryOwner` binary break

Not pursued, on the author's instruction: Metalama.Premium always uses the version of Metalama it was compiled
against.

## Style and test-coverage remarks

- The `<remarks>` on `PullConstructorParameterTransitiveAspect._parameter`, and the one in its serializer, were worse
  than misplaced: both described a durable field and a `ToDurable` call on the deserialization boundary, neither of
  which exists. They were left behind by the reverted experiment. Rewritten to say what the code does.
- `WithCancellationMemoryLeakTests` now says why it deliberately does not derive from `UnitTestClass`.
- **The two inverted controls.** Your diagnosis of the mechanism was not quite right, `GarbageCollectionHelper.Collect`
  already performs five blocking compacting rounds with finalizers, but the conclusion was: asserting liveness alone
  holds whether the documented route retains the object or something else does, and goes on holding after that route
  is removed. `MemoryLeakAssert.RetainedThrough` now requires the retention chain to pass through a named type. Both
  controls pass, so the routes are the ones recorded.
- The doc register: the passages you quoted are gone, and one more of the same kind with them.

## Verification

Unit tests 2303 on `net8.0` and 2247 on `net48`, aspect tests 2273, all with 0 failures.
`PullParameterTests.CrossProjectIntegration`, which runs on `net48` only, is included in that net48 figure.

On the CI-mode build I need to correct something I said earlier in this pull request. My previous "0 errors" was for
the production projects, which is what I had built; building the whole `.slnf` with
`-p:ContinuousIntegrationBuild=true` reports 381 style errors, all of them in the hand-written **test input** files of
the aspect, linker and template suites, none in any file this branch touches. That is a pre-existing condition of the
repository, not a result of this branch, but I should not have described the build as clean without saying which
projects I had built. Each project this branch modifies builds with 0 errors in CI mode.

— Claude for @gfraiteur

Issue #1838, opened 2026-08-08
The same type gets two different serializable type ids


gfraiteur:
## Summary

`SerializableTypeIdGenerator` has two overloads that are required to produce the same string for the same type, and they do not agree for any value type, nor for an annotated reference type.

The requirement is stated by the generator itself, in the comment on the reflection overload:

> The id must be byte-identical to the one the `ITypeSymbol`/`IType` overloads produce (via the syntax generator), because `CompileTimeType` equality and the `CompileTimeTypeFactory` cache key on this string.

## Measured

Fields of the given types, in a compilation with `struct S { }`:

| Type | `IType` overload | `ITypeSymbol` overload | Same |
|---|---|---|---|
| `object` | `Y:global::System.Object!` | `Y:global::System.Object!` | yes |
| `object?` | `Y:global::System.Object?` | `Y:global::System.Object?!` | **no** |
| `int` | `Y:global::System.Int32` | `Y:global::System.Int32!` | **no** |
| `int?` | `Y:global::System.Int32?` | `Y:global::System.Int32?!` | **no** |
| `S` | `Y:global::S` | `Y:global::S!` | **no** |
| `S?` | `Y:global::S?` | `Y:global::S?!` | **no** |
| `List<object?>` | `Y:…List<global::System.Object?>!` | same | yes |

## Cause

The two overloads decide the trailing marker by different tests.

```csharp
// ITypeSymbol overload
if ( symbol.NullableAnnotation != NullableAnnotation.None ) { id += '!'; }

// IType overload
if ( type.IsNullable == false && type.IsReferenceType != false ) { id += '!'; }
```

`NullableAnnotation != None` is true for every value type, whose annotation is `NotAnnotated`, and it is also true for an annotated reference type, which is why `object?` acquires a marker that contradicts its own `?`. The `IType` overload tests what the marker is meant to express, which is that the type is a non-nullable reference type.

## Impact

Not demonstrated, and I did not attempt to construct a failing scenario. Resolution tolerates the stray marker, because applying the non-nullable annotation to a value type is a no-operation, so both identifiers resolve to the same type. What the divergence can affect is anything that treats the identifier as a key rather than resolving it:

- `CompileTimeTypeFactory` keys its cache on the identifier and is fed by the `ITypeSymbol` overload, whereas `DurableRefFactory` and `TemplateCompilerRewriter` use the `IType` overload. Two entries can exist for one type.
[...]

PR #1852, opened 2026-08-09
Derive the nullability marker of a SerializableTypeId from the whole type (#1837)

### LINKED ISSUE #1837: The nullability marker of a SerializableTypeId must state the nullable context of the whole reference, not the annotation of its outermost position
## The intended design

A type reference belongs to a single nullable context, that of the place it was written, so one bit describes every position of it. The identifier records nullability with two markers that are not alternatives:

- `?` belongs to one position and means that this position is nullable, exactly as in C#.
- `!` belongs to the whole identifier and means that the reference was written in an annotated context, so that every position without a `?` is non-nullable rather than oblivious.

The absence of `!` means an unannotated context, in which every position is oblivious. `Y:…List<…String>?!` is therefore well formed and necessary: the list is nullable, the context is annotated, so the `string` is non-nullable.

## What was wrong when writing the identifier

The marker was derived from the outermost position alone, which does not determine the context:

- an outermost value type is `NotAnnotated` in either context, because it can never be oblivious, so `KeyValuePair<string, string>` and `(string A, string B)` carried no marker and their `string` positions came back oblivious;
- an outermost annotated reference type says that the position is nullable, not that the context was unannotated, so `List<string>?` carried no marker and its `string` came back oblivious.

The context has to be recovered from the whole tree of the type instead. Only a reference type and a type parameter prove anything, and only when they are `NotAnnotated`: `None` is what an unannotated context produces, and `Annotated` proves nothing either, because `string?` in an unannotated context is a warning rather than an error and Roslyn annotates it all the same.

## What was wrong when reading it back

The resolver could produce only two of the three states. It had an operation that adds the non-nullable annotation, applied when the identifier carries the marker, and none that makes a type oblivious. A type looked up in the compilation is already not annotated, so an identifier written in an unannotated context, which correctly carries no marker, resolved to a non-nullable type. Where a type was built rather than looked up, the result depended on what the Roslyn API used happened to default to, so `List<string>` came back oblivious at its outer position and not annotated at its argument.

## Type parameters

The nullability of a type parameter has to be read from its symbol and not from `IType.IsNullable`, which answers whether the constraint allows null rather than what the annotation of the reference is:

| | symbol annotation | `IType.IsNullable` |
|---|---|---|
| unconstrained parameter, used in an annotated context | `NotAnnotated` | `null` |
| `class` or `notnull` constrained parameter, its declaration | `None` | `False` |

Either disagreement made the two overloads of `GetSerializableTypeId` produce different identifiers for the same type. The generic context appended after a `|` resolves the parameter as its declaration declares it, and the marker records the context of the reference, which the declaration cannot supply: without it, a use of `T` in an annotated context resolved back to the oblivious declaration.

## Resolution

Fixed by #1852, which derives the marker from the whole type, adds the mirror of the non-nullable operation so that an unannotated identifier resolves to an oblivious type, reads the nullability of a type parameter from its symbol, and passes the annotation of the element of an array to `CreateArrayTypeSymbol`, which defaults it to `None` and had been discarding it.

`ATypeResolvesFromItsIdentifierToItself` asserts the round trip in both nullable contexts over a corpus chosen for the cases above. The format is documented on `SerializableTypeId`.

This supersedes the rule established by #1838, under which the marker described the outermost type. It now describes the context, so a value type and an annotated reference type can carry it and a type parameter is no longer excluded.

## Related

- #1797, which put this format on the durable reference of every named type.
- #1835, the symptom that led here, and #1836.
- #1838 and #1839, which fixed the parts of this that were reachable without changing what the marker means.

— Claude for @gfraiteur

gfraiteur:
Follows #1836, which left this open.

Fixes #1837

## The design

The nullability markers of a `SerializableTypeId` describe different things and are not alternatives:

- `?` belongs to one position and means that this position is nullable, exactly as in C#.
- `!` belongs to the whole identifier and means that the type was written in an annotated nullable context, so that every position without a `?` is non-nullable rather than oblivious.

A reference belongs to a single nullable context, that of the place it was written, so one bit describes all of its positions. The absence of `!` means an unannotated context, in which every position is oblivious.

## Writing the marker

It was derived from the outermost position alone, which does not determine the context:

- an outermost value type is `NotAnnotated` in either context, because it can never be oblivious, so `KeyValuePair<string, string>` and `(string A, string B)` carried no marker and their `string` positions came back oblivious;
- an outermost annotated reference type says that the position is nullable, not that the context was unannotated, so `List<string>?` lost its argument the same way.

The whole tree of the type is searched instead. Only a reference type and a type parameter prove anything, and only when they are `NotAnnotated`: a value type proves nothing, and an annotated position proves nothing either, `string?` in an unannotated context being a warning rather than an error and annotated all the same.

## Reading it back

The resolver could produce only two of the three states. It could set the nullable annotation, from `?`, and the non-nullable one, from `!`, and had no operation that made a type oblivious. A type looked up in the compilation is already not annotated, so an identifier written in an unannotated context resolved to a non-nullable type. Where a type was built rather than looked up, the result depended on what the Roslyn API used happened to default to, so `List<string>` came back oblivious at its outer position and not annotated at its argument.

`AddObliviousAnnotation` is the mirror of `AddNonNullableAnnotation` and is applied wherever that one is not. It applies to a reference type and to a type parameter alone, a value type being not annotated in either context.

## Type parameters

The nullability of a type parameter is read from its symbol rather than from `IType.IsNullable`, which answers whether the constraint of the parameter allows null and not what the annotation of the reference is. The two disagree in both directions, and either disagreement made the two overloads of `GetSerializableTypeId` produce different identifiers for the same type:

| | symbol annotation | `IType.IsNullable` |
|---|---|---|
| unconstrained parameter, used in an annotated context | `NotAnnotated` | `null` |
| `class` or `notnull` constrained parameter, its declaration | `None` | `False` |

The generic context appended after a `|` resolves the parameter as its declaration declares it. The marker records the context of the reference, which is a different thing and which the declaration cannot supply: without it, a use of `T` in an annotated context resolved back to the oblivious declaration.

## Also

The element of an array lost its annotation on resolution. `CreateArrayTypeSymbol` takes it as a parameter that defaults to `NullableAnnotation.None`, and the default discarded what the resolved element carried, so `string[]` of an annotated context came back with an oblivious element.

The format is documented on `SerializableTypeId`, which is updated to the semantics above.

## Tests

`ATypeResolvesFromItsIdentifierToItself` asserts the round trip in both nullable contexts, over a corpus chosen for the cases above: a value type over reference types, a tuple, a nullable generic type, a nullable argument, and arrays of one and two ranks.

`TestTypeParameter` no longer asserts that the identifier of a type parameter carries no marker. That was the rule of #1838, which this supersedes: whether the marker appears depends on the annotation of the reference and not on the constraint, so the declaration of a parameter constrained to a value type carries it and the declaration of one constrained to `class` does not. The round trip is what is asserted.

## Verification

Unit tests 2550 passed, aspect tests 2278 passed, template tests 317 passed, no failures. `Metalama.Framework.Engine` builds with `ContinuousIntegrationBuild=True` and no warning. No expected output changed anywhere, the identifiers that appear in the template baselines being unaffected: an annotated outer position proves nothing and a value type has no informative position.

### DIFF
--- a/Metalama.Framework/src/Metalama.Framework.Engine/SerializableIds/SerializableTypeIdGenerator.cs
+++ b/Metalama.Framework/src/Metalama.Framework.Engine/SerializableIds/SerializableTypeIdGenerator.cs
@@ -8,10 +8,13 @@ using Metalama.Framework.Engine.SyntaxGeneration;
 using Metalama.Framework.Engine.Utilities;
 using Metalama.Framework.Engine.Utilities.Caching;
 using Metalama.Framework.Engine.Utilities.Roslyn;
+using Metalama.Framework.Code.Types;
+using Metalama.Framework.Engine.CodeModel.Abstractions;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using System;
 using System.Text;
+using CodeTypeKind = Metalama.Framework.Code.TypeKind;
 
 namespace Metalama.Framework.Engine.SerializableIds;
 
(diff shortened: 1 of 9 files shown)


Metalama.Framework/src/Metalama.Framework.Engine/SerializableIds/SerializableTypeIdGenerator.cs
```diff
@@ -24,47 +27,98 @@ public static class SerializableTypeIdGenerator
     private static readonly WeakCache<Type, SerializableTypeId> _reflectionTypeIdCache = new( isStaticCache: true );
 
     /// <summary>
-    /// Determines whether the nullability marker is appended to the identifier of a type, which happens for a
-    /// reference type that is not annotated, and for no other type.
+    /// Determines whether a position of a type carries information about the nullable context it was written in, which
```
gfraiteur: "a position" of a type is not precise language; I don't understand what it means. Use official C# or Roslyn terminology.

Metalama.Framework/src/Metalama.Framework.Engine/SerializableIds/SerializableTypeIdGenerator.cs
```diff
@@ -91,14 +145,97 @@ public static SerializableTypeId GetSerializableTypeId( this TypeSyntax typeSynt
...
+    /// <summary>
+    /// Answers what <see cref="IsWrittenInAnnotatedContext(ITypeSymbol)"/> answers, for a reflection type, which
+    /// carries no nullable annotation at all.
+    /// </summary>
+    /// <remarks>
+    /// A reflection type cannot say which context it was written in, so an annotated one is assumed whenever the type
+    /// has a position that could be oblivious. This is the assumption the identifier of a reflection type has always
+    /// made, and it is what keeps the three overloads producing the same string for a type of an annotated context.
+    /// </remarks>
+    private static bool IsWrittenInAnnotatedContext( Type type )
```
gfraiteur: Most likely a conceptual error. Reflection types are always null-oblivious.

Metalama.Framework/src/Metalama.Framework.Engine/SerializableIds/SerializableTypeIdGenerator.cs
gfraiteur: The term is **type reference**, and it comes from Roslyn rather than from the C# specification.

The nullable metadata specification of Roslyn describes `NullableAttribute` as encoding one byte for each type reference of a type, walked in order, which is exactly the thing the marker has to describe: `List<string>` has two of them, `string[]` has two, `KeyValuePair<int, int>` has three of which none is informative. The public API names them the same way: the annotation of a type reference is `ITypeSymbol.NullableAnnotation`, and the nested ones are reachable through `INamedTypeSymbol.TypeArgumentNullableAnnotations` and `IArrayTypeSymbol.ElementNullableAnnotation`.

The C# specification itself has no single word for it. It speaks of the nullable annotation of a type, and where it needs to distinguish the outer type from the ones inside it, it says so in words. So "type reference" is the closest thing to an official name, and it is the one the implementation of this feature uses.

Every use of "position" in the documentation of this change is replaced by it in ffb071c0be. The word survives elsewhere in these files with its ordinary meaning, the position of an element of a tuple, which is unrelated.

`AddObliviousAnnotation` is renamed `AddNullObliviousAnnotation` in the same commit.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/SerializableIds/SerializableTypeIdGenerator.cs
gfraiteur: You are right about the fact, and I would rather you decide the consequence, because the two possible answers conflict and I do not know which constraint wins.

A reflection type carries no nullable annotation, so it is null-oblivious. On that reading this method should return `false` always, and the identifier of a reflection type should never carry the marker.

What stops me from making that change is the invariant that the three overloads of `GetSerializableTypeId` have to produce the same string for the same type, because `CompileTimeType` equality and the cache of `CompileTimeTypeFactory` key on that string. `typeof(string)` reached through reflection would then be `Y:global::System.String`, while the same type reached through a symbol in an annotated context is `Y:global::System.String!`, and the two would no longer meet in the cache.

The code before this pull request resolved the conflict in favour of the invariant, and said so: it appended the marker for any non-nullable reference type and its comment read "a reference type is non-nullable oblivious here (there is no annotation on a reflection Type), which the symbol side renders as '!'". I kept that decision and extended it to the nested type references, so that a reflection type and a symbol of an annotated context still agree. That is why `typeof(KeyValuePair<string, string>)` now carries the marker where it did not before.

So the question is which of the two is the error:

1. treating a reflection type as annotated, which is what both the old code and this change do, and which keeps the three overloads in agreement while describing the type wrongly;
2. treating it as oblivious, which describes it correctly and splits the identifier of a type according to how it was reached.

If you want the second, the invariant has to be given up or satisfied some other way, and I would want to look at how `CompileTimeTypeFactory` uses the identifier before changing it. Tell me which you prefer and I will do it.

— Claude for @gfraiteur

gfraiteur: You were right and my objection did not survive checking. Fixed in b3f30e39bc.

A reflection type is null-oblivious, `typeof` not being able to name a nullable reference type at all, so the unmarked identifier is what describes it. The marker is no longer appended by that overload, and the helper that computed it is gone.

The invariant I cited turned out not to be the obstacle. `ReflectionTypeId_EqualsCodeModelTypeId` compares `typeof(T)` against `Factory.GetTypeByReflectionType(typeof(T))`, so both sides start from the same reflection type: it never asserted that a reflection type and a symbol read from source agree. What it asserted, once nullability is set aside, is that the two agree on the qualification of names and on the rendering of a generic type, of an array and of `Nullable<T>`. It now removes the marker before comparing and says so, since the two sides answer nullability differently by design: the reflection type is oblivious and `GetTypeByReflectionType` returns a non-nullable type, the annotated context having been the norm since before Metalama existed.

The two identifiers asserted by `CompileTimeTypeFactoryTests` lose the marker for the same reason, both being built from a reflection type.

Nothing else moved: unit 2550, aspect 2278, template 317, no failures, and `Metalama.Framework.Engine` builds with `ContinuousIntegrationBuild=True` and no warning.

On your other suggestion, that `StripNullabilityAnnotation` take a `deep` parameter implemented with a type rewriter: I agree it is the right shape, and it would let this test say `StripNullabilityAnnotation( deep: true )` instead of trimming the marker off the string. I have not done it here. It adds a parameter to a public member of `IType` and needs a rewriter of its own with tests, which is a wider change than what this pull request is about, and the test reads honestly as it stands because what it compares is precisely two identifiers that differ by the marker alone. I would rather do it as its own change. Say the word and I will open an issue for it.

— Claude for @gfraiteur

PR #1855, opened 2026-08-11
Make durable references scope-dependent (#1811)

### LINKED ISSUE #1811: Performance: optimize durable references in compile-time builds
A durable reference exists for one reason: an object that outlives a single design-time request must not hold a
`Compilation`, because the analysis process is long-lived and Roslyn produces a new `Compilation` per keystroke. The
implementation follows from that: `FullRef.ToDurable` and `DurableRefFactory.FromDeclarationOrType` collapse a
reference to a `SerializableDeclarationId` or a `SerializableTypeId`, and resolving it again walks the symbol table.

**That motivation does not exist at compile time.** A batch compilation has one compilation, which lives for the whole
build, and nothing outlives it. Converting a reference to an identifier and resolving it again buys nothing there, and
the two halves of the round trip are not free.

## What the round trip costs

Both halves run on the compile-time path today, because the call sites are shared:

- **Writing.** `FullRef.ToSerializableId` resolves the symbol; `IntroducedRef.ToSerializableId` goes through
  `ConstructedDeclaration`, which is `GetTarget( RefFactory.CanonicalCompilation )`, so the declaration is resolved
  merely in order to name it.
- **Reading.** `SerializableDeclarationId.ResolveToDeclaration` goes to `DocumentationIdHelper`, which enumerates the
  merged global namespace, collects the candidates of an identifier and orders them. For a constructor the parser then
  compares parameter lists to choose between the source constructor and the one an aspect extended.

Against holding a symbol, which is what the live reference does, this is a large constant on an operation that is not
rare.

## Where it is paid

The references made durable for [#1797](https://github.com/metalama/Metalama/issues/1797) and
[#1799](https://github.com/metalama/Metalama/issues/1799) are all on code that runs at compile time as well as at
design time:

- `PullConstructorParameterTransitiveAspect._parameter` and the `IType` references of the pull strategies
  (`IntroduceParameterPullStrategy`, `DefaultParameterPullStrategy`, `OnConstructedEpilogueEmitter`), resolved on every
  constructor the recursive pull visits;
- `TransitiveAspectInstance.TargetDeclaration`;
- `Query.CreateBaseTypeResolver`, which resolves the base type once per query execution;
- the fabric amender and the validator surfaces of #1799.

None of these needs an identifier during a batch build.

## Proposal

Introduce an `IDurableRefFactory` service, resolved from the service provider like the rest of the engine's services,
with one implementation per scope:

- **design time**: the current identifier-based implementation;
- **compile time**: an implementation that returns the live reference, or a thin wrapper over it, since the compilation
  outlives every object that could hold it.

The call sites keep asking for a durable reference and stop deciding what that means, which is the property worth
having: the discipline documented in `Metalama.Framework/docs/design-time-memory.md` stays expressible as
"make it durable", and its cost stops being charged to the compiler.

## One thing the design has to keep

**Serialization needs a real identifier whatever the scope.** A `TransitiveAspectInstance` is written to the transitive
manifest during a batch build so that a consuming project can read it, and `RefSerializer` requires a
`SerializableDeclarationId` for that. So the two meanings that "durable" currently conflates have to be separated
rather than merged:

- *durable for retention*, which is a design-time concern and what this proposal makes scope-dependent;
- *durable for serialization*, which is neither, and must keep producing an identifier at the moment of writing.

A compile-time implementation that returned a live reference and left serialization to ask for the identifier when it
writes would satisfy both, but the boundary is the part that needs care: today a field typed `IDurableRef` carries the
promise that serializing it will work, and that promise would move to the serializer.

## Not measured

This is reasoned from the code paths above, not from a profile. Before acting on it, measure a batch build of a project
that uses `PullStrategy.IntroduceParameterAndPull` and a project that uses fabric queries, so that the size of the
constant is known rather than assumed. It may well be small enough not to be worth the added seam.

— Claude for @gfraiteur

gfraiteur:
## Summary

A durable reference exists so that an object outliving a request does not keep a compilation in memory. That is a design-time requirement: the analysis process is long-lived and Roslyn produces a new compilation on every keystroke. A batch build has one compilation that outlives everything it produces, so collapsing a reference to an identifier and resolving it again through the symbol table buys nothing there, and neither half of the round trip is free.

- Introduce `IDurableRefFactory`, resolved per project, with one implementation per scope. `SerializableDurableRefFactory` produces today's identifier-backed references; `LiveDurableRefFactory` produces a `LiveDurableRef<T>` that holds the reference it was made from.
- `LiveDurableRef` derives its identifier from the serializable reference rather than computing one, so both kinds carry the same identifier by construction. Asking a `FullRef` for `ToSerializableId()` directly answers with a *declaration* id even for a named type, which would silently drop type arguments and nullability (the defect of #1797) on the serialization path.
- An identifier-based reference caches the reference it last resolved to, through a weak reference, so a repeated resolution against a live compilation does not walk the symbol table again. `MetalamaDurableRefKind` overrides the scope's choice, including disabling that cache.
- The scope discriminator is a new internal `ExecutionScenario.IsBatchCompilation`, **not** `!IsDesignTime`: `Introspection` is not design-time, but `Metalama.Framework.Workspaces` and the aspect explorer keep many compilations alive, so live references there would be a genuine leak.
- `UserCodeRetentionPolicy.IsPinning` treated every `IDurableRef` as reaching nothing, and it runs *only* at compile time, which is exactly where live references are now the default. It now consults `IDurableRefImpl.ReachesCompilation`, and `MetalamaDiagnoseMemoryLeaks` forces the serializable kind, so the diagnostic keeps analysing the graph design time would produce.

## Test coverage

An aspect test is a batch compilation, so the aspect suite would have stopped covering identifier resolution. The `.4.12.0` aspect project now runs with `MetalamaDurableRefKind=SerializableWithoutCache` and is listed in `Metalama.Framework.LatestRoslyn.slnf`, the filter that already drives the test run.

Registering it revealed that **the `.4.12.0` test projects were built on every CI run and none of their tests were ever executed**, and that 78 of them were failing. The same 78 failed with and without this change, so they are unrelated to it. Fixed in three steps:

1. `TestRunnerFactory` resolved the factory type through a name that four `metalamaTests.json` files qualify with `Metalama.Testing.AspectTesting`, which in the variant is `Metalama.Testing.AspectTesting.4.12.0`. The assembly qualification is now a hint.
2. The HTML tests need `Metalama.Extensions.HtmlWriter`, which the variant deliberately does not reference: `ForbiddenConstants: [ROSLYN_4_12_0]`.
3. The extension-block tests need C# 14, which Roslyn 4.12 cannot parse: `RequiredConstants: [ROSLYN_5_0_0_OR_GREATER]`.

`RefTests` is now an abstract base run once per kind (`LiveRefTests`, `SerializableRefTests`, `UncachedSerializableRefTests`); the tests that resolve identifiers directly moved to `SerializableRefResolutionTests`, which runs once. New suites cover the resolution cache, serialization round-trips, and a control recording that a durable reference of a batch compilation does retain its compilation.

## Behavioural change, not a compilation break

No public signature changes, so no user code stops compiling. What changes is what a durable reference *holds* during a batch build. The case to be aware of: on .NET Framework `msbuild.exe` the `CompileTimeDomain` never unloads, so a user fabric storing a durable reference in a `static` field would pin one compilation per project across a persistent compiler server. It is bounded by project count rather than by keystrokes, and `MetalamaDurableRefKind` is the kill switch.

## Not done

The issue asks for a measurement before acting, and it has not been produced yet. The nopCommerce build benchmark is the right instrument, and `MetalamaDurableRefKind` allows an A/B on identical binaries. That work is in progress and is worth having before this merges, because if the weak resolution cache alone captures most of the win, the scope machinery could be dropped.

## Issues Fixed

Fixes #1811 - Optimize durable references in compile-time builds


Metalama.Framework/docs/design-time-memory.md
```diff
@@ -66,10 +66,37 @@ A reference obtained from the code model is bound to the compilation it came fro
...
-`IRef<T>.ToDurable()` converts it to an `IDurableRef<T>`, backed by a `SerializableDeclarationId` or a
-`SerializableTypeId`, which reaches nothing. Resolve it later with `GetTarget( compilation )` against whichever
-compilation is current. When what is at hand is the declaration rather than a reference, call
-`declaration.ToDurableRef()`: it produces an equal reference and never allocates the compilation-bound one.
+`IRef<T>.ToDurable()` converts it to an `IDurableRef<T>`, which may be held for as long as its holder lives. Resolve it
+later with `GetTarget( compilation )` against whichever compilation is current. When what is at hand is the declaration
+rather than a reference, call `declaration.ToDurableRef()`: it produces an equal reference and never allocates the
+compilation-bound one.
+
+**What a durable reference stores depends on the scope, and only the guarantee is common to both.** The rule above is a
```
gfraiteur: Remove AI slop, be accurate.Don't lead with mysteries. Update CLAUDE.md with copywriting rules"
- Use accurate software engineering language. Don't use analogies or slang.
- Prefer professional, midly formal language.
- Don't lead with mysteries or rhetorical constructs.
- No uncommon acronyms or abbreviations.
- Assume the reader is not a native english speaker

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/References/DurableRef.cs
```diff
@@ -3,28 +3,132 @@
...
 
 namespace Metalama.Framework.Engine.CodeModel.References;
 
+/// <summary>
+/// The base implementation of <see cref="IDurableRef{T}"/>, that is, of a reference that may be stored in an object
+/// outliving the run that produced it.
+/// </summary>
+/// <remarks>
+/// <para>
+/// This class stores nothing. Each derived class owns the storage its own <see cref="Id"/> comes from, because the two
```
gfraiteur: Inaccurate copywriting.

Metalama.Framework/docs/design-time-memory.md
gfraiteur: Done, and the rules are now in `CLAUDE.md` under Code Style, as a "Copywriting rules" list: be accurate; use accurate software engineering language and no analogies or slang; do not lead with a mystery or a rhetorical construct; no uncommon acronyms or abbreviations; assume the reader is not a native English speaker; no bold or italics for emphasis inside a paragraph. Each rule carries a concrete counter-example taken from this pull request, so the next reader sees what the rule is prohibiting.

This section is rewritten. It now opens with the subject rather than with a question:

> A durable reference stores a serializable identifier at design time, and stores the original reference during a batch compilation.

The rest is reworded in the same direction: "does not reduce memory consumption in that scenario, and both operations have a cost" replaces "buys no retention and is not free", the bold lead-ins are gone, and the two trailing points are ordinary paragraphs rather than a list introduced by "Two consequences are worth knowing".

I applied the same pass to every piece of prose in the pull request: the XML documentation, the code comments, the Markdown documents, the MSBuild comments and the test documentation. Roughly forty passages changed.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/References/DurableRef.cs
gfraiteur: You are right, and this one was inaccurate rather than merely badly written. The summary said "This class stores nothing" immediately above the declaration of the `_resolvedRefCache` field, so it was false as written.

The remarks now state what is actually true, which is that `Id` is abstract because the derived classes obtain it differently:

> `Id` is abstract, because the derived classes obtain it differently. `DeclarationIdRef<T>` and `TypeIdRef<T>` store the identifier and hold no reference to a compilation. `LiveDurableRef<T>` stores the `IFullRef<T>` it was created from, and computes the identifier from it when the identifier is requested.

The documentation of the field is rewritten as well. It described the field as "a cache and nothing else" and used "Losing this field to a collection", which are the constructs you flagged in the other comment. It now says that the field is a cache, that no other member reads it to produce a value that must always be available, and that the next resolution resolves the identifier again after the garbage collector clears it.

— Claude for @gfraiteur

Metalama.Framework/docs/testing.md
```diff
@@ -27,7 +27,7 @@ Two supporting projects are not test suites themselves but underpin the above: [
...
-- `Metalama.Framework/Metalama.Framework.LatestRoslyn.slnf` (`IsTestOnly = true`) — this **solution filter is what actually runs the Framework tests**, against the latest Roslyn only. It excludes the `.4.12.0` variant projects for speed.
+- `Metalama.Framework/Metalama.Framework.LatestRoslyn.slnf` (`IsTestOnly = true`) — this **solution filter is what actually runs the Framework tests**. It excludes the `.4.12.0` variant projects for speed, with one exception: `Metalama.Framework.Tests.AspectTests.4.12.0` is listed, so the aspect test suite also runs against Roslyn 4.12.0 and with `MetalamaDurableRefKind=SerializableWithoutCache`. There are two reasons for this exception. The `.4.12.0` projects were compiled by every run of the build, but none of their tests were executed. In addition, an aspect test is a batch compilation, and the durable references of a batch compilation store the reference they were created from, so the identifier resolution code would otherwise no longer be covered by this suite. See issue [#1811](https://github.com/metalama/Metalama/issues/1811).
```
gfraiteur: don't mention any specific roslyn version, just mention that the LatestRoslyn filter includes only the projects built against the latest roslyn version.

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/References/DurableRefFactory.cs
```diff
@@ -18,37 +27,24 @@ public static IDurableRef<T> FromTypeId<T>( SerializableTypeId id )
...
-    /// <c>ICovariant&lt;Derived&gt;</c> would come back as <c>ICovariant&lt;T&gt;</c>, and it includes the nullable
-    /// annotation, so <c>IService?</c> would come back as <c>IService</c>. Neither loss is reported, and the caller has
-    /// no reason to think the conversion is lossy. See issue #1797.
-    /// </para>
-    /// </remarks>
+    /// <inheritdoc cref="SerializableDurableRefFactory.FromDeclarationOrType{T}"/>
     public static IDurableRef<T> FromDeclarationOrType<T>( ICompilationElement declarationOrType )
         where T : class, ICompilationElement
-        => declarationOrType switch
+        => SerializableDurableRefFactory.Instance.FromDeclarationOrType<T>( declarationOrType );
```
gfraiteur: Don't use `SerializableDurableRefFactory.Instance`. Instead use the RefFacfory from ICompilationElement.Compilation.

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/References/DurableRefFactory.cs
```diff
@@ -18,37 +27,24 @@ public static IDurableRef<T> FromTypeId<T>( SerializableTypeId id )
...
+
+    /// <summary>
+    /// Returns the factory that implements a given <see cref="DurableRefKind"/>.
+    /// </summary>
+    /// <exception cref="ArgumentOutOfRangeException">
+    /// <paramref name="kind"/> is <see cref="DurableRefKind.Default"/>. That value designates no implementation,
+    /// because <c>AspectPipeline</c> then selects the implementation according to the execution scenario.
+    /// </exception>
+    public static IDurableRefFactory GetFactory( DurableRefKind kind )
```
gfraiteur: This method is a smell, should be removed.

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/References/DurableRefFactory.cs
gfraiteur: Done. `FromDeclarationOrType` now obtains the factory from the compilation of its argument:

```csharp
public static IDurableRef<T> FromDeclarationOrType<T>( ICompilationElement declarationOrType )
    where T : class, ICompilationElement
    => ((CompilationModel) declarationOrType.Compilation).RefFactory.DurableRefFactory
    .FromDeclarationOrType<T>( declarationOrType );
```

This is a behavior change and not only a cleanup. The method previously forced the serialized representation on every caller, so a batch compilation that called it received a reference of the wrong representation. It now returns the representation selected for the project that declares the argument.

`FromDeclarationId` and `FromTypeId` still create identifier-based references directly. They receive an identifier and not a declaration, so there is no compilation to obtain a factory from, and no reference to store. Their documentation now states that.

— Claude for @gfraiteur

Metalama.Framework/src/Metalama.Framework.Engine/CodeModel/References/DurableRefFactory.cs
gfraiteur: Removed. It had a single caller, `ServiceProviderFactory.WithProjectScopedServices`, which reads the `MetalamaDurableRefKind` option. The mapping is now a switch expression at that call site, where the option is read, so there is no longer a function whose only purpose is to translate an enumeration value into an instance.

The `ArgumentOutOfRangeException` for `DurableRefKind.Default` is gone with it. That value is excluded by the enclosing condition, so the unreachable case is now an `AssertionFailedException`.

— Claude for @gfraiteur

Metalama.Framework/docs/testing.md
gfraiteur: Rewritten without any version number:

> It includes only the projects built against the latest Roslyn version, with one exception: the aspect test suite is also listed in its variant built against the previous supported Roslyn version, which additionally sets `MetalamaDurableRefKind=SerializedWithoutCache`.

One point to confirm, because your wording and the current state of the filter do not agree. The filter no longer contains only the projects built against the latest Roslyn version: I added `Metalama.Framework.Tests.AspectTests.4.12.0` to it, which is what makes those tests run at all. I therefore kept the exception in the sentence rather than stating a rule that the file contradicts.

If you would rather the filter did contain only the latest-Roslyn projects, then that entry should be removed. The consequence is that the aspect test suite would no longer run against the older Roslyn, and the identifier resolution path would lose the coverage this issue is about. Tell me which you prefer and I will apply it.

— Claude for @gfraiteur

gfraiteur: Removed from the filter, and the documentation now states the rule without an exception:

> a solution filter for fast local builds. It includes only the projects built against the latest Roslyn version. The variants built against the older supported Roslyn versions are reached through the solution.

I had the model of the build wrong: I took the filter to be what runs the tests in continuous integration, rather than a filter for fast local builds, and added the project to it on that basis. The filter is unchanged from its original content again, so the code inspection it also drives is unaffected as well.

The rest of the work that made that suite runnable is kept, because it is useful wherever the suite is run from:

- `TestRunnerFactory` resolves the `TestRunnerFactoryType` option against the assemblies that are loaded, instead of requiring the assembly name written in `metalamaTests.json` to match. Four `metalamaTests.json` files name `Metalama.Testing.AspectTesting`, which does not exist under that name in a variant project.
- The tests that require `Metalama.Extensions.HtmlWriter`, which the variant does not reference, and the tests that require C# 14, which the older Roslyn cannot parse, are skipped through `ForbiddenConstants` and `RequiredConstants`.
- The variant sets `MetalamaDurableRefKind=SerializedWithoutCache`, so the identifier resolution code is covered whenever that suite runs.

Together these took that suite from 78 failures to zero.

— Claude for @gfraiteur

PR #1919, opened 2026-09-03
Update engineering and move the build image to Visual Studio 2026

### LINKED ISSUE #1920: Copy the binary log of a failed nested reference-assembly build to a chosen directory
## Problem

When the nested reference-assembly build fails, Metalama reports `LAMA0082` and names the path of the binary log of that build. The message of the diagnostic can quote only a few lines of the output, because a Roslyn diagnostic cannot contain line breaks, so that binary log is the only artifact from which such a failure can be diagnosed after the fact.

`CompileTimeAssemblyLocator` writes the log into its cache directory, which is under the temporary directory of the machine. That location is unreachable in two situations that matter:

- A continuous integration agent that runs the build in a container discards the temporary directory together with the container, so the log is gone before anyone can collect it. An artifact rule of the build server cannot reach it either, because the temporary directory of the container is not mounted.
- A user who reports a failure has no straightforward way to attach the log, because its path contains a version and a hash of the project.

## Enhancement

Add the `MetalamaAssemblyLocatorBinaryLogDirectory` MSBuild property. When it is set and the nested build fails, the binary log is copied into that directory, which can be a location that outlives the build, such as the artifacts directory of a repository.

The property is empty by default, so the behaviour is unchanged unless it is set. Copying the log assists a diagnosis and is not required for one, so a failure to copy it is logged and does not replace the diagnostic that describes the failure of the nested build.

## Implementation

Implemented in #1919. The property follows the existing `MetalamaAssemblyLocator*` family, so it is declared in `MSBuildPropertyNames`, `IProjectOptions`, `DefaultProjectOptions`, `MSBuildProjectOptions`, `ProjectOptionsWrapper` and the compiler-visible property list, and it is consumed by `CompileTimeAssemblyLocator.CopyBinaryLog`.

The `Issue1744` standalone scenario sets it to the artifacts directory of the repository, from which the build publishes the log. That is how the failure of the diagnostic classifier fixed in the same pull request was diagnosed: without the copied log, the only information available was that the nested build had reported nine warnings and three errors.

## Note on the public interface

`IProjectOptions` is public, so the new member is a breaking change for an external implementation of that interface. An implementation that derives from `DefaultProjectOptions`, or that wraps another instance through `ProjectOptionsWrapper`, is unaffected.

— Claude for @gfraiteur

gfraiteur:
## Summary

Build image and engineering:

- Update PostSharp.Engineering from 2023.2.412 to 2023.2.422.
- Pin the Visual Studio Build Tools of the build image to 18.9.2, the Dev18 stable release. Engineering derives the container layer name from the major version, so the layer moves from `vs17` to `vs18`, and `MSBuildVersion` moves from 17.14 to 18.9 because the pin accepts only a matching installation. This follows `platform-support.md`, under which PB-2027.0 drops Visual Studio 2022.
- Pin the .NET SDK to 10.0.400, the version that Visual Studio 2026 18.9 installs. A single `dotNetSdkVersion` constant feeds both the container component and `global.json`, so the image and the pin cannot drift apart.
- Remove the .NET 9 SDK from the image. It was required only because the build project targeted `net9.0`, so that project now targets `net10.0`. The .NET 8 SDK disappears as well, because the checked-in Dockerfile was stale relative to the component list.
- Remove the .NET Framework 4.7.2 and 4.8 targeting packs and developer packs from the image. The .NET SDK obtains the reference assemblies of a .NET Framework target framework from the `Microsoft.NETFramework.ReferenceAssemblies` packages, which is why `net462` and `net481` build although no component installs their packs, and no project or build script uses a tool of the developer packs.

Defects found while making the above work, both of which affect users and not only continuous integration:

- Do not leak `MSBuildExtensionsPath` into a nested build. `DotNetTool` and `MSBuildTool` remove the environment variables that would make a child process build with a mixture of two .NET SDK installations, and this one was missing from both lists. MSBuild imports `NuGet.targets` through it, so a child that resolves its own software development kit but keeps the inherited value runs the targets of one feature band on the engine of another and fails with `MSB4062`. The other half of the fix is in engineering 2023.2.421.
- Quote a diagnostic that carries no message identifier in `LAMA0082`. The diagnostic quoted the last four lines of the failing nested build when no line carried an identifier, and the number of lines in the epilogue of a build depends on the version of MSBuild: the MSBuild of the .NET SDK 10.0.4xx prints the destination of the binary logger, which pushed the diagnostic out of the quoted window and left a message that reported only how many warnings and errors there were. Lines that begin with the position of a diagnostic are now quoted instead, which is invariant across languages for the same reason a message identifier is. This is the defect that the `Issue1744` scenario detected.

New option:

- `MetalamaAssemblyLocatorBinaryLogDirectory` copies the binary log of a failed nested reference-assembly build into a chosen directory. That log is the only artifact from which such a failure can be diagnosed after the fact, and a continuous integration agent that builds in a container discards the temporary directory together with the container. The `Issue1744` scenario sets it to the artifacts directory of the repository, from which the build publishes it.

Also:

- Simplify three null checks that the analyzers of the .NET SDK 10.0.4xx report as `IDE0270`, which `CodeQuality.targets` promotes to errors when `ContinuousIntegrationBuild` is true.
- Record the findings in `CLAUDE.md`, `Metalama.Framework/docs/platform-support.md` and `Metalama.Framework/docs/testing.md`.

## Breaking Changes

- `IProjectOptions` gains the `AssemblyLocatorBinaryLogDirectory` member. An external implementation of that interface has to add the member. An implementation that derives from `DefaultProjectOptions`, or that wraps another instance through `ProjectOptionsWrapper`, is unaffected.

## Issues Fixed

Fixes #1920 - Copy the binary log of a failed nested reference-assembly build to a chosen directory


gfraiteur:
Breaking change: `IProjectOptions` gains the `AssemblyLocatorBinaryLogDirectory` member.

The interface is public, so an external implementation of it has to add the member. An implementation that derives from `DefaultProjectOptions`, or that wraps another instance through `ProjectOptionsWrapper`, is unaffected, because both received the member in this branch. The member returns `null` by default, which keeps the previous behaviour: the binary log of a failed nested reference-assembly build stays in the cache directory unless a directory is named.

— Claude for @gfraiteur

Metalama.Framework/src/tests/Standalone/Issue1744/Issue1744.csproj
```diff
@@ -17,6 +17,12 @@
              every run exercises the failure rather than reading a cached result. See README.md. -->
         <MetalamaAssemblyLocatorHooksDirectory>$(MSBuildThisFileDirectory)AssemblyLocatorHooks</MetalamaAssemblyLocatorHooksDirectory>
         <MetalamaAssemblyLocatorSalt>Issue1744.HooksFailure</MetalamaAssemblyLocatorSalt>
+
+        <!-- The nested build writes its binary log under the temporary directory of the machine, which a continuous
+             integration agent discards together with the container that ran the build. That log is the only artifact
+             from which the failure can be diagnosed after the fact, so it is copied into the artifacts directory of
+             the repository, from which the build publishes it. -->
+        <MetalamaAssemblyLocatorBinaryLogDirectory>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)..\..\..\..\..\artifacts\logs'))</MetalamaAssemblyLocatorBinaryLogDirectory>
```
gfraiteur: Use $(RepoDirectory)

gfraiteur: Done in 9964989890. It now reads:

```xml
<MetalamaAssemblyLocatorBinaryLogDirectory>$([MSBuild]::NormalizeDirectory(`$(RepoDirectory)artifacts\logs`))</MetalamaAssemblyLocatorBinaryLogDirectory>
```

`RepoDirectory` is set by the `Directory.Build.props` of the solution, and the one under `Standalone` chains up to it, so it is available here. `NormalizeDirectory` follows the idiom of `BuildOptions.props` in the engineering software development kit.

Evaluating the property resolves it to `C:\src\Metalama-2027.0\Metalamartifacts\logs\`, which is the directory that the build publishes. That check was worth making explicitly: the copy happens only when the nested build fails, so a wrong path would have been silent instead of a build failure.

— Claude for @gfraiteur

Issue #1921, opened 2026-09-03
.NET 11 Support


gfraiteur:
This is a meta-issue to group all user stories, enhancements and bugs related to supporting .NET 11 in Metalama 2027.0.

Issue #1045, opened 2025-09-08
.NET 10 Support


gfraiteur:
This is a meta feature to store what's new with .NET 10

Issue #1039, opened 2025-09-08
C# 14


gfraiteur:
This is a meta feature to group all C# 14 features

PR #1925, opened 2026-09-04
Add the .NET 11, C# 15 and Roslyn 5.10 impact analysis and the 2027.0 decision record

gfraiteur:
This adds `Metalama.Framework/docs/2027.0/`, which analyses what .NET 11, C# 15 and Roslyn 5.10 require of Metalama and Metalama.Premium for the 2027.0 release, records the decisions taken about them, and proposes the user stories that deliver the work.

It is documentation only: 66 files under `Metalama.Framework/docs/2027.0/`, no product change. It is merged so that the decision record lives beside `platform-support.md` and `updating-roslyn.md`, and so that the links that the stories carry keep resolving once the stories become issues.

No issue has been created yet.

## What to read, and in what order

| File | What it is |
| --- | --- |
| `README.md` | The index, the verified inputs and the method |
| `DECISIONS.md` | The twelve decisions that govern the release, each with its reasoning |
| `OPEN-QUESTIONS.md` | The five questions that are not decided, each with options and an owner |
| `user-stories/` | The 37 proposed stories, one file per story, with an index and the ordering in its `README.md` |
| `01` to `08` | The theme documents, one per subsystem, which the stories cite |
| `analysis-reports/` | The raw analyses, kept as evidence and cited by the theme documents |

`DECISIONS.md` and `user-stories/README.md` are the two files worth reviewing closely. The theme documents are reference material for the stories rather than prose to read end to end.

## Method

Eight analyses mapped the impact per subsystem. Each of the 91 findings was then verified three times independently: a code lens that read the cited path end to end and tried to refute the claim, a semantics lens that checked the external premises against the Roslyn sources and the csharplang proposals, and a scope lens that checked the issue tracker for work already done. No finding was refuted. Nearly every one needed a detail corrected, which is what the second pass was for. A completeness review then asked what the analysis had missed and reported twelve gaps, of which seven became stories and the rest were cleared with a file reference each.

No project was built and no test was run. This is analysis only.

## The main conclusions

C# 15 support depends on a Roslyn that Metalama does not consume today. Reading the three release branches of `dotnet/roslyn` shows a train one minor apart: `release/stable` at 5.10, `release/insiders` at 5.11 and `main` at 5.12. Roslyn 5.10 has no `LanguageVersion.CSharp15`, maps `Latest` to C# 14, gates all six C# 15 features on `Preview`, and still marks the union declaration and the with element experimental in the grammar. Roslyn 5.11 is the first version that offers C# 15 as a supported language version.

Two defects reach users today and need no C# 15 to appear. A `net11.0` project whose language version the software development kit implies is silently rewritten to C# 12, with a warning that carries no code and gives the wrong reason. Every build with the .NET 11 software development kit reports that the kit is unsupported, because the ceiling `11.0` is compared against the full version `11.0.100`.

Unions break the pipeline in the syntax layer rather than in the code model. Roslyn lowers a union to a plain struct, so no enumeration switch crashes, but `UnionDeclaration` is missing from twelve hand-written kind lists and twenty visitors, which makes a `partial union` report as not partial, positions introduced members outside the type, and crashes the linker for any member-producing advice.

Introducing a case into a union has two forms that differ in kind. For a type carrying `UnionAttribute` it is ordinary constructor introduction and a generated partial part can express it. For a `union` declaration the case list lives in exactly one part, so the operation requires rewriting the part the user wrote, and the design-time result cannot be expressed at all.

The C# 15 engine work does not have to wait for the Roslyn move. The build this repository already consumes declares the union, closed-hierarchy, collection-argument and labeled-jump members and marks them `[RSEXPERIMENTAL006]`, so an opt-in that this repository controls unblocks the engine work now. Story S-31 adds that opt-in as a harness that ships disabled, and S-13 turns it off rather than deleting it, because Roslyn publishes the interface of a feature behind an experimental marker one or more releases before it declares the language version, so the same interval recurs at every language version.

## What the review changed

Five review rounds are answered and every thread is resolved. Six comments changed a conclusion rather than confirming it.

The criterion for a payload variant is a new Roslyn application programming interface, not a new language feature. The `closed` modifier shows why: it adds no syntax node, and `SyntaxKind.ClosedKeyword` is still a new enumeration member, so reading or emitting it needs a build against a Roslyn that declares it.

Roslyn 5.0 probably serves a serviced Visual Studio 18.0 and not only Rider and the C# Dev Kit. Its release branch carries Roslyn 5.0 and was still receiving commits in late August 2026, nine months after that version shipped, which a branch does not get unless the channel is serviced. This makes the lower variant harder to drop, and it casts doubt on the statement in `platform-support.md` that the long-term servicing channel opens only in November 2026. That document is not edited here.

The Roslyn move is a sequence of merges paced by the software development kit release candidates, not a single wait for a stable version, and a Roslyn that no kit has shipped is not worth adopting.

The .NET 11 software development kit is required, and `global.json` names it as the main kit of the product. The mechanism is the compile-time compilation rather than the target framework: the language version provider caps the compile-time language version by the major version of the installed kit, so with only the .NET 10 kit an aspect test cannot exercise a C# 15 construct in compile-time code. Story S-09 installs it. The `net11.0` target framework remains out of scope, because no .NET 11 interface justifies an asset or a test leg.

The payload variant that predates C# 15 answers false and reports nothing. The analysis had proposed a design-time diagnostic and had reasoned outward from the divergence without asking whether the code that produces it can exist in such a host. It cannot: a Roslyn without `LanguageVersion.CSharp15` reaches every C# 15 feature only under `Preview`, which `CompileTimeAspectPipeline.VerifyLanguageVersion` refuses, so the host has already flagged the code and a Metalama diagnostic would restate it.

`DECISIONS.md` states the decisions rather than their history. Fourteen numbered sections with revision sub-sections became twelve, each stating one decision in its final form, and the cross-references were renumbered in every citing document.

The later rounds also changed the shape of the story set. The union work became a meta-story, `S-18. C# 15 unions: support in target code`, with six sub-stories `S-18-1` to `S-18-6` merged as one stack of pull requests. The defect stories are titled by the failure rather than by the remedy. The three discretionary stories, which are the introduction of a closed class and of a union case, are last and form the cut line of the release. Two proposals were withdrawn: a guard against experimental syntax in templates, which cannot be written without naming an experimental Roslyn member and is unnecessary because the preview language version is already refused, and a rule that would route union declarations by type comparison rather than by kind.

## Corrections made during the work

These are recorded in the documents rather than quietly fixed, because each reversed an earlier conclusion.

The new C# 15 issue is a child of the `.NET 11 Support` meta issue #1921 rather than its sibling. The tracker settles it: issue #1039, `C# 14`, is a Feature whose parent is #1045, `.NET 10 Support`, with twenty sub-issues.

Roslyn 5.10 is the current stable branch rather than a discarded state of `main`, which an earlier section had inferred from the package index alone.

The boundary that decides the payload variant set sits between Roslyn 5.10 and 5.11, not between 5.11 and 5.12.

The `MSBuildExtensionsPath` mitigation in PostSharp.Engineering is complete, in 2023.2.421. This repository pins 2023.2.420, so the container story takes the dependency rather than writing the fix.

## What needs a decision

One question blocks one story, and it is written up with options and a recommendation in `OPEN-QUESTIONS.md`: whether case introduction into a `union` declaration ships, given that it works at build time only and the editor cannot show the added case.

Everything else waits on the measurement that `platform-support.md` already requires after 2026-11-10, which section 7 of `DECISIONS.md` sharpens to one question: does any Visual Studio in support on 2027-01-31 present a Roslyn whose non-experimental features need an interface an older Roslyn lacks.

Three statements of `platform-support.md` follow from the analysis and are recorded rather than applied, because that document is the doctrine.

## After this merges

The 37 stories become issues, grouped under a new "C# 15" feature issue that is itself a sub-issue of #1921. Ten of them target `Metalama.Premium`, `Metalama.Vsx`, `Metalama.Samples`, `Metalama.Documentation` or `Metalama.Compiler` rather than this repository. Once the issues exist, the 37 story files are replaced by their issue links, and `user-stories/README.md` stays as the index, the ordering and the record of the thirteen planning decisions, which the issues do not carry.

The work-in-progress snapshot of the open pull requests, which was scaffolding for the analysis and would be wrong within weeks, was removed before merging.

https://claude.ai/code/session_01XCPq4KxXAha1Hz14dcD9ca


Metalama.Framework/docs/2027.0/README.md
```diff
@@ -0,0 +1,118 @@
...
+```
+PB-2027.0 = VS 2026 LTSC · VS Code C# Dev Kit / Rider current · .NET 10 SDK · User=net10.0 ·
+            .NET Framework 4.7.2 · Roslyn 5.0–5.x · Core=net10.0 / Desktop=net472
+```
+
+Two consequences of the baseline drive this analysis.
+
+- Visual Studio 2027 ships in November 2026 with .NET 11 and C# 15, and both it and the Visual Studio 2026 long-term
+  servicing channel present a Roslyn at or above 5.10. The latest payload variant, `Roslyn.5.10.0`, serves them, and
+  the `Roslyn.5.0.0` variant serves Rider and the C# Dev Kit.
```
gfraiteur: I think Roslyn 5.0 also serves an older LTSC version of Visual Studio that will still be supported on 01-31-27. Verify.

Metalama.Framework/docs/2027.0/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+The list below is read from `MessageID.RequiredVersion` in `src/Compilers/CSharp/Portable/Errors/MessageID.cs` on
+the `main` branch of `dotnet/roslyn` on 2026-09-03, and from the grammar file
+`eng/src/GenerateMetaSyntaxRewriter/Syntax-5.10.0.xml` in this repository. The Roslyn gate column names the gate on
+`main`; in the consumed 5.10 preview every feature is gated on `LanguageVersion.Preview`.
+
+| Feature | Roslyn gate | New syntax in the 5.10 grammar | Proposal |
+| --- | --- | --- | --- |
+| Union types | `LanguageVersion.CSharp15` | `UnionDeclarationSyntax`, a new `TypeDeclarationSyntax`; kinds `UnionDeclaration` and `UnionKeyword` | `dotnet/csharplang`, unions |
+| Closed hierarchies | `LanguageVersion.CSharp15` | None. The `closed` modifier is a token of kind `ClosedKeyword` in the modifier list | `dotnet/csharplang`, closed hierarchies |
```
gfraiteur, line 66: A new token type is relevant from our perspective since it's an API addition. If we need a new API to cater for a feature, then we need a build of our projects for that Roslyn API version (although Metalama.Framework does not care of the roslyn implementation itself)

Metalama.Framework/docs/2027.0/user-stories.md
```diff
@@ -0,0 +1,3068 @@
...
+- Option A, full support. Stories S-09 and S-11, and the dependent feature stories S-12 to S-22, must be written,
+  reviewed and tested in roughly six weeks after the stable Roslyn exists. None of the feature work can be validated
+  before it, because a test that requests C# 15 today is skipped and the suite still reports success.
+- Option B, the plumbing and a defensive subset of the work. A C# 15 project builds, reports the designed diagnostics
+  instead of crashing, and is not silently downgraded by the MSBuild clamp, while the language features are
+  documented as unsupported.
+- Option C, defer the Roslyn move. This option is not viable, because three published packages would keep a
+  dependency on a Roslyn package that nuget.org does not serve, which is the restore failure of #1106.
+
+Answer: Option A, recorded in [`DECISIONS.md`](DECISIONS.md) section 1. The schedule risk of the window between the
```
gfraiteur: The decision is to support all C# 15 features (non-experimental) in __user code__ i.e. ** all existing Metalama ad vising and code model features must work with any new C# 15 feature**. However adding support for these features in the __advising__ API (typically introductions) is discretionary and depends on how we meet the deadlines.

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
```diff
@@ -0,0 +1,183 @@
...
+a `closed` modifier that an aspect emits at build time is absent at design time. Nothing reports the difference.
+
+- Option A, stay silent. This is the behaviour that follows from doing nothing. The editor and the command line
+  disagree and the user has no indication of it, which is the failure mode that the platform doctrine exists to
+  prevent.
+- Option B, report. The draft in `analysis-reports/12-csharp15-api-drafts.md` proposes a design-time warning
+  reported once per project, with an opt-out, from the diagnostic analyzer. It fires in an editor whose user can
+  act on it only by changing the integrated development environment.
+
+The draft recommends Option B. One measurement could change the answer, which is Q6. Settled by the product owner.
```
gfraiteur: This has been recently implemented.

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
```diff
@@ -0,0 +1,183 @@
...
+case added to a `union` declaration under Option A of Q1. Both are design-time only, both fire in situations the
+user cannot repair from the editor, and both therefore need a severity and an opt-out chosen deliberately. The
+drafts propose a warning with an opt-out for the first. Settled when the stories are written.
+
+### Q4. Does an aspect introduce a closed class?
+
+Answered on 2026-09-04 in section 5f of [`DECISIONS.md`](DECISIONS.md): yes. A closed class is an ordinary class
+with one more modifier, and the analysis had already found every part of the writer identified and cheap. The
+answer raises the stake of Q2, because an aspect that introduces a closed class emits the modifier at build time
+and nothing at design time on the hosts that the lower Roslyn variant serves.
```
gfraiteur: Seems a simple but low-priority feature to me.

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
```diff
@@ -0,0 +1,183 @@
...
+else in the 5 generation, and the `main` branch reads minor version 12. The derivation is an inference from a
+cadence and the checklist requires a measurement against a real installation after 2026-11-10. The private runtime
+must be measured at the same time, because the embedded Core flavour of the payload depends on it.
+
+### Q6. Which Roslyn version does a current Rider present, and the Visual Studio Code C# Dev Kit?
+
+Blocking, for the variant set, and already tracked as checklist item 2 of
+[`platform-support.md`](../platform-support.md).
+
+Rider was measured at Roslyn 5.0.0 on 2026-09-01 and the C# Dev Kit was not measured. Rider builds its own Roslyn
```
gfraiteur, line 54: The question does not really matter unless Rider relies on a Roslyn version that no supported version of VS uses as of 01-31-27 and that has non-experimental features requiring new APIs.

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
```diff
@@ -0,0 +1,183 @@
...
+collections cached against it, that a builder must be equated with the symbol it becomes for a type whose members
+the compiler synthesizes, and that the order of synthesized members must be deterministic, which is a defect that
+pull request #1879 had to fix for records. Every one of those hazards was refuted except the ordering of aspect
+instances, which fails for a different reason than the one anticipated.
+
+### Q8. How is a closed type from a referenced assembly handled?
+
+Not blocking. Section 9 records that the existing derived type index already gives the complete set of derived
+types for a closed type of the current compilation, because the language requires every subtype to be in the same
+module. A closed type that comes from a referenced assembly is the case the index does not answer, and the drafts
```
gfraiteur: Not important to us because Metalama makes no guarantee to generate valid C# code and verify all C# preconditions. It remains the responsibility of the aspect author.

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Coverage gaps found by the completeness review
+
+The synthesis ended with a review that asked what the analysis had missed. Twelve gaps were reported and are
+recorded here rather than silently closed. Each names where to look. None of them invalidates a finding or a story;
+they are places the analysis did not reach.
+
+### Q12. Static members in interfaces has no story
+
+It is one of the six features gated on `LanguageVersion.CSharp15`, and the only one that no story delivers. The
```
gfraiteur: Story should be filed, or this paragraph is stale.

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
```diff
@@ -0,0 +1,183 @@
...
+Roslyn feature status row describes it as non-virtual static interface members on runtimes without default
+interface implementation support, which is why it matters here: the aspect test project has a `net48` leg. Either a
+story delivers it, following the test directory convention that the C# 15 language version story establishes, or
+the analysis records why none is needed.
+
+### Q13. Three neighbouring repositories are absent from the analysis
+
+`Metalama.Vsx` appears nowhere, although two package pinning rules of `Directory.Packages.md` are derived from the
```
gfraiteur: Stale?

Metalama.Framework/docs/2027.0/user-stories.md
```diff
@@ -0,0 +1,3068 @@
...
+  is known to be fragile, because two bands under one `dotnet` directory already produced an `MSB4062` restore
+  failure through a stale `MSBuildExtensionsPath`. That mitigation is now complete: the merged pull request #1919
+  removes `MSBuildExtensionsPath` in `DotNetTool.cs:61` and in `MSBuildTool.cs:55`, and the matching change is in
+  PostSharp.Engineering 2023.2.421. The remaining objection to this option is therefore the cost of a second feature
+  band in the image, and not a missing mitigation.
+- Option B, install the .NET 11 SDK alone. One band and no conflict, but the product is then built by a compiler
+  whose default language version differs from the one the tests assume.
+- Option C, keep one SDK and accept that `net11.0` stays declared and untested.
+
+Answer: Option C, recorded in [`DECISIONS.md`](DECISIONS.md) sections 6b and 6c. The container change has no
```
gfraiteur: Will be tested through Metalama.Tests.DotNetSdk.

Metalama.Framework/docs/2027.0/user-stories.md
```diff
@@ -0,0 +1,3068 @@
...
+
+`Metalama.Framework.Workspaces`, `Metalama.Testing.AspectTesting` and `Metalama.LinqPad` declare a dependency on a
+Roslyn version that nuget.org does not serve. That failure already reached a user in #1106.
+
+- Option A, gate general availability only, which reproduces that failure for anyone who restores a preview.
+- Option B, gate every public publication, which means either moving to the stable Roslyn earlier or renumbering the
+  latest variant to the stable 5.9.0, which exposes the same public application programming interface as the consumed
+  preview and is available today, and adding a build-time check that refuses to pack a prerelease Roslyn pin.
+
+Recommendation: Option B. It removes a class of user-facing failure at the cost of one interim renumbering. Option B
```
gfraiteur: We are going to follow their RC builds and update the SDK and Roslyn every time. After they publish .NET SDK GA, we will wait until they also publish Roslyn 5.12 public packages, and they are usually late with that.

Metalama.Framework/docs/2027.0/user-stories.md
```diff
@@ -0,0 +1,3068 @@
...
+1. The cheap work that removes uncertainty or repairs a defect that users meet today. None of it is blocked by
+   anything. S-02, the variant gating, is first of all, because it is a decision, it costs little, and eight later
+   stories cannot start until it is taken. S-04 repairs the two MSBuild comparisons that misreport a supported .NET
+   SDK and silently downgrade the language version of a `net11.0` project, which users meet now. Then S-06, the
+   language version display mapping, S-05, the residue of the previous baseline, S-31, the untested refactoring
+   entry points, S-35, the host process classification, S-24, the `Metalama.Premium` build-file residuals, and
+   S-27, the change-visibility code action of `Metalama.Premium`. Then the two medium ones that remove a class of
+   silent failure rather than a single defect: S-03, the type-declaration predicate, and S-07, the declaration-kind
+   switches.
+2. S-01, the move of `Metalama.Compiler` to the stable Roslyn 5.12. This is the exception to the rule that a large
```
gfraiteur: We will need to move Metalama.Compiler to a preview Roslyn 5.12 otherwise we won't be able to make progress in the preview. Moving to the stable Roslyn branch is an RC requirement for us.

Metalama.Framework/docs/2027.0/user-stories.md
```diff
@@ -0,0 +1,3068 @@
...
+  S31["S-31 Code refactoring provider entry points"]
+  S32["S-32 Visual Studio Tools and the flowed pins"]
+  S33["S-33 Sample solutions on PB-2027.0"]
+  S35["S-35 Host process classification"]
+```
+
+## Stories
+
+### S-01. Move Metalama.Compiler to the stable Roslyn 5.12
+
```
gfraiteur: We can't wait for the stable Rolyn 5.12. We will have to merge many times at each SDK RC. However I don't see the point of updating to a version that has never been shipped in the SDK.

Metalama.Framework/docs/2027.0/README.md
gfraiteur: Verified as far as this session can, and the evidence supports you. Recorded in section 14.2 of `DECISIONS.md`.

The branch `release/dev18.0` of `dotnet/roslyn` exists, its `eng/Versions.props` reads major 5 and minor 0, and it was still receiving commits on 2026-08-25, which is nine months after Visual Studio 18.0 shipped. Two dependency-flow branches for the same release were updated on 2026-08-26 and 2026-08-31. A release branch does not get serviced for a channel nobody can install, so a serviced Visual Studio 18.0 almost certainly exists and presents Roslyn 5.0.

That contradicts the statement in `platform-support.md` that the Visual Studio 2026 long-term servicing channel opens only in November 2026 and is therefore the first pinnable version. I have not edited that document, since it is the doctrine, but section 14.2 records that the statement is doubtful.

The consequence is that the `Roslyn.5.0.0` variant is required by a Visual Studio host and not only by Rider and the C# Dev Kit. That makes it harder to drop and makes the divergence of Q2 a Visual Studio problem too.

What I could not do is read the support end date. The Visual Studio 2026 servicing page states it directly and this session's network policy blocks that site, so the measurement in the platform checklist is still needed.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/README.md
gfraiteur: Agreed, and this sharpens the criterion rather than merely correcting the table. Recorded in section 14.1 of `DECISIONS.md`.

The question is not whether a supported host presents a Roslyn with a non-experimental C# 15 feature, but whether it presents one whose non-experimental features need an interface that an older Roslyn does not have. The `closed` modifier is the case that shows why the distinction matters: it adds no syntax node, which is why the table recorded no new syntax, and yet `SyntaxKind.ClosedKeyword` is a new enumeration member. Reading or emitting the modifier therefore requires a build against a Roslyn that declares it, exactly as you say.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories.md
gfraiteur: Recorded as section 12 of `DECISIONS.md`, which states the doctrine in your words: Metalama must be able to advise any user code, and exposing a new language feature in the advising and introduction interfaces is a separate decision taken feature by feature.

It changed the plan rather than confirming it. I had proposed moving the union work to 2027.1 to reduce scope, and the doctrine makes that unavailable, because most of that work is the advising path. In particular the story that injects and links advice applied to a union is required, not discretionary: deferring it does not stop users writing unions, it only leaves Metalama crashing when it meets one.

What may move is the introduction interface, which is one large story and two medium ones. Section 13 marks those three as nice to have, and each carries a priority field naming them as the first candidate to slip to 2027.1.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories.md
gfraiteur: Understood, and this reverses what the story assumed. Recorded in section 14.3 of `DECISIONS.md`, together with your two related comments.

The story treated the move to a stable Roslyn as the starting condition for everything downstream. You are describing the opposite: the move to a preview Roslyn 5.12 is what makes progress possible during the preview, and the stable branch is a release candidate requirement rather than a gate on starting.

That also removes the scheduling worry the analysis had, which was that the whole C# 15 plan waited on a November Roslyn and left about six weeks. It does not remove the cost, since the measured work in that repository is a merge of an upstream branch rather than a rebase.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories.md
gfraiteur: Recorded in section 14.3 of `DECISIONS.md`. The story described a single move; the practice you describe is a sequence of merges paced by the release candidates, with a wait after general availability for Microsoft to publish the Roslyn packages, which they are usually late in doing. The story now reflects that.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories.md
gfraiteur: Agreed, and the last sentence retires a question the analysis had spent some effort on. Recorded in section 14.3 of `DECISIONS.md`.

I had derived that the latest variant should be renumbered to 5.12 because 5.10 and 5.11 serve nobody, then corrected that when the branch train showed 5.10 is the current stable. Your rule settles it without needing either derivation: a Roslyn that no software development kit has shipped is not worth adopting, whatever the branch train says.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
gfraiteur: Noted, and Q2 is updated to say so. The question is now whether to report rather than how, since the mechanism exists.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
gfraiteur: That matches the analysis, which sized the writer at medium and found every part of it identified and cheap. Q4 now records it as simple and low priority, which places it first among the three discretionary stories of section 13 and therefore the likeliest to survive a cut.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
gfraiteur: Narrowed accordingly in Q6. It matters only if Rider relies on a Roslyn that no supported Visual Studio uses on 2027-01-31 and whose non-experimental features need an interface an older Roslyn lacks.

Your other comment makes that less likely still: if Roslyn 5.0 serves a serviced Visual Studio 18.0, the lower variant is required whatever Rider presents, and the measurement stops being able to remove it.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
gfraiteur: Understood, and Q9 now records that answer with your reasoning. It also removes a proposal the analysis had made: the blanket eligibility rule that would refuse advice on a union with one clear message was justified by the fear of emitting code the compiler rejects, and under this rule that is the aspect author's responsibility rather than ours.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
gfraiteur: Stale, and now fixed. Story S-30 was filed for it after this paragraph was written. The investigation found that the feature adds no syntax, appears in only two Roslyn files, and needs no product change, so the story is test coverage on the .NET Framework leg. Q12 says so and keeps the original question, because it records why the story exists.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
gfraiteur: Stale, yes. Stories S-32, S-33 and S-34 cover Visual Studio Tools for Metalama, the samples and the conceptual documentation. Q13 is marked answered, and the same correction applies to Q14, Q16 and Q17, which the review did not flag but which were stale for the same reason: they were written before the stories existed.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories.md
gfraiteur: Good, and that confirms from a different direction the conclusion that no `net11.0` leg belongs in this repository's own matrix. Recorded in section 14.4.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-02-variant-gating-decision.md
```diff
@@ -0,0 +1,65 @@
...
+
+- Issue type: User Story
+- Labels: `enhancement`, `Area-Build-Engineering`, `Area-Framework`
+- Milestone: `2027.0`
+- Repositories: `metalama/Metalama`
+- Size: S
+- Blocked by: nothing
+- Findings: [CM-10](../03-code-model-unions-closed.md)
+
+---
```
gfraiteur: This story is written in a cryptic way. It should just say to move (rename) Roslyn 5.10 to 5.12 in source code I think.

Metalama.Framework/docs/2027.0/user-stories/S-03-type-declaration-by-type-test.md
```diff
@@ -0,0 +1,63 @@
...
+- Repositories: `metalama/Metalama`
+- Size: M
+- Blocked by: nothing
+- Findings: [CM-2](../03-code-model-unions-closed.md), [CM-6](../03-code-model-unions-closed.md),
+  [LK-3](../04-linker-and-advice.md), [DT-1](../05-design-time-workspaces-linqpad.md),
+  [DT-6](../05-design-time-workspaces-linqpad.md)
+
+---
+
+`SyntaxKindExtensions.IsTypeDeclaration` at
```
gfraiteur: This is against our pattern of routing declarations, nodes, ... by kind and not by type comparison. I don't see the real use of this story.

Metalama.Framework/docs/2027.0/user-stories/S-02-variant-gating-decision.md
gfraiteur: I see S-09 and I don't understand how this story differs.

Metalama.Framework/docs/2027.0/user-stories/S-18-2-union-rewriters-and-visitors.md
```diff
@@ -0,0 +1,56 @@
+### S-13. Give the compile-time path and the design-time classifier a union dispatch
```
gfraiteur: The title is quite bad. Do you mean to implement unions in rewriters and visitors?

Metalama.Framework/docs/2027.0/user-stories/S-18-3-union-linker-and-advising.md
```diff
@@ -0,0 +1,67 @@
+### S-14. Inject, link and validate advice applied to a union, and read its synthesized members
```
gfraiteur: Bad title. Do you mean that the linker should support advising unions?

gfraiteur: There should be a meta-story to support unions in target code since it seems to be the subject of several stories here. And it should probably be implemented as a PR stack, merged as one.

gfraiteur: So consider a numbering like this: S-13 is the support for union types in target code, S-13-1, S-13-2 are the sub stories.

Metalama.Framework/docs/2027.0/user-stories/S-29-introduce-union-and-case-attribute-form.md
```diff
@@ -0,0 +1,83 @@
+### S-17. Introduce a union type and introduce a case into a type carrying the union attribute
+
```
gfraiteur, line 2: Bad ordering. Should be at the end since it's nice to have.

Metalama.Framework/docs/2027.0/user-stories/S-16-closed-modifier-in-code-model.md
```diff
@@ -0,0 +1,51 @@
+### S-18. Read closed hierarchies in the code model
```
gfraiteur: The title would typically be: `Code Model: support for closed modifier`. Read user story titles of 2026.0 for inspiration and conventions

Metalama.Framework/docs/2027.0/user-stories/S-19-reject-labeled-break-in-templates.md
```diff
@@ -0,0 +1,59 @@
...
+- Milestone: `2027.0`
+- Repositories: `metalama/Metalama`
+- Size: M
+- Blocked by: S-02, S-09, S-11
+- Findings: [TP-3](../02-syntax-generator-and-templates.md), [LK-9](../04-linker-and-advice.md),
+  [UT-17](../06-user-tfm-patterns-tests-docs.md)
+
+---
+
+`TemplateAnnotator` has no visit for a labeled statement, so the scope of a labeled loop is not propagated and the
```
gfraiteur: I think we already reject any labels in templates.

Metalama.Framework/docs/2027.0/user-stories/S-28-introduce-closed-class.md
```diff
@@ -0,0 +1,48 @@
...
+
+- Add the settable closed property to `INamedTypeBuilder`, with the getter on `INamedType` delivered by S-18.
+- Validate in the setter that the type kind is a class and that the type is neither sealed nor static, which are the
+  restrictions the language states.
+- Store the value in `NamedTypeBuilderData` and expose it on `IntroducedNamedType`.
+- Emit the closed keyword in `GetTypeSyntaxModifierList` and suppress `abstract` there, while `IsAbstract` keeps
+  reporting true.
+- Emit the modifier before `partial`, because `partial` must sit immediately before the type keyword.
+- Gate the reference to `SyntaxKind.ClosedKeyword` on the latest Roslyn variant, per S-02.
+- Report a diagnostic rather than emitting the modifier when the target compilation provides neither
```
gfraiteur: We generally don't do that, we let the compilation fail if polyfills are missing.

Metalama.Framework/docs/2027.0/user-stories/S-20-experimental-syntax-guard-in-templates.md
```diff
@@ -0,0 +1,51 @@
...
+- Labels: `enhancement`, `Area-Framework-Templates`
+- Milestone: `2027.0`
+- Repositories: `metalama/Metalama`
+- Size: M
+- Blocked by: S-13, S-15
+- Findings: [TP-4](../02-syntax-generator-and-templates.md), [TP-5](../02-syntax-generator-and-templates.md)
+
+---
+
+A collection expression with a with-element crashes the template compiler in run-time scope today, with an
```
gfraiteur: This story cannot be implemented because we have no way to detect an experimental featurel Why? Because we cannot have references to an experimental API in our source code, because by definition experimental APIs can be removed, breaking our assemblies. So what we do is to forbid use of any experimental lang feature.

Metalama.Framework/docs/2027.0/user-stories/S-01-langversion-clamp-and-sdk-ceiling.md
```diff
@@ -0,0 +1,67 @@
+### S-01. Build: LangVersion clamp and .NET SDK ceiling comparison
```
gfraiteur: Bugs must be titled so the title describes the problem,  not the corrective action

Metalama.Framework/docs/2027.0/user-stories/S-02-language-version-mapping-and-manifest.md
```diff
@@ -0,0 +1,51 @@
+### S-02. Language version: display mapping and compile-time manifest fallbacks
```
gfraiteur: Same problem with bug title.

Metalama.Framework/docs/2027.0/user-stories/S-03-remove-previous-baseline-residue.md
```diff
@@ -0,0 +1,60 @@
+### S-03. Build: residue of the previous platform baseline
```
gfraiteur: Should say here "Remove residues"

Metalama.Framework/docs/2027.0/user-stories/S-04-refactoring-provider-entry-points.md
```diff
@@ -0,0 +1,86 @@
+### S-04. Design time: code refactoring provider entry points
```
gfraiteur: Same problem with bug title. Does not say the problem.

Metalama.Framework/docs/2027.0/user-stories/S-20-experimental-syntax-guard-in-templates.md
gfraiteur: Withdrawn, and you are right on both halves of the reasoning.

It cannot be implemented for the reason you give: detecting an experimental node means naming its kind, and this source may not reference an experimental Roslyn member, which can be removed and would break the assemblies that carry it.

It is also unnecessary, which the story should have noticed. An experimental feature is gated on `LanguageVersion.Preview`, and `CompileTimeAspectPipeline.VerifyLanguageVersion` already reports `PreviewCSharpVersionNotSupported` unless the project sets `AllowPreviewLanguageFeatures`. Forbidding the preview version forbids every experimental feature at once, without naming any of them, which is exactly the answer you describe. The unsafe expression is covered by that rule and needs nothing of its own.

What survives is the part that had nothing to do with experimental syntax: a collection expression with a with-element crashes the template compiler today, and the regeneration removes the crash, so the story becomes the test coverage that proves it. It is retitled `C# 15 collection expression arguments: support in templates`, resized to small, and the withdrawn proposal is recorded in its context and its not-in-scope section so that it is not proposed again.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-01-langversion-clamp-and-sdk-ceiling.md
gfraiteur: Fixed here and in the other three defect stories. The title now names the failure:

`Build: a net11.0 project is silently compiled as C# 12, and every .NET 11 software development kit is reported as unsupported`

Both defects of the story are in it, because they are separate faults in the same file and the second is live today.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-02-language-version-mapping-and-manifest.md
gfraiteur: Retitled to `Language version: the unsupported-version diagnostics crash instead of naming the version`, which is the observable failure: the display mapping has no arm for the numeric value of C# 15, so it throws while formatting `LAMA0051` or `LAMA0052` and the user sees `LAMA0001` with a request to open a support ticket.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-03-remove-previous-baseline-residue.md
gfraiteur: Done: `Build: remove the residues of the .NET 8 and .NET 9 removal`.

I also changed its issue type from bug to enhancement, because an imperative title describes a task rather than a defect, and the two rules would otherwise contradict each other on the same story. Tell me if you would rather keep it typed as a bug; the four items it covers are genuinely faults, two of them tests that silently never run.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-04-refactoring-provider-entry-points.md
gfraiteur: Retitled to `Design time: renaming a code refactoring provider disables it in the editor and no test detects it`, which is the failure: the three implementation types are named as string literals and are absent from the constant list that the test compares against the real type names, so the suite stays green while the editor loses the refactorings.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-02-variant-gating-decision.md
gfraiteur: Merged into the renaming story, which is now `S-13. Build: rename the latest Roslyn variant to 5.12 and regenerate the syntax model`. Its opening paragraph says plainly that the latest variant is renamed from 5.10 to 5.12 in the source, and that the C# 15 Roslyn members are reached through conditional compilation in that variant.

Your second comment is the reason for merging rather than rewriting: the two could not merge separately without breaking the build, so they were never really two stories.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-03-type-declaration-by-type-test.md
gfraiteur: Agreed, and the story is deleted.

The remedy it proposed contradicted the routing pattern, and it was also unnecessary: a union is a new declaration kind, so it is admitted by adding the arm to the switches that already route by kind. Those switches are the subject of S-08, which is a defect story of its own because the same four places already fall through for records, record structs and extension blocks. Nothing about unions needs a type comparison.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-18-2-union-rewriters-and-visitors.md
gfraiteur: Yes, that is what it means, and the title now says so: `C# 15 unions: rewriters and visitors`. It follows the 2026.0 convention of naming the feature, then the facet.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-18-3-union-linker-and-advising.md
gfraiteur: All three points are applied.

The title is now `C# 15 unions: linker and advising`, which is what the story means.

The meta-story exists: `S-18. C# 15 unions: support in target code`, with the six sub-stories `S-18-1` to `S-18-6` in the numbering scheme you propose. `S-18-1` is the code model, then the rewriters and visitors, the linker and advising, the design-time partial part, the comparers, and the pattern and extension libraries. They are recorded as one stack of pull requests merged together, so the code model change and the users of it never sit in the repository apart.

I kept the number 18 rather than 13, because 13 is now the Roslyn variant renaming that S-18 is blocked by. The scheme is the one you describe.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-29-introduce-union-and-case-attribute-form.md
gfraiteur: Moved. The three discretionary stories are now last and in one block: S-28, the introduction of a closed class, which is the smallest and therefore the likeliest to survive a cut, then S-29, then S-30. The ordering section of the index calls that block the cut line of the release.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-16-closed-modifier-in-code-model.md
gfraiteur: Retitled to `Code model: support for the closed modifier`.

I read the 2026.0 titles and applied the convention to the whole set. It is `<facet>: <what>` for a story that touches one subsystem, and `C# 14 <feature>: <facet>` for the feature stories, which is why the union and closed stories now read `C# 15 unions: code model`, `C# 15 unions: linker and advising` and so on.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-19-reject-labeled-break-in-templates.md
gfraiteur: Checked in the source, and only the jump is rejected today, not the label.

`TemplateAnnotator` overrides `VisitGotoStatement` and calls `ReportUnsupportedLanguageFeature` for the `goto` keyword, at `TemplateAnnotator.cs:2601-2605`. It has no `VisitLabeledStatement` override, and there is no other rejection of a labeled statement in `Metalama.Framework.Engine/Templating`: the only three occurrences of `LabeledStatement` there are generation and handling in `TemplateCompilerRewriter` and `TemplateExpansionContext`, not a diagnostic.

So a label in a template is accepted today and simply unreachable, because the only statement that could jump to it is refused. C# 15 makes it reachable, which is why the story exists. Following your earlier comment, the story now rejects the label itself as well as the labeled `break` and `continue`, for consistency and because the annotator cannot classify a label as compile-time or run-time.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-28-introduce-closed-class.md
gfraiteur: Removed. The bullet that reported a diagnostic when the target compilation provides neither `IsClosedTypeAttribute` nor `CompilerFeatureRequiredAttribute` is gone from the story, and no other story proposes a polyfill check. A missing polyfill fails the compilation, as elsewhere.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-01-langversion-clamp-and-sdk-ceiling.md
```diff
@@ -0,0 +1,67 @@
...
+set `LangVersion` that is not one of `12.0`, `13.0`, `14.0`, `default`, `latest`, `latestMajor` or `preview` down to
+`12.0`, and the warning at `:243-248` carries no `Code` attribute and explains the rewrite falsely. Separately,
+`_MetalamaSdkVersion` at `:399` strips only the prerelease label and is then compared with `VersionGreaterThan`
+against `MaximumSdkVersion` `11.0` at `:412`, so every .NET 11 SDK, including `11.0.100`, reports `LAMA0601`
+although `Metalama.Framework.props:33` declares .NET 11 supported and `:39` names 10.0 and 11.0 as the supported
+versions. The second defect is live today.
+
+#### Context
+
+Section 6b of [`DECISIONS.md`](../DECISIONS.md) removes the build container work from the release and names these two
```
gfraiteur: I reverted this decision, we will add .NET 11 SDK.

Metalama.Framework/docs/2027.0/DECISIONS.md
```diff
@@ -0,0 +1,769 @@
...
+earlier proposal to support labeled break and continue in templates and to rename labels when inlining. The
+diagnostic belongs to the template compiler, beside the other rejections of syntax that the annotator cannot
+classify.
+
+Run-time code that an aspect transforms, and that uses a labeled break or continue outside a template, is not
+affected by this decision and must keep working once the syntax model is regenerated from the stable grammar.
+
+## 5. Introducing a closed class or a union: to be analyzed
+
+Whether an aspect may introduce a closed class, and whether it may introduce a union, is not decided. It needs a
```
gfraiteur: Analyzed. Nice to have for 2027.0, will overflow to 2027.1 if lack of time.

Metalama.Framework/docs/2027.0/DECISIONS.md
```diff
@@ -0,0 +1,769 @@
...
+## 6. No `net11.0` leg is added to the full test matrix without a reason
+
+Adding `net11.0` beside `net10.0` in every test project is not justified unless there is a .NET 11 application
+programming interface that Metalama wants to use. The question of whether such an interface exists must be answered
+before any test matrix story is written. What remains necessary regardless is that the .NET 11 SDK works as a build
+host, because it is in the supported set: the language version clamp and the supported-toolchain check must be
+correct under it, and a scenario that exercises the .NET 11 SDK as a build host is worth having even when the
+target framework of the test project stays `net10.0`.
+
+## 6b. The .NET 11 SDK in the build container is not a priority
```
gfraiteur: We will add .net 11 sdk to the build container and update global.json (make it our main sdk) otherwise we can't test C# 15 features (we need the sdk compiler)

Metalama.Framework/docs/2027.0/DECISIONS.md
```diff
@@ -0,0 +1,769 @@
...
+is `11.0` and the comparison uses `VersionGreaterThan` against the full version, so an .NET 11 SDK of `11.0.100`
+compares as greater than the ceiling and every build with it reports that the SDK is unsupported. This is a defect
+of the version comparison and it is verified by a test of that comparison, not by installing an SDK in the image.
+
+The language version clamp in `Metalama.Framework.targets` must not rewrite the language version that a `net11.0`
+project implies. That defect is likewise a property of the condition and is verified without the SDK.
+
+## 6c. The answer to decision 6: no .NET 11 interface is wanted
+
+Established on 2026-09-04 by the analysis in
```
gfraiteur: Indeed no .NET 11 TFM is needed.

Metalama.Framework/docs/2027.0/DECISIONS.md
```diff
@@ -0,0 +1,769 @@
...
+
+Issues #865, #866, #867 and #869, which ask for the introduction of enums, delegates, records and structs, are all
+open user stories with no milestone, no assignee and no body, imported from the previous tracker. Union
+introduction belongs with them rather than with the C# 15 work.
+
+The one thing that would change this recommendation is a customer scenario that needs an aspect-generated closed
+hierarchy. None is known.
+
+## 2b. A consequence of decision 2 that needs a further decision
+
```
gfraiteur, line 164: Consolidate this document, don't present it as a history of decisions.

Metalama.Framework/docs/2027.0/DECISIONS.md
```diff
@@ -0,0 +1,769 @@
...
+when the story is implemented, not when it is written.
+
+The analyses under `analysis-reports` may go further, because they exist to establish feasibility and size, and a
+proposed shape is often the shortest way to show that something is possible. A proposed shape in an analysis is
+evidence for the estimate. It is not a specification, and a story must not quote it as one.
+
+This applies in particular to the union work of section 5c, where the natural temptation is to fix the signature of
+an introduction method in the story text.
+
+### 7b. Refinement: the interfaces are drafted, as material to think with
```
gfraiteur: Consolidate.

Metalama.Framework/docs/2027.0/DECISIONS.md
gfraiteur: Done. `DECISIONS.md` is rewritten as the position that holds today rather than as the sequence that produced it.

Fourteen numbered sections with revision sub-sections (2b, 4b, 5b to 5f, 6b to 6d, 7b, 8b, 8c and 14.1 to 14.4) become twelve sections, each stating one decision in its final form. Nothing that reversed an earlier decision announces that it did: the introduction of unions, cases and closed classes is one section that says they are in scope and discretionary, the variant set is one section that states the set and the criterion, and the answers to the review are folded into the sections they concern rather than collected at the end. The document loses 283 lines and no decision.

Three consequences were carried through the rest of the directory. The cross-references were renumbered in every citing document, 104 of them. The citations of the questions that are now answered were replaced by the fact they cited, so no document points at an entry that no longer exists. `OPEN-QUESTIONS.md` is reduced to the seven questions that remain open, with their identifiers kept stable so that a gap in the numbering means answered rather than missing.

One correction fell out of the pass, from the file rather than from the review. The document said the `MSBuildExtensionsPath` mitigation in PostSharp.Engineering was incomplete. It is complete upstream, in 2023.2.421, and `Directory.Packages.props:12` pins 2023.2.420, so S-09 takes the dependency rather than writing the fix.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/DECISIONS.md
gfraiteur: Recorded, including the part that was still open. Section 9 of `DECISIONS.md` now states that the .NET 11 software development kit is installed in the build container and named by `global.json` as the main kit of the product, and gives the mechanism: the language version provider caps the compile-time compilation by the major version of the installed kit, so a C# 15 construct in compile-time code cannot compile without it.

Story S-09 no longer asks which kit `global.json` pins. It moves the `dotNetSdkVersion` pin to the .NET 11 kit and records beside the constant that the product is therefore built with a kit that no supported Visual Studio installs, which the desktop MSBuild path of the container still exercises.

One thing in that story was wrong and is corrected. It said the `MSBuildExtensionsPath` mitigation was missing. Pull request #1919 removes that variable in `DotNetTool.cs:61` and in `MSBuildTool.cs:55`, and the fix is in PostSharp.Engineering 2023.2.421, while `Directory.Packages.props:12` still pins 2023.2.420. The scope bullet is now to move that pin.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/user-stories/S-01-langversion-clamp-and-sdk-ceiling.md
gfraiteur: Applied. S-01 no longer says that the container work is out of the release. It now says that both of its defects are properties of a version comparison and are verified without an installed kit, so the story is independent of the container work rather than a replacement for it, and its not-in-scope section points at S-09.

S-09 is the story that installs the kit, and section 9 of `DECISIONS.md` records the decision.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/DECISIONS.md
gfraiteur: Kept, and now stated in one place. Section 9 of `DECISIONS.md` holds both halves together: the software development kit is wanted as a toolchain, and the `net11.0` target framework is not wanted for the product or for the tests, with the evidence that no .NET 11 interface justifies one.

The analysis report that reached the opposite conclusion about the container is corrected on that point. Its conclusion about the target framework stands and is the one section 9 cites.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/DECISIONS.md
gfraiteur: Recorded. The introduction of a closed class, of a union, and of a case into a union are one section of `DECISIONS.md` now, section 4, which states that all three are in scope for 2027.0 and discretionary under section 2, in that order: the closed class is the smallest and therefore the likeliest to survive a cut, and the two union stories go together because the second is meaningless without the first.

The stories carry the same priority field, and the ordering section of the story index calls that block the cut line of the release.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/DECISIONS.md
gfraiteur: Consolidated. The two sections you marked here were the interface design rule and its refinement, which are now one section, section 11: a story states the capability, the scope and the acceptance criteria and not the shape of an interface, and the interfaces are nevertheless drafted as illustrative material that carries no authority. The findings of the drafts are in the same section instead of a separate one.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
```diff
@@ -0,0 +1,118 @@
...
+  does not repair.
+- Option B, ship the attribute form only. Nothing diverges, and an aspect cannot add a case to a union that a user
+  wrote with the concise syntax.
+
+The analysis recommends Option A, and states that if only one form fits the release it should be the attribute
+form. Settled by the product owner.
+
+### Q2. Does the lower Roslyn variant report the divergence, or stay silent?
+
+Blocking, for one story, and it applies to every C# 15 reader.
```
gfraiteur: I don't understand the relevance of this question. We are loaded by the hosts, so if the host does not support Roslyn 5.12 (and since we refuse preview lang features), we have no case in supporting these new features, and returning false silently is okay. Anyway the host will report squiglles for invalid syntax or forbidden preview features

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
```diff
@@ -0,0 +1,118 @@
...
+- Option C, remove the members on the lower variant rather than answering false. Both options above take as given
+  that the public members exist in every host and answer false where the engine cannot answer, which follows from
+  the public assembly not being built per Roslyn version. Whether that assembly could carry a variant-specific
+  part has not been examined.
+
+The draft recommends Option B. Settled by the product owner.
+
+### Q3. Is the severity and the opt-out of each new design-time diagnostic right?
+
+Not blocking. Two diagnostics are drafted: the one of Q2, and the one that reports that the editor cannot show a
```
gfraiteur: Probably not a smart question, or it should be expressed in a concrete way.

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
gfraiteur: You are right, and the question is removed. Section 6 of `DECISIONS.md` now decides it: the lower variant answers false and reports nothing.

I checked the argument in the source before applying it, because it turns on the preview refusal. `CompileTimeAspectPipeline.VerifyLanguageVersion` reports `PreviewCSharpVersionNotSupported` when the effective language version is `Preview`, unless the project sets `MetalamaAllowPreviewLanguageFeatures`, which `DefaultProjectOptions` leaves false. A host whose Roslyn predates C# 15 has no `LanguageVersion.CSharp15` at all and reaches every C# 15 feature only under `Preview`. So there is no configuration in which valid C# 15 code sits in such an editor and Metalama is the only thing that could report the divergence: the host has already flagged the syntax it cannot parse or the feature it does not offer.

That is what the analysis missed. It reasoned from the divergence outward, treating the false answer as the failure, and never asked whether the code that produces the divergence can exist in that host. The draft in `analysis-reports/12-csharp15-api-drafts.md` recommended reporting on that basis, and section 11 now records why the recommendation is not followed.

Six stories cited the deferred decision and now state the outcome instead. S-13, S-16 and S-18-1 pin the false value and the absence of a diagnostic with a test, rather than deciding a behaviour. S-05, S-26, S-27 and S-28 drop the sentences that waited on it.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Metalama.Framework/docs/2027.0/OPEN-QUESTIONS.md
gfraiteur: Removed. It was not a smart question, and your other comment took away most of its subject.

It asked about the severity and the opt-out of two design-time diagnostics. One of them was the divergence report, which is now decided not to exist. The other is the diagnostic that says the editor cannot show a case added to a `union` declaration, which exists only if Q1 chooses to ship that form. A question about the severity of one hypothetical diagnostic is not a product decision, so it is now one sentence inside Q1: under Option A the severity and the opt-out are chosen when the story is written.

`OPEN-QUESTIONS.md` is down to five entries: Q1, the two calendar measurements, the closed type read from a referenced assembly, and the three `platform-support.md` corrections that await your approval.

— Claude for @gfraiteur

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #1974, opened 2026-09-06
Re-derive the design-time binding ceilings against the Visual Studio 2026 long-term servicing channel baseline


gfraiteur:
Every assembly of the `net472` design-time payload sits exactly at the binding-redirect ceiling that governs it, and none has headroom. The ceilings were measured against Visual Studio Professional 2026 18.9, which is a Stable-channel release and not the floor of PB-2027.0. The floor is the Visual Studio 2026 long-term servicing channel baseline, which had not shipped when the measurement was taken. This issue tracks taking the measurement again against that baseline, and lowering the pins if it turns out to sit lower.

## The current position

Measured on the payload built from `Metalama.Framework.CompilerExtensions.Resources`:

| Assembly | Assembly version in the payload | Redirect upper bound | Headroom |
| --- | --- | --- | --- |
| `System.Text.Json` | 10.0.0.8 | 10.0.0.8 | none |
| `System.IO.Pipelines` | 10.0.0.8 | 10.0.0.8 | none |
| `Microsoft.Bcl.AsyncInterfaces` | 10.0.0.8 | 10.0.0.8 | none |
| `System.Diagnostics.DiagnosticSource` | 10.0.0.8 | 10.0.0.8 | none |
| `System.Threading.AccessControl` | 10.0.0.8 | 10.0.0.8 | none |
| `System.Threading.Tasks.Extensions` | 4.2.4.0 | 4.2.4.0 | none |
| `System.Memory` | 4.0.5.0 | 4.0.5.0 | none |
| `System.Buffers` | 4.0.5.0 | 4.0.5.0 | none |
| `System.Numerics.Vectors` | 4.1.6.0 | 4.1.6.0 | none |
| `System.Runtime.CompilerServices.Unsafe` | 6.0.3.0 | 6.0.3.0 | none |

## Why this needs a measurement rather than a judgement

A reference above the `oldVersion` upper bound of a redirect is not redirected, and the common language runtime then rejects the load because the manifest of the located assembly does not match the reference. The whole payload is one revision below that outcome on every one of the ten assemblies, so a baseline whose redirects sit lower moves several of them outside their range at once rather than one.

The failure is also the silent kind described in `Metalama.Framework/docs/platform-support.md`: `ServiceHub.RoslynCodeAnalysisService` logs the load failure and the integrated development environment shows no diagnostics, no code lens and no generated code. It cannot be found by reading the editor.

## Scope

[...]

Issue #1876, opened 2026-08-31
Remove explicit support for .NET 8 and .NET 9


gfraiteur:
## Summary

Metalama stops declaring .NET 9 as a supported target. Every `net9.0` target framework is replaced by `net10.0`, not simply removed. The supported .NET targets become .NET Framework 4.7.2 / 4.8, .NET 8 and .NET 10.

.NET 9 is a Standard Term Support release, and Microsoft ended its support in May 2026. .NET 10 is the Long Term Support release that succeeds it, so it is the target that replaces it.

Projects that target `net9.0` keep building. NuGet resolves the `net8.0` asset for them, which is the same source compiled for an earlier target. No public API changes, so this is not a breaking change under the usual definition, and the `breaking` label does not apply.

## Out of scope: .NET 11

Adding a `net11.0` target framework is not part of this story. .NET 11 is still in preview. A project that targets `net11.0` resolves the `net10.0` asset, which is what this story delivers.

## Scope in this repository

Product projects that declare `net9.0`, each of which moves to `net10.0`:

- `Metalama.Framework/src/Metalama.Framework.Workspaces` — `net8.0;net9.0`
- `Metalama.Framework/src/Metalama.Extensions.DiffEngine` — `net472;net8.0;net9.0`, plus the `net9.0` entries in `MetalamaExtensionAssemblies.props`, `build/Metalama.Extensions.DiffEngine.props` and `buildTransitive/Metalama.Extensions.DiffEngine.props`
- `Metalama.Framework/src/Metalama.Extensions.HtmlWriter` — the same shape

Build configuration and documentation:

- `Directory.Packages.props` — the `MicrosoftBuildVersion` version conditioned on `'$(TargetFramework)'=='net9.0'`, and the comment above it. The condition moves to `net10.0`, and the pinned version has to be reviewed against the lowest Visual Studio version supported on that target.
- `Directory.Packages.md` — the runtime target framework table and the conditional version example
- `Metalama.Framework/docs/extensibility.md` — it instructs extension authors to target `net472`, `net8.0` and `net9.0`. That instruction becomes `net472`, `net8.0` and `net10.0`.

[...]

PR #1975, opened 2026-09-06
Move the user-surfacing package pins to the .NET 10.0 line (#1903)

### LINKED ISSUE #1903: Re-derive the .NET 8.0 line pin of user-surfacing packages against PB-2027.0
`Directory.Packages.md` classifies every dependency by audience and pins the user-surfacing ones to the .NET 8.0 line (`System.*` 8.0.x, `Microsoft.Extensions.*` 8.0.x). That pin was derived when `net8.0` was a target framework of our own projects and a supported target framework for user applications. #1876 removed both, so the rationale has to be re-derived.

## Why the current rationale no longer holds

The pin protects a user application that targets `net8.0` and resolves the `netstandard2.0` asset of a user-surfacing package. Raising a user-surfacing dependency to a higher .NET major forces a transitive upgrade on that application.

Under PB-2027.0, recorded in `Metalama.Framework/docs/platform-support.md`, that configuration is outside the tested set. #1886 makes it report a warning. A constraint we accept in order to protect a configuration we do not test is a constraint we pay for twice.

## What has to be decided

For each user-surfacing pin on the 8.0 line, decide whether the .NET 10.0 line can replace it. The answer is not uniform, and two limits bind independently of the target framework of the user application:

- A package consumed inside a design-time host is capped by what the host ships, because the host-provided assembly wins on load. That is the bucket model of `Directory.Packages.md`, and it is unaffected by this issue.
- A package that is also restored for `netstandard2.0` or `net472` must still publish an asset for those target frameworks at the higher version.

## The pins in scope

The user-surfacing entries of `Directory.Packages.props` currently on the 8.0 line:

`Microsoft.Extensions.DependencyInjection`, `Microsoft.Extensions.DependencyInjection.Abstractions`, `Microsoft.Extensions.Caching.Memory`, `Microsoft.Extensions.Hosting.Abstractions`, `Microsoft.Extensions.Logging`, `Microsoft.Extensions.Logging.Abstractions`, `Microsoft.Extensions.Logging.Console`, `System.Drawing.Common`, `System.DirectoryServices`, `System.IO.Packaging`, `System.Threading.AccessControl`, `System.IO.Pipelines`, `Microsoft.Bcl.AsyncInterfaces`.

`System.Collections.Immutable` and `System.Reflection.Metadata` are governed by the Roslyn version rather than by this rule, and `System.Threading.AccessControl` is additionally referenced by an embedded-resource path in `Metalama.Framework.CompilerExtensions` that names the asset folder of the package. Both need a version whose package layout the path can address.

## Scope

1. Rewrite the audience paragraph of `Directory.Packages.md` so that it derives the line from PB-2027.0 rather than from .NET 8 support.
2. Raise each pin above to the 10.0 line where the two limits allow it, and record the reason where they do not.
3. Verify that the `netstandard2.0` and `net472` assets of our packages still restore.

This issue is separated from #1876, which removes the target frameworks and does not change a single version pin.

— Claude for @gfraiteur

gfraiteur:
Closes #1903. Companion pull requests: metalama/Metalama.Premium#96 and metalama/Metalama.Vsx#69.

`Directory.Packages.md` pinned every user-surfacing dependency to the .NET 8.0 line, to protect a user application targeting `net8.0` from a forced transitive upgrade. #1876 removed `net8.0` both as a target framework of our own projects and as a supported target framework for user applications, so the configuration that pin protected left the tested set of PB-2027.0 and the pin lost its justification. This re-derives it.

## The version of the line is 10.0.8, and it is a measurement

Every package that reaches the `net472` design-time payload is capped by the binding redirect that governs it in `devenv.exe.config`. Measured on Visual Studio Professional 2026 18.9, those upper bounds are assembly version `10.0.0.8`, and `4.2.4.0` for `System.Threading.Tasks.Extensions`. The `lib/net462` assets of NuGet 10.0.8 and 4.6.3 carry exactly those, so they are the highest versions the payload may take. The `lib/net462` asset of 10.0.11 carries `10.0.0.11`, which is above the bound.

After the change every assembly of the payload sits exactly at its ceiling, and none above it:

| Assembly in the net472 payload | Before | After | Ceiling |
| --- | --- | --- | --- |
| `System.Text.Json` | 10.0.0.11, above the bound | 10.0.0.8 | 10.0.0.8 |
| `System.IO.Pipelines` | 8.0.0.0 | 10.0.0.8 | 10.0.0.8 |
| `Microsoft.Bcl.AsyncInterfaces` | 8.0.0.0 | 10.0.0.8 | 10.0.0.8 |
| `System.Diagnostics.DiagnosticSource` | 6.0.1.0 | 10.0.0.8 | 10.0.0.8 |
| `System.Threading.AccessControl` | 8.0.0.0 | 10.0.0.8 | 10.0.0.8 |
| `System.Threading.Tasks.Extensions` | 4.2.0.0 | 4.2.4.0 | 4.2.4.0 |
| `System.Memory`, `System.Buffers`, `System.Numerics.Vectors`, `System.Runtime.CompilerServices.Unsafe` | unchanged | 4.0.5.0, 4.0.5.0, 4.1.6.0, 6.0.3.0 | identical |

The `System.Text.Json` row is a defect this pull request fixes rather than introduces. The value came from `Roslyn.5.10.0.props` and predates this work.

## A constraint the shipped artifact contradicts

`System.IO.Pipelines`, `System.Diagnostics.DiagnosticSource`, `Microsoft.Bcl.AsyncInterfaces` and `System.Threading.Tasks.Extensions` were held on their old lines by a rule stating that a deployed Metalama.Vsx re-resolves them through its own central package management, so that raising them could break an already-installed extension.

Inspecting an installed Metalama.Vsx 2026.1.5 shows otherwise. It merges `Metalama.Backstage`, `Metalama.Backstage.Tools` and `Metalama.Framework.DesignTime.Rpc` into a single `Metalama.Repacked` assembly carrying its own identity, so it holds its own copy of that code and our version choices do not reach it. It contains 65 files, and none of those four packages is among them; the only loose framework assemblies it ships are `Microsoft.Bcl.HashCode`, `System.Configuration.ConfigurationManager`, `System.IO.Hashing`, `System.Reflection.MetadataLoadContext`, `System.Resources.Extensions`, `System.Text.Encodings.Web` and `System.Text.Json`. Visual Studio provides the rest, under the redirects above.

The coupling that does exist is at the build of that repository, which enables central transitive pinning and therefore fails with NU1605 if its pins sit below what our packages require. Its pins are raised in the companion pull request, and the two repositories have to merge together.

## System.Text.Json is governed by the host, not by the redirect

The core flavour of the payload deliberately does not embed `System.Text.Json`, because the parent assembly load context, which is Roslyn, provides it. Assembly binding rolls forward and never back, so the floor is the lowest version any host in the supported set provides. `Microsoft.CodeAnalysis` 5.0 requires the 9.0 line and the Rider backend presents it, so that is the floor.

`SystemTextJsonMinVersion` therefore moves from 8.0.6 to 9.0.10 rather than to 10.0.8, and the Roslyn 5.0 variant moves from 9.0.0 to 9.0.10. The fallback follows the lowest variant, because the projects that are built once and linked into every variant read it, and a higher value makes the Roslyn 5.0 variant a package downgrade. This also removes the vulnerable 8.0.6 from the dependency graph that a user application acquires through `Metalama.Backstage`.

## What else the change carries

- `Microsoft.Bcl.TimeProvider` moves with `Microsoft.Bcl.AsyncInterfaces`, which its `net462` and `netstandard2.0` groups require on the same line, and `Microsoft.Extensions.TimeProvider.Testing` follows it.
- The override of `System.Text.Json` in `Metalama.Patterns/Directory.Packages.props` is removed. It existed to keep `System.Threading.Tasks.Extensions` 4.6.3 out of the user-surfacing packages, which is now the central version.
- `platform-support.md` carried a stale example: a `CoreAssemblyToEmbed` naming `runtimes/win/lib/net8.0/System.Threading.AccessControl.dll`, which #1876 had already removed. The point it illustrated is kept and the example corrected.
- The audience section of `Directory.Packages.md` is rewritten, and the gotcha about separately-deployed downstream consumers is restated against the evidence above.

## Verification

- `Build.ps1 build`: all six solutions built and packed, zero errors.
- `dotnet restore` clean on every top-level solution.
- Continuous integration mode build, zero warnings, on every project touched: `Metalama.Backstage`, `Metalama.Framework.DesignTime.Rpc`, `Flashtrace`, `Metalama.Patterns.Caching`, `Metalama.Patterns.Caching.Backend`, `Metalama.Tool`, `Metalama.Backstage.Desktop.Windows`.
- Tests on both target frameworks: `Metalama.Framework.Tests.UnitTests` 2778 and 2720, `Metalama.Backstage.Tests` 1653 and 1653, `Metalama.Patterns.Caching.UnitTests` 483 and 469, `Flashtrace.UnitTests` 13 and 13.
- The produced packages were read back to confirm the versions reach the `netstandard2.0` and `net472` dependency groups.
- Interactive design-time verification in Visual Studio 2026 on `Metalama.Framework.TestApp` and on the Premium code fixes project.

## What reviewers should weigh

There is now no headroom anywhere in the payload: every assembly is exactly at its bound. The bounds were measured against Visual Studio 2026 18.9, which is a Stable-channel release and not the floor of PB-2027.0. Re-measuring them against the Visual Studio 2026 long-term servicing channel baseline, once it ships, is item 1 of the verification checklist of `platform-support.md` and is tracked by #1974.

— Claude for @gfraiteur


gfraiteur:
Both review findings were valid. One is fixed in 3e6d40f; the other no longer applies after the comment rewording in 5f7f56b.

## The out-of-band section contradicted the change

Correct, and this was the more serious of the two: the section stated that the central pin of `System.Threading.Tasks.Extensions` is held at 4.5.4 for the installed Metalama.Vsx versions, while this change pins it at 4.6.3. A reader had no way to tell which statement governed. The paragraph now records that every member of the family sits at its ceiling, 4.6.3 included, and points at the section that explains why being a flowed dependency does not lower that ceiling.

The measurement table above it is unchanged and stays correct: it lists the assembly versions of several versions of each package, 4.5.4 and 4.6.3 among them, and is a record of measurements rather than a statement of what is pinned.

## System.Diagnostics.DiagnosticSource was missing from the list

Also correct at the time. That passage has since been rewritten for a separate reason, and the replacement names all four packages: `System.IO.Pipelines`, `System.Diagnostics.DiagnosticSource`, `Microsoft.Bcl.AsyncInterfaces` and `System.Threading.Tasks.Extensions`. No further change was needed.

## Note on the rest of the review

The prose of this change was rewritten in 5f7f56b, before this review was read, for a reason worth stating: the comments described what had changed and which issue changed it, which is history that the commit already records and that misleads once it ages. They now describe the situation that holds and the reason for it. Two of the review's file summaries therefore no longer match the diff.

The point about host compatibility requiring human validation stands, and is why the pull request description carries the ceiling table and the note that there is now no headroom at any position. The interactive design-time verification was run in Visual Studio 2026 on `Metalama.Framework.TestApp` and on the Premium code fixes project.

— Claude for @gfraiteur

PR #1993, opened 2026-09-08
Design proposal: type facets in the code model

gfraiteur:
Design proposal only. No production code is changed, and nothing described in the document is implemented.

The document proposes a uniform way to expose the structure that is specific to a kind of declared named type: a
delegate signature, an enum underlying type and members, the synthesized members of a record, and the cases and
`Value` property of a C# 15 union. The code model exposes that structure in four different ways today. Two are
defective: a member reached by a string literal, and one member with several meanings. The third is a derived
interface reached by a type test, which is correct for a type that a type expression forms and does not extend to a
declared type. The fourth is flat members on `INamedType`, which is what this proposal replaces.

### The proposal

- `INamedType.Facets`, of type `ITypeFacetCollection : IReadOnlyCollection<ITypeFacet>`, with one nullable read-only
  property per facet, and a `TypeFacetKind` enum. `INamedType` gains one member for any number of facets.
- `IDelegateFacet`, `IEnumFacet`, `IRecordFacet`, `IUnionFacet` and `IUnionCase`. Each facet names the members that
  the compiler synthesizes, so that no consumer reaches one by its identifier.
- Facets expose invokers through the type of their members, because `IMethod`, `IConstructor` and `IFieldOrProperty`
  already derive from the invoker interfaces. A facet declares a method that creates an expression only where no
  member carries the operation, following `ITupleType.CreateCreateInstanceExpression`.
- Writing stays on builder interfaces derived from `INamedTypeBuilder`, following `IExtensionBlockBuilder`, and not
  on the facets.

No member is made obsolete, and the proposal carries no user-facing breaking change.

### Three decisions worth reviewing first

Section 4.4 recommends holding `IsUnion`, `IsUnionDeclaration` and `UnionCaseTypes` back from #1991 rather than
shipping them and duplicating them in `Facets.Union` later. It also records why two of the three would be wrong as
specified there.

Section 4.5 excludes tuples. A tuple is not declared: the type is formed by a type expression at the place where it
is used, which is why `TupleType` sits in `CodeModel/Source/ConstructedTypes/` beside the array and the pointer.
`ITupleType` belongs to the family of `IArrayType` and `IPointerType`, where the type test is the correct
mechanism. The section records the alternative that an earlier revision proposed, which was an `ITupleFacet` with
`ITupleType` obsoleted, and the price it carried.

Section 4.2 establishes that a type can have two facets at once. `record union` is not supported, but the attribute
form of a union accepts a record, and `[Union] public record Wrapper( int Value );` is a record and a union with one
case of type `int`. The collection therefore does not promise that `Count` is at most one, and there is no ordering
of facets by priority.

Section 7 lists five open questions.

### Next steps

The document is written to be reviewed and then split. Section 6 proposes five implementation pull requests,
starting with the delegate facet, which needs no C# 15, no Roslyn variant and no preview language version, and which
has about fifteen call sites to convert. That conversion is the acceptance test of the shape, before the union facet
is built on it.

— Claude for @gfraiteur

### DIFF
--- a/Metalama.Framework/docs/future/type-facets.md
+++ b/Metalama.Framework/docs/future/type-facets.md
@@ -342,10 +342,11 @@ expression for it.
 
 The member that this calls for is `IMethodBaseInvoker`, declaring the operations that `IMethodInvoker` and
 `IConstructorInvoker` have in common, with `IMethodBase` deriving from it. Creating an instance of a union case
-would then be the invocation of its creation member. That is a story of its own, because it changes `IMethodBase`
-and because the relationship between the three invoker interfaces has to be settled: `IMethod` and `IConstructor`
-already derive from the two specific invokers. It is not part of this proposal, and until it exists a consumer that
-invokes a creation member tests whether it is an `IMethod` or an `IConstructor`.
+would then be the invocation of its creation member. That is a story of its own, issue
+[#1999](https://github.com/metalama/Metalama/issues/1999), because it changes `IMethodBase` and because the
+relationship between the three invoker interfaces has to be settled: `IMethod` and `IConstructor` already derive
+from the two specific invokers. It is not part of this proposal, and until it exists a consumer that invokes a
+creation member tests whether it is an `IMethod` or an `IConstructor`.
 
 ### 2.4. Writing is on builders, not on facets
 
@@ -449,10 +450,12 @@ The flags are the discriminators and the facets are the structure, so they answe
 the same one twice. That is the same relationship that `TypeKind.Delegate` and `TypeKind.Enum` already have with
 their facets.
 
-What does not go on `INamedType` is the structure. Pull request #1991 proposes `IsUnion`, `IsUnionDeclaration` and
-`UnionCaseTypes`. Under this proposal `IsUnion` ships as a flag, and the other two are replaced by `Facets.Union`.
-The recommendation is to hold those two back from #1991 rather than ship them and duplicate them later, because a
-shipped public member cannot be withdrawn.
+What does not go on `INamedType` is the structure. Pull request
+[#1991](https://github.com/metalama/Metalama/pull/1991) proposes `IsUnion`, `IsUnionDeclaration` and
+`UnionCaseTypes`. Under this proposal `IsUnion` ships as a flag, and the other two are replaced by `Facets.Union`,
+which issue [#1941](https://github.com/metalama/Metalama/issues/1941) delivers. The recommendation is to hold those
+two back from #1991 rather than ship them and duplicate them later, because a shipped public member cannot be
+withdrawn.
 
 Both would be wrong as specified there in any case. `UnionCaseTypes` is documented as empty when
 `IsUnionDeclaration` is false, which describes the current implementation from syntax rather than the contract:
@@ -472,7 +475,7 @@ the elements of a tuple are a view over the fields of the underlying `ValueTuple
 that describe the structure of a type that a type expression forms, that are reached by a type test, and that are
 constructed at one site. Section 1.3 states why that mechanism is correct for that family and does not extend to a
 declared type. `ITupleType` and `ITupleElement` move to `Metalama.Framework.Code.Types` so that they sit with that
-family.
+family, which is issue [#1998](https://github.com/metalama/Metalama/issues/1998).
 
 `ITypeFacetCollection` has no tuple property in this proposal, and `INamedType.IsTuple` is the flag. An
 `ITupleFacet` declaring the members of `ITupleType` may be added later for completeness, which would make the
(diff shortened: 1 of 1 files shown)


Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,503 @@
...
+`IsRecord` is a flag on `INamedType`. Adding `IsUnion`, `IsUnionDeclaration` and `UnionCaseTypes` beside it, which
+is what pull request #1991 proposes, continues that pattern. The count grows: the union introduction story adds the
+`Value` property and the case constructors, which would be two more members. Each one is meaningless for the great
+majority of named types.
+
+## 2. The public interfaces
+
+### 2.1. One collection of facets
+
+A facet is the structure that a declared named type has because of its kind, and that other named types do not
```
gfraiteur, line 86: This should go to the namespace Metalama.Framework.Code.Types.
ITupleType and ITupleElement should also be moved to that namespace.

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,503 @@
...
+have. A type has a collection of facets, which is empty for an ordinary class, struct or interface.
+
+```csharp
+// Metalama.Framework/Code/TypeFacetKind.cs
+[CompileTime]
+public enum TypeFacetKind
+{
+    Delegate,
```
gfraiteur, line 97: Add None.

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,503 @@
...
+
+```csharp
+// Metalama.Framework/Code/IRecordFacet.cs
+[CompileTime]
+public interface IRecordFacet : ITypeFacet
+{
+    /// <summary>
+    /// Gets the <c>EqualityContract</c> property, or <c>null</c> when the type is a record struct, which has none.
+    /// </summary>
+    IProperty? EqualityContract { get; }
```
gfraiteur: Rename to EqualityContractProperty

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,503 @@
...
+var returnType = eventType.Facets.Delegate?.InvokeMethod.ReturnType;
+```
+
+The invariant is that the collection contains exactly the typed properties that are not `null`, and that `Count` is
+their number. The invariant is not that `Count` is at most one. Section 4.2 gives the case where two facets
+coexist.
+
+### 2.2. Facets expose the special members as properties
+
+A facet names every member that the compiler synthesizes for that kind of type, so that no consumer has to reach a
```
gfraiteur, line 170: Add implementation note: lazy expressions with [Memo]

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,503 @@
...
+
+```csharp
+// Metalama.Framework/Code/IUnionFacet.cs
+[CompileTime]
+public interface IUnionFacet : ITypeFacet
+{
+    /// <summary>
+    /// Gets the form in which the union is written.
+    /// </summary>
+    UnionForm Form { get; }
```
gfraiteur: Rename to Kind, UnionKind since we already use this term.

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,503 @@
...
+    /// <summary>
+    /// Gets the creation member of the case: the constructor synthesized for a union declaration, the union
+    /// constructor of the attribute form, or the static <c>Create</c> method of the union member provider.
+    /// </summary>
+    IMethodBase CreationMember { get; }
+
+    /// <summary>
+    /// Creates an expression that creates an instance of the union holding a value of this case.
+    /// </summary>
+    IExpression CreateCreateInstanceExpression( IExpression value );
```
gfraiteur: Remove CreateCreateInstanceExpression. What is actually needed is an IMethodBaseInvoker on CreationMember.

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,503 @@
...
+
+### 4.3. No `TypeKind.Union` and no `IUnionType`
+
+The proposal adds no value to `TypeKind` and no interface derived from `INamedType` for unions. Section 1.3 gives
+the two reasons. Roslyn made the same choice: `IsUnion` and `UnionCaseTypes` are declared on `ITypeSymbol`, and
+there is no `IUnionTypeSymbol`.
+
+### 4.4. No flat union members on `INamedType`
+
+Pull request #1991 proposes `INamedType.IsUnion`, `INamedType.IsUnionDeclaration` and
```
gfraiteur: Properties IsUnion, IsDelegate, IsEnum, IsTuple, IsRecord are kept/added in INamedType both for API ease and for performance as the other way around would allocate two objects just to check a bool property.

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,503 @@
...
+constructed at one site. Section 1.3 states why that mechanism is correct for that family and does not extend to a
+declared type.
+
+A facet describes the other thing: the structure that a named type has because of what its declaration says. A
+delegate, an enum, a record and a union are declared. A tuple is not, so `ITupleType`, `ITupleElement`,
+`TypeKind.Tuple` and `TypeFactory.CreateTupleType` are unchanged, and `ITypeFacetCollection` has no tuple property.
+
+The cost of this decision is that a consumer that wants the structure of an arbitrary type has to know two
+mechanisms: a type test for a type that a type expression forms, and the facet collection for a declared one. An
+earlier revision of this document proposed the alternative, which is to declare `ITupleFacet` with the members of
```
gfraiteur: An ITupleFacet might be added later for completeness.

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,503 @@
...
+when the type of an event is not a well-formed delegate, and it returns `false` after the conversion.
+
+Pull request 1 is also the acceptance test of the design. Fourteen of the fifteen call sites assert that the
+`Invoke` method exists, while `Facets.Delegate` is `null` for a malformed type. If several converted sites end up
+asserting that the facet is not `null`, the nullable collection property is the wrong shape for those consumers, and
+the design has to be revised before the union facet is built on it.
+
+## 7. Open questions
+
+1. Whether `IDelegateFacet` exposes `BeginInvoke` and `EndInvoke`. They exist for a delegate compiled for .NET
```
gfraiteur: No BeginInvoke, EndInvoke, it's an obsolete pre-async pattern.

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,503 @@
...
+   Framework and not for one compiled for .NET. Exposing them adds two nullable members for a case that no known
+   consumer needs.
+2. Whether `IRecordFacet` is worth its nullable members. `EqualityContract`, `CloneMethod` and `CopyConstructor` are
```
gfraiteur: Yes, keep nullable members, don't split the interface.

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,503 @@
...
+   all `null` for a record struct. Splitting the interface in two would remove the nullability and add a type.
+3. Whether the enum members deserve an abstraction of their own, in the manner of `ITupleElement`, carrying the
```
gfraiteur: No, enum members are fields, it would be cumbersome to define a new interface that behaves like IField but removes properties.

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,503 @@
...
+   constant value and the index, rather than `IReadOnlyList<IField>`.
+4. Whether the `Delegate`, `Enum` and `Record` property names of `ITypeFacetCollection` pass the naming analyzers of
```
gfraiteur: Not a concern.

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,503 @@
...
+   this repository.
+5. Whether the union facet is populated for a union read from a referenced assembly. The compiled form does not
```
gfraiteur: Need a UnionKind.External to disambiguate

gfraiteur:
All thirteen comments are applied. Three of them had a consequence worth stating before the next pass.

**The `Kind` rename collides with the base member.** `Form` is now `Kind` and `UnionForm` is now `UnionKind`, as
asked. `ITypeFacet.Kind` returns `TypeFacetKind` and `IUnionFacet.Kind` now returns `UnionKind`, so the derived
member has to be declared `new`, and a consumer reads a different type depending on the interface it holds. The
document applies the rename and records the alternative, which is to rename the base member to `FacetKind`, as open
question 1.

**The namespace move is a user-facing break.** `ITupleType` and `ITupleElement` are in `Metalama.Framework.Code`
today, so moving them to `Metalama.Framework.Code.Types` breaks source and binary compatibility. It is now its own
pull request in section 6, carrying the `breaking` label, and open question 3 asks whether it ships as a plain
namespace change or with a compatibility measure. The new facet types are declared in
`Metalama.Framework.Code.Types` from the start, except `ITypeFacetCollection`, which is in
`Metalama.Framework.Code.Collections` beside `IExtensionBlockCollection`.

**`IMethodBaseInvoker` needs one more decision.** `IUnionCase.CreateCreateInstanceExpression` is removed and
`CreationMember` carries the invocation, as asked. `IMethod` and `IConstructor` already derive from
`IMethodInvoker` and `IConstructorInvoker`, so whether those two derive from `IMethodBaseInvoker` or keep their
members independently is open question 2. It is a separate pull request in section 6, because it changes
`IMethodBase`, and the union facet depends on it.

The other ten are applied without qualification: `TypeFacetKind.None`; `EqualityContractProperty`; the lazy
`[Memo]` members, stated in section 2.2 and as implementation guideline 3; `UnionKind.External`; the flags
`IsDelegate`, `IsEnum`, `IsTuple` and `IsUnion` added to `INamedType` beside `IsRecord`, with the allocation
argument recorded in section 4.4, which now says that the flags are the discriminators and the facets are the
structure; the note in section 4.5 that an `ITupleFacet` may be added later for completeness; and the four
resolutions on `BeginInvoke`, the record nullable members, the enum members and the property names, which moved to a
new section 8.

Section 4.4 is the one that changed most: it no longer argues against flat Boolean members, only against flat
structure. The recommendation for #1991 is now to ship `IsUnion` and hold back `IsUnionDeclaration` and
`UnionCaseTypes`.

— Claude for @gfraiteur

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,551 @@
...
+
+```csharp
+// Metalama.Framework/Code/Types/ITypeFacet.cs
+[CompileTime]
+public interface ITypeFacet
+{
+    /// <summary>
+    /// Gets the kind of the facet.
+    /// </summary>
+    TypeFacetKind Kind { get; }
```
gfraiteur: Rename to FacetKind

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,551 @@
...
+    /// <summary>
+    /// Gets the type to which the facet belongs.
+    /// </summary>
+    INamedType Type { get; }
+}
+```
+
+```csharp
+// Metalama.Framework/Code/Collections/ITypeFacetCollection.cs
+public interface ITypeFacetCollection : IReadOnlyCollection<ITypeFacet>
```
gfraiteur, line 128: [InternalImplement] (all interfaces)

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,551 @@
...
+
+```csharp
+// Metalama.Framework/Code/Types/IUnionFacet.cs
+[CompileTime]
+public interface IUnionFacet : ITypeFacet
+{
+    /// <summary>
+    /// Gets the kind of the union.
+    /// </summary>
+    new UnionKind Kind { get; }
```
gfraiteur: rename to UnionKind

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,551 @@
...
+when the type of an event is not a well-formed delegate, and it returns `false` after the conversion.
+
+Pull request 1 is also the acceptance test of the design. Fourteen of the fifteen call sites assert that the
+`Invoke` method exists, while `Facets.Delegate` is `null` for a malformed type. If several converted sites end up
+asserting that the facet is not `null`, the nullable collection property is the wrong shape for those consumers, and
+the design has to be revised before the union facet is built on it.
+
+## 7. Open questions
+
+1. Whether `IUnionFacet.Kind` shadows `ITypeFacet.Kind` or the base member is renamed. `ITypeFacet.Kind` returns
```
gfraiteur: Solved by renaming properties.

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,551 @@
...
+   `TypeFacetKind` and `IUnionFacet.Kind` returns `UnionKind`, so the derived member has to be declared `new`, and a
+   consumer sees a different type depending on the interface through which it reads the property. Renaming the base
+   member to `FacetKind` removes the shadowing at the price of a name that is less uniform with the rest of the code
+   model.
+2. Whether `IMethodInvoker` and `IConstructorInvoker` derive from `IMethodBaseInvoker`, or keep their members
```
gfraiteur: IMethodBaseInvoker a later story.

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,551 @@
...
+   independently. See section 2.3.
+3. Whether the move of `ITupleType` and `ITupleElement` to `Metalama.Framework.Code.Types` ships as a plain
+   namespace change or with a compatibility measure, and in which release.
```
gfraiteur: No compatibility measure, 2027.0.

gfraiteur:
The six comments of the second round are applied, and no question is open any more.

The two kind properties are named apart, so neither shadows the other: `ITypeFacet.FacetKind` returns
`TypeFacetKind`, and `IUnionFacet.UnionKind` returns `UnionKind`. The `new` modifier is gone.

`[InternalImplement]` is on `ITypeFacet`, `ITypeFacetCollection`, `IDelegateFacet`, `IEnumFacet`, `IRecordFacet`,
`IUnionFacet` and `IUnionCase`. `IUnionBuilder` already carried it. Section 2 states the consequence that follows
from the attribute, which is that a facet can gain a member in a later release without a user-facing breaking
change.

`IMethodBaseInvoker` is now described in section 2.3 as a story of its own rather than as part of this proposal. Its
pull request is removed from section 6, the union facet no longer depends on it, and the section says what a
consumer does in the meantime, which is to test whether the creation member is an `IMethod` or an `IConstructor`.

The namespace change of `ITupleType` and `ITupleElement` is recorded in section 3 as shipping in 2027.0 with no
compatibility measure.

Section 7 is now the list of resolved questions and states that none is open. Section 6 lists six implementation
pull requests.

— Claude for @gfraiteur

Metalama.Framework/docs/future/type-facets.md
```diff
@@ -0,0 +1,552 @@
...
+    Declaration,
+
+    /// <summary>The type is a class or a struct that carries <c>UnionAttribute</c>.</summary>
+    Attribute,
+
+    /// <summary>
+    /// The type is a union declared in a referenced assembly. The compiled form does not record the authoring form,
+    /// so the two forms above cannot be told apart there.
+    /// </summary>
+    External
```
gfraiteur: Rename External to None, value 0.

gfraiteur:
`UnionKind.External` is renamed to `UnionKind.None`, with the explicit value zero, and it is now the first member.
`TypeFacetKind.None` gains the explicit value zero as well, so the two enumerations agree.

The documentation of the member states the meaning it has where it is read, which is on `IUnionFacet.UnionKind`.
The type is a union in every case there, so `None` says that the kind of the union is not known, which happens for a
union declared in a referenced assembly because the compiled form does not record the authoring form. Tell me if you
would rather the member read as "the type is not a union", which is the other sense the name carries, and I will
change the wording.

The bullet in section 7 is updated to match.

— Claude for @gfraiteur

Issue #2031, opened 2026-09-15
Revisit the multicast selection of union constructors against PostSharp


gfraiteur:
`Metalama.Extensions.Multicast` exists to reproduce the multicasting of PostSharp, so that an aspect migrated from
PostSharp selects the same targets. Issue #1946 added one exclusion to that selection without a PostSharp behaviour to
compare it against, because PostSharp does not support the unions of C# 15. The exclusion must be revisited once
PostSharp defines what an assembly-level multicast aspect does with a union.

#### What was decided in #1946

`MulticastImplementation.IsCandidateInstanceConstructor` admits an implicitly declared constructor only when it has no
parameter, which is the parameterless constructor that the compiler adds to a type that declares none. A union
declaration always receives that constructor, because every case constructor of a union takes exactly one parameter.
An aspect with `MulticastTargets.InstanceConstructor` therefore selected it, and materializing it produced code that
the compiler rejects with CS9375, because a constructor declared in a union must chain through a `this(...)`
initializer.

The parameterless constructor of a union declaration is now excluded from the selection. Section 13 of
`Metalama.Framework/docs/2027.0/DECISIONS.md` records the choice and the alternative that was not taken, which was an
eligibility rule on constructor advice in the engine.

#### Why this needs a second look

The choice was made from the language rules alone. Two questions have no answer today, and both are answers that
PostSharp will give.

- Whether PostSharp selects the implicit parameterless constructor of a union at all. If it does not select it, the
  exclusion is the compatible behaviour and nothing changes. If it selects it and reports a diagnostic, Metalama
  reporting nothing is a difference that a user migrating an aspect would notice.
- Whether the case constructors of a union are multicast targets. They are implicitly declared and take one parameter
  each, so the current selection skips them, and an aspect that instruments every constructor of every type therefore
  instruments no constructor of a union. That may or may not be what PostSharp does.

#### What to do

Compare the behaviour of `Metalama.Extensions.Multicast` on a union with the behaviour of PostSharp once PostSharp
supports unions, and align the selection or record the difference as an intended one. Update section 13 of
`DECISIONS.md` and the aspect test `Union.cs` of `Metalama.Extensions.Multicast.AspectTests`, which pins the current
selection, including the contrast case of an ordinary struct whose implicit parameterless constructor is selected.

The underlying defect, that materializing the implicit parameterless constructor of a union produces CS9375, is not
specific to multicasting and belongs to the constructor advice of the engine. That part is unaffected by this issue.

[...]

Issue #2030, opened 2026-09-15
[Observable] generates an equality operator comparison for a struct that declares none


gfraiteur:
The `[Observable]` aspect generates `field != value` in the setter that it introduces for a property whose type is a
value type, and the compiler reports CS0019 when that type declares no equality operator. The generated code does not
compile, and the user has no way to change the comparison.

#### Reproduction

```csharp
public struct S
{
    public int X { get; set; }
}

[Observable]
public class C
{
    public S Value { get; set; }
}
```

The generated setter is:

```csharp
set
{
    if (_value != value)
    {
        _value = value;
        OnPropertyChanged("Value");
    }
}
```

and the compiler reports:

```
error CS0019: Operator '!=' cannot be applied to operands of type 'S' and 'S'
```

A struct declared as a `record struct`, or one that declares the operator itself, is not affected, because the
compiler synthesizes the operator for a record struct.

#### Cause

`ClassicObservabilityStrategyImpl` chooses the comparison from the type kind alone:

```csharp
var comparisonKind = fieldOrProperty.Type is INamedType nt && nt.SpecialType != SpecialType.ValueTask_T
    ? EqualityComparisonKind.EqualityOperator
    : EqualityComparisonKind.DefaultEqualityComparer;
```

Every named value type other than `ValueTask<T>` is therefore compared with the equality operator, whether or not it
declares one. The intended fallback, `EqualityComparisonKind.DefaultEqualityComparer`, already exists and generates
`EqualityComparer<T>.Default.Equals`, which works for every type.

#### Proposed change

Select `EqualityComparisonKind.EqualityOperator` only when the type declares or inherits an applicable `operator ==`,
and `EqualityComparisonKind.DefaultEqualityComparer` otherwise. A primitive type, an enum and a string keep the
operator, because the language defines it for them.

#### Affected code

- `Metalama.Patterns/src/Metalama.Patterns.Observability/Implementation/ClassicStrategy/ClassicObservabilityStrategyImpl.cs:660-663`
- `Metalama.Patterns/src/Metalama.Patterns.Observability/Implementation/ClassicStrategy/Templates.cs:294-299`

#### Notes

The defect is reachable without C# 15 and is not specific to unions, but a union of C# 15 makes it more likely to be
met: a union declaration is a struct, the compiler synthesizes no equality operator for it, and a union is a natural
type for an observable property. It was found while adding the aspect test `UnionPropertyIsNotWarned` of issue #1946,
[...]

Issue #2029, opened 2026-09-15
Empty ineligibility justification for an inheritable aspect whose rule is under ExceptForInheritance


gfraiteur:
An inheritable aspect whose eligibility rule is declared under `ExceptForInheritance` reports `LAMA0037` with an
empty justification when the rule is the only one that the target fails and the target cannot be inherited. The
message reads `The aspect 'X' cannot be applied to the type 'Y' because .`, so the user is told that the aspect is
ineligible and is not told why.

#### Reproduction

Applying `[Observable]` of `Metalama.Patterns.Observability` to a struct is enough. The aspect is declared
`[Inheritable]`, and its eligibility is:

```csharp
void IEligible<INamedType>.BuildEligibility( IEligibilityBuilder<INamedType> builder )
{
    builder.MustNotBeStatic();
    builder.ExceptForInheritance().MustSatisfy( x => x.TypeKind is TypeKind.Class, x => $"{x} must be a class or a record class" );
}
```

The reported diagnostic is:

```
error LAMA0037: The aspect 'Observable' cannot be applied to the type 'S' because .
```

The expected justification is `'S' must be a class or a record class`.

#### Cause

`CompilationAspectSource.CreateAspectInstance` asks for the justification of the scenario that matches the aspect
rather than the scenario that failed:

```csharp
var requestedEligibility = aspectInstance.IsInheritable ? EligibleScenarios.Inheritance : EligibleScenarios.Default;
var reason = aspectClass.GetIneligibilityJustification( requestedEligibility, ... )!;
```

A rule added through `ExceptForInheritance` reports `EligibleScenarios.Inheritance` when its predicate fails, which
means that the declaration remains eligible for inheritance. `AndEligibilityRule.GetIneligibilityJustification`
returns the justification of the first rule whose eligibility does not cover the requested scenario, and that rule
covers `Inheritance`, so the loop skips it and the method returns `null`.

The eligibility is nevertheless `None`, because `AspectInstance.ComputeEligibility` removes
`EligibleScenarios.Inheritance` from the result when the target declaration cannot be inherited, which is the case
of a struct. The reason the aspect is ineligible is therefore the rule that the justification loop skipped.

#### Affected code

- `Metalama.Framework/src/Metalama.Framework.Engine/Aspects/CompilationAspectSource.cs:119-122`
- `Metalama.Framework/src/Metalama.Framework.Engine/Aspects/AspectInstance.cs:129-139`
[...]

PR #2032, opened 2026-09-15
Support C# 15 unions in the pattern and extension libraries (#1946)

### LINKED ISSUE #1946: C# 15 unions: pattern and extension libraries
**Size** L · **Blocked by** #1941 · **Repositories** `metalama/Metalama` · **Story** [S-18-6](``https://github.com/metalama/Metalama/blob/develop/2027.0/Metalama.Framework/docs/2027.0/user-stories/S-18-6-unions-in-pattern-and-extension-libraries.md``) · **Findings** [UT-14](https://github.com/metalama/Metalama/blob/develop/2027.0/Metalama.Framework/docs/2027.0/06-user-tfm-patterns-tests-docs.md), [UT-14a](https://github.com/metalama/Metalama/blob/develop/2027.0/Metalama.Framework/docs/2027.0/06-user-tfm-patterns-tests-docs.md), [UT-14b](https://github.com/metalama/Metalama/blob/develop/2027.0/Metalama.Framework/docs/2027.0/06-user-tfm-patterns-tests-docs.md), [UT-14c](https://github.com/metalama/Metalama/blob/develop/2027.0/Metalama.Framework/docs/2027.0/06-user-tfm-patterns-tests-docs.md), [UT-14d](https://github.com/metalama/Metalama/blob/develop/2027.0/Metalama.Framework/docs/2027.0/06-user-tfm-patterns-tests-docs.md)

---

A union is an ordinary struct with an opaque value property for every library that reads the code model, and that
premise has four consequences in the pattern and extension libraries plus one in the reference graph. Two of them are
product defects: the immutability classification calls a union mutable, which produces spurious observability
warnings, and the caching key generation and the cache item serialization treat a union as an opaque struct and
produce a silently wrong result.

#### Context

The immutability rule must key on the declaration form and not on the Roslyn union flag, because that flag is also
true for a hand-written type carrying the union attribute whose state is unconstrained. The caching defect has a
practical discriminator, which is the interface that the compiler makes every union implement, but the affected
projects target `net472` and `netstandard2.0` and cannot bind to it at compile time, so the check has to be made in
another way. The remaining three consequences are tests: the observability aspect already rejects a union with two
diagnostics and that behaviour is pinned rather than changed, and the multicast selector is correct for every target
except the implicit parameterless constructor, where materializing an override produces a compiler error. The
reference graph item needs one override in the core reference index walker, without which a reference from a union to
its case types is never attributed and an architecture rule under-reports with no diagnostic; the Premium half is two
architecture rule tests, filed separately, because Premium cannot compile a union until its own latest variant is
renamed.

#### Scope

- Classify a union declaration in the immutability library as shallowly immutable, and as deeply immutable when every
  case type is, treating an interface, a type parameter, a nullable value type and a nested union case type
  conservatively, and document the rule against the definition the library gives rather than against the `readonly`
  modifier.
- Make the caching key generation and the cache item serialization treat a union by its case value rather than as an
  opaque struct, and choose a discriminator that the target frameworks of the affected projects can express.
- Add the observability test that pins the two diagnostics with which the library already rejects a union.
- Prevent the multicast selector from selecting the implicit parameterless constructor of a union, or make
  constructor advice ineligible on a union in the engine, and record which of the two was chosen.
- Attribute a reference from a union declaration to its case types in the core reference index walker, by entering
  the union type declaration as the current declaration before visiting its case list.

#### Acceptance criteria

- A union of immutable case types does not produce an observability warning, and a hand-written type carrying the
  union attribute is not classified from the attribute alone.
- A cached method whose parameter is a union produces a different cache key for two different cases, and a cached
  value of union type survives a serialization round trip.
- The multicast and observability tests are committed with their expected output.

#### Not in scope

This story does not add the architecture rule tests of `Metalama.Premium`, which are filed in that repository.

— Claude for @gfraiteur

gfraiteur:
## Summary

A union declaration is an ordinary struct for every library that reads the code model. That premise made four of the
pattern and extension libraries wrong on a union, and left the reference graph blind to its case types.

- **Immutability.** `GetImmutabilityKind` classifies a union declaration as `Shallow`, and as `Deep` when every case
  type is deeply immutable. An interface, a type parameter, a nullable value type and a nested union case are treated
  conservatively. The rule rests on the definition that `ImmutabilityKind.Shallow` gives, which is about instance
  fields and automatic property setters, and not on the type being declared `readonly`. It keys on
  `INamedType.Facets.Union.UnionKind == Declaration` and never on `IsUnion` alone, because the state of the attribute
  form is unconstrained.
- **Observability.** The classification above removes the spurious `LAMA5161` that was reported for a property of
  union type. The rejection of `[Observable]` on a union declaration is unchanged and is now pinned by a test.
- **Caching keys.** `FormatterRepository` chooses a `UnionFormatter` for a union, which formats the type and the value
  of the current case. The default formatter called `ToString`, for which a union declaration synthesizes no override,
  so every value of a union produced the name of the union type: a cached method whose parameter is a union returned
  the entry of one case for another, and `[InvalidateCache]` invalidated by the same collapsed key. The new public
  `Flashtrace.Formatters.Utilities.UnionHelper` recognizes a union by the interface that the compiler makes it
  implement, matched by full name because `net472` and `netstandard2.0` cannot bind to it.
- **Caching serialization.** `JsonCachingSerializer` converts a union by the type and the value of its case. The
  `Value` property of a union has no setter, so the default conversion of `JsonSerializer` deserialized the default
  value of the union and the cached value was lost.
- **Multicasting.** The selector no longer admits the implicit parameterless constructor of a union declaration.
  Materializing it produced code that the compiler rejects with CS9375, because a constructor declared in a union must
  chain through a `this(...)` initializer. Section 13 of `DECISIONS.md` records the choice and marks it provisional.
- **Reference graph.** `ReferenceKinds` gains `UnionCaseType`, and `ReferenceIndexWalker.VisitUnionDeclaration`
  indexes the case list with the union as the origin. Without it an architecture rule placed on a case type
  under-reported with no diagnostic.

## Public API

- `ReferenceKinds.UnionCaseType`, a new member of the existing flags enumeration. `ReferenceKinds.All` includes it, so
  a rule that uses the default reference kinds now also matches a union case type. That is the intended fix and is
  reachable only from C# 15 source.
- `Flashtrace.Formatters.Utilities.UnionHelper`, a new public static class with `IsUnion` and
  `TryGetValuePropertyGetter`. It is public because `Metalama.Patterns.Caching.Backend` needs the same test and
  internals are made visible to test projects only.

## Behaviour change to review

`JsonCachingSerializer` now copies the `JsonSerializerOptions` given by the caller, because it adds the union converter
to them and modifying the caller's instance would change every other use of it. A caller who modified that instance
after constructing the serializer previously saw the modification take effect, and no longer does. The window is
narrow, because `System.Text.Json` freezes the options on first use in any case. I did not apply the `breaking` label;
say so if you read this differently.

## Verification

Built and tested against `Metalama.Premium` with a local dependency on this branch: the Premium build is clean and its
66 architecture and 36 validation tests pass. Premium was then restored to its build server dependency.

## Defects found on the way, filed separately

- #2029 — `LAMA0037` reports an empty justification for an inheritable aspect whose eligibility rule is declared under
  `ExceptForInheritance`. Reachable without C# 15. The new observability test pins the message as it is today.
- #2030 — `[Observable]` generates an equality operator comparison for a struct that declares none, so the generated
  code does not compile. Reachable without C# 15. The new observability test avoids it with a get-only property.
- #2031 — revisit the multicast selection of union constructors once PostSharp supports unions, because this library
  reproduces the multicasting of PostSharp and there is no behaviour to compare the new exclusion against.

## Issues Fixed

Fixes #1946 - C# 15 unions: pattern and extension libraries

— Claude for @gfraiteur


Metalama.Patterns/src/Flashtrace.Formatters/Utilities/UnionReflection.cs
```diff
@@ -0,0 +1,95 @@
...
+/// A type of another assembly that declares an interface of the same full name is recognized as a union as well. That
+/// is the cost of matching by name, and the name belongs to a namespace that the base class library reserves for the
+/// compiler.
+/// </para>
+/// <para>
+/// The results are cached per type, so a repeated query costs one dictionary lookup.
+/// </para>
+/// </remarks>
+[PublicAPI]
+public static class UnionReflection
```
gfraiteur: Rename to UnionHelper

Metalama.Patterns/src/Flashtrace.Formatters/Utilities/UnionReflection.cs
```diff
@@ -0,0 +1,95 @@
...
+    private static readonly ConcurrentDictionary<Type, Func<object, object?>?> _caseValueGetters = new();
+
+    /// <summary>
+    /// Returns a function that reads the value of the current case of a union, or <c>null</c> when the type is not a
+    /// union.
+    /// </summary>
+    /// <param name="type">The type to evaluate.</param>
+    /// <returns>A function whose parameter is the union, boxed when it is a value type, and whose return value is the
+    /// value of the current case, or <c>null</c> when <paramref name="type"/> is not a union.</returns>
+    public static Func<object, object?>? GetCaseValueGetterOrNull( Type type ) => _caseValueGetters.GetOrAdd( type, CreateCaseValueGetter );
```
gfraiteur: Rename to `bool TryGetValuePropertyGetter( Type type, [notNullWhen...] Func<object, object?>? getter )` and change to try pattern

gfraiteur:
Both review comments are addressed in 593699da9e. They are answered here rather than in their threads, because the
threads belong to a review that is still pending, and a reply to such a thread would require a second pending review
on the pull request, which GitHub refuses.

**Rename to `UnionHelper`.** Done. The file was renamed with the class, so the change is recorded as a rename.

**`TryGetValuePropertyGetter` and the try pattern.** Done:

```csharp
public static bool TryGetValuePropertyGetter( Type type, [NotNullWhen( true )] out Func<object, object?>? getter )
```

Two notes on it. The parameter is `out`, which the sketch in the comment omitted and the pattern requires.
`NotNullWhenAttribute` resolves on every target framework of the package: .NET Framework 4.7.2 and .NET Standard 2.0
take it from the `NullableAttributes` polyfill that PostSharp.Engineering compiles into the project, and the same
pattern is already used in `Metalama.Patterns.Caching.Aspects` and `Metalama.Patterns.Contracts`.

`IsUnion` is kept and now delegates, as `TryGetValuePropertyGetter( type, out _ )`. The private cache field and factory
were renamed to `_valuePropertyGetters` and `CreateValuePropertyGetter`, and the documentation now speaks of the
`Value` property rather than of a case value getter, so that the vocabulary follows the new name.

The two call sites in `FormatterRepository` each collapse to a single expression, because the test and the getter now
arrive together.

The Metalama.Patterns solution builds with no new warning, and the formatter and caching unit tests pass on both
target frameworks.

— Claude for @gfraiteur

gfraiteur:
Answering the two comments of the Copilot review. They are answered here rather than in their threads, because a
review that is still pending exists on this pull request, and GitHub allows one pending review per user, so the
creation of any review comment is refused with `422`.

**The public API section names the old type.** Correct. The section was written before the review that renamed the
type in 593699da9e, and it still named `UnionReflection` and `GetCaseValueGetterOrNull`. The description now names
`UnionHelper`, `IsUnion` and `TryGetValuePropertyGetter`, in both places where the old name occurred. The
implementation was already the current one.

**A nullable case cannot be deserialized.** The two observations are right and the conclusion does not follow. Reading
`IUnion.Value` does box a `Nullable<T>` that has a value as its underlying type, and the constructor that takes that
case does have a parameter of type `Nullable<T>`: for `union U( int?, string )` the constructors are
`U(System.Nullable<System.Int32>)` and `U(System.String)`, and `Value` reports `System.Int32`.

The premise that `Nullable<T>.IsAssignableFrom(typeof(T))` is false is not true. It returns `true`, on both target
frameworks of the project, so the fallback that searches for an assignable constructor already matched
`U(System.Nullable<System.Int32>)` and the case already round-tripped. `ANullableCaseSurvivesTheRoundTrip` was added
to `UnionSerializerTests` as the comment asks, and it passes on .NET 10 and on .NET Framework 4.7.2 without any change
to the converter.

The comment does point at a real weakness next to the one it describes, which 453bc72ae7 fixes. The search took the
first assignable constructor, and the order in which `Type.GetConstructors` returns the constructors is not defined by
the runtime. A union that declares two cases which one boxed value satisfies, as `object` and `int?` both do for a
boxed `int`, was therefore rebuilt from either of them. A nullable case is now preferred over a merely assignable one,
so the result follows the declaration rather than the order. The residual ambiguity is documented on the converter:
the `Value` property carries no case identifier, so no reading of it can tell those two cases apart.

— Claude for @gfraiteur

PR #2041, opened 2026-09-19
Move the Docker tests to the shared harness

gfraiteur:
The Docker tests were run by a launcher of this repository, `DockerTests.ps1`, which discovered tests by a directory named after a platform. PostSharp.Engineering now carries a harness for exactly that, so the launcher is generated and what remains here is the tests themselves.

**Layout.** A test is a directory holding a manifest, a Dockerfile, an entry point that runs on the host and the cases that run in the container. The platform moves from the directory's position into the manifest, which is why `NonStandardDotNetRoot` becomes two directories: the Linux and Windows halves need different images, and a test directory holds one Dockerfile.

`ReferenceAssemblyArchitectureMismatch` changes more than the others. It used to copy its script into the image and run it as the entry point; the repository is mounted, so it now reads the script from there and the image carries the two .NET installations and nothing else. Its command runs under Windows PowerShell rather than pwsh, because that image has no PowerShell 7 and a test image is not required to carry one.

**Build configurations** become one per platform in a sub-project of their own, with a composite over them. They run on the agent rather than in the product build container: a test that needs a container of its own cannot be started from inside one.

**Engineering** moves from 2023.2.454 to 2023.2.461. 460 provides the harness and recreates the substituted drive in a Windows test container, without which a NuGet configuration naming `X:` paths cannot be read there — that is what `NonStandardDotNetRootWindows` needs. 461 adds two things: the generated cleanup step now also runs the command the agent names in `BUILDAGENT_CLEANUP_SCRIPT`, after removing the containers, so a build stops leaving root-owned files in the checkout for the next build to fail on; and `DockerBuild.ps1` stops printing the values of its `-e` arguments, which on a developer machine meant credentials in clear on the console.

**Known failure, not introduced here.** `CompilerLogs` does not find the directory it expects under `/tmp`, and fails identically through the former launcher.

— Claude for @gfraiteur


gfraiteur:
Superseded by #2044, which brings the same migration forward from 2026.1 (#2043, now merged) instead of carrying a second, independently written copy on this line.

The two differ only cosmetically — `Container.ps1` entry points and unhyphenated `NonStandardDotNetRootLinux` / `…Windows` directories here, against `test.ps1` and `NonStandardDotNetRoot-Linux` / `-Windows` there. Merging them produced about twenty rename/rename conflicts, including ones where git cross-matched the byte-identical Linux and Windows halves of `NonStandardDotNetRoot`, so converging on the 2026.1 version keeps the two lines from diverging further.

#2044 also carries the `CompilerLogs` fix that this branch lacks: the test asserted the pre-#1650 `/tmp/Metalama`, which nothing writes to on Unix any more.

Closing rather than merging. The branch stays on the remote.

— Claude for @gfraiteur

PR #2042, opened 2026-09-19
Consume PostSharp.Engineering 2023.2.461

gfraiteur:
From 2023.2.427. Three changes reach this branch; only `Directory.Packages.props`, `DockerBuild.ps1` and `.teamcity/settings.kts` change, and no build configuration is added or removed.

**Root-owned files in the checkout.** A container runs as root while the agent does not, so what a build writes into the mounted repository is owned by root on the host and the agent user cannot remove it afterwards. Swabra logs `unable to delete` and continues, and then the *next* build of any kind — not necessarily a Docker one — fails at checkout. The generated cleanup step now also runs the command the agent names in `BUILDAGENT_CLEANUP_SCRIPT`, after removing the containers. All seven `DockerCleanup` steps carry it; an agent that names no command gets none, which is every Windows agent and every developer machine.

This was verified on the PostSharp side by four builds in two pairs, one pair per Linux agent: a first run that invokes the script, and a consecutive second run at zero permission denials and zero Swabra deletion failures.

**Substituted drive on Windows.** `DockerBuild.ps1` recreates it inside a test container, so a NuGet configuration naming absolute host paths can be read there.

**Secrets on the console.** `DockerBuild.ps1` stops printing the values of its `-e` arguments. TeamCity masked these by value, so the exposure was on developer machines rather than in CI logs.

The `DockerBuild.ps1` diff is large because it spans 34 engineering releases; it is a generated file and was produced by `Build.ps1 generate-scripts`.

— Claude for @gfraiteur


gfraiteur:
Merged into #2041 (`5a68a16c99`) rather than into `develop/2026.1`, and verified contained: this branch's `43b36303e8` is an ancestor of that PR's head. Closing in favour of #2041.

One consequence worth stating: the 2026.1 line therefore stays on PostSharp.Engineering 2023.2.427, so `BUILDAGENT_CLEANUP_SCRIPT` does not reach its `DockerCleanup` steps. The branch is left on the remote, so this can be reopened if 2026.1 should get the bump on its own.

— Claude for @gfraiteur

